PR #24, opened 2016-10-24
Added new menu item "move system root to usb"

jupe76:
Hi,

I've added the new menu item "move system root to usb", see https://community.openhab.org/t/raspi-2-move-system-root-to-usb-ssd/15243

I've tested it multiple times, works fine here :-)

### DIFF
--- a/openhabian-setup.sh
+++ b/openhabian-setup.sh
@@ -353,7 +353,7 @@ openhab_shell_interfaces() {
   echo "OK"
 }
 
-wifi-setup-rpi3() {
+wifi_setup_rpi3() {
   echo -n "[openHABian] Setting up RPi 3 Wifi... "
   if ! is_pithree ; then
     if [ -n "$INTERACTIVE" ]; then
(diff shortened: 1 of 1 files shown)


ThomDietrich:
Hey! Great addition! I've committed a few minor fixes, apart from that everything looks nice ;) Two things:

- Please have a look if you agree with my changes
- We should have an emergency ready, helping the user if his system does not reboot correctly at the end. Would you add a line to the last message box saying something like "If reboot not successful - move cmdline.txt back"

jupe76:
Hi,
thanks for correcting the typos, I was somewhat lazy with that. 
The only thing I'm really unsure is "a USB device" vs. "an USB device" and "an SDD" vs. "a SSD".
I've learned to use "an", if the following word starts with a vocal (a,e,i,o,u), so to my knowledge it should read "an USB device" and "a SSD".

I've added a line about to copy back cmdline.txt.sdcard in case of boot failure.

ThomDietrich:
Hey!
no problem, I'm a bit of a perfectionist at that, so kind of the other side of the same problem :sweat_smile: 

>I've learned to use "an", if the following word starts with a vocal (a,e,i,o,u)

Almost right :) This is a very common mistake. The rule is true but you have to apply it to the pronunciation of a word. That's why it is "a [**y**u-es-be]" and "an [**e**s-di-card]".

ThomDietrich:
@jupe76 I totally forgot about that but did you try this?
https://community.openhab.org/t/openhabian-hassle-free-rpi-image/13379/185?u=thomdietrich

jupe76:
> @jupe76 I totally forgot about that but did you try this?
https://community.openhab.org/t/openhabian-hassle-free-rpi-image/13379/185?u=thomdietrich

Hi, saw it but did'nt try it so far. I'll perhaps give it a shot the next days.

Issue #12, opened 2016-09-05
optional component: authentication


ThomDietrich:
https://community.openhab.org/t/authentication-in-oh2/1723/28

probably  via nginx revproxy

BClark09:
Is there anyway I can help here?

ThomDietrich:
Hey @BClark09 !
I would like to provide one option allowing a user to select authentication and set up a certificate. It would be acceptable to only perform the first few steps and then tell the user what to do next.

Because of the many options and parameters it might be hard to provide one setup fitting all users.

How about that:

1. Prompt for authentication username and password
2. Install nginx, configure http and self-signed https with authentication enabled
3. Ask the user if he wants to set up a signed certificate, either by providing it or by creating one using Let's Encrypt.

What do you think is feasible?

BClark09:
I think the three steps are fine as is. You could just prompt for a username, password and a domain name (which can be left blank):

- If the domain is blank, automatically detect the public IP address and use that for nginx configuration?
- If the domain is blank use OpenSSL. Otherwise use Let's Encrypt?

Also a couple of considerations of using Let's Encrypt and Certbot:

- It will be important to test that the domain name entered points to the detected public IP address of the Pi, and prevent the script from going further if it doesn't.
- The cron setup should be at a randomly generated hour of the day, otherwise you may end up with many different Pi's all trying to connect to the Let's Encrypt servers at the same time.

ThomDietrich:
Sounds doable. We'll say how far we can go.
If you want to help out, feel free to post a few commands / steps here or create a pull request.

To get an idea: https://github.com/ThomDietrich/openhabian/blob/master/openhabian-setup.sh#L415-L445

BClark09:
I'm not sure if the file writing part is the best way to go about it but this should be a good start? I have a spare RPi 2 if you would like me to test any builds with this. 

```bash
#!/bin/bash 

## These need setting from the user...
username="openhab"
password="habopen"

## This is set to the public ip if the user hasn't entered a valid domain name
domain="foo.bar"

## These are set depending on the user's prefered options...
VALIDDOMAIN=false
SECURE=true
AUTH=true

## Action
apt -y install nginx
rm -rf /etc/nginx/sites-enabled/default

if [ "$AUTH" = true ]; then
	apt -y apache2-utils
	htpasswd -b -c /ect/.htpasswd $username $password
fi

if [ "$SECURE" = true ]; then
	USINGCERTBOT=false
[...]

Issue #11, opened 2016-09-05
optional component: set up https+letsencrypt 


ThomDietrich:
https://community.openhab.org/t/securing-openhab-with-free-ssl-lets-encrypt-certificates/7918

ThomDietrich:
Fixed in https://github.com/ThomDietrich/openhabian/pull/26 and following.
Thanks @BClark09

PR #26, opened 2016-10-30
Added NGINX (auth/HTTPS) setup

BClark09:
Implements #12 and #11

### DIFF
--- a/includes/nginx.conf
+++ b/includes/nginx.conf
@@ -1,0 +1,40 @@
+#################################
+# openHABian NGINX Confiuration #
+#################################
+
+## Redirection
+#server {
+#   listen                          80;
+#   server_name                     DOMAINNAME;
+#   return 301                      https://$server_name$request_uri;
+#}
+
+## Reverse Proxy to openHAB
+server {
+    listen                          80;
+#   listen                          443 ssl;
+    server_name                     DOMAINNAME;
+#   add_header                      Strict-Transport-Security "max-age=31536000; includeSubDomains";
+
+## Secure Certificate Locations
+#   ssl_certificate                 CERTPATH;
+#   ssl_certificate_key             KEYPATH;
+
+    location / {
+        proxy_pass                              http://localhost:8080/;
+        proxy_buffering                         off;
+        proxy_set_header Host                   $http_host;
+        proxy_set_header X-Real-IP              $remote_addr;
+        proxy_set_header X-Forwarded-For        $proxy_add_x_forwarded_for;
+        proxy_set_header X-Forwarded-Proto      $scheme;
+
+## Password Protection
+#       auth_basic                              "Username and Password Required";
+#       auth_basic_user_file                    /etc/nginx/.htpasswd;
+    }
+
+## Let's Encrypt webroot location
+#   location /.well-known/acme-challenge/ {
+#       root                                    /var/www/DOMAINNAME;
+#   }
+}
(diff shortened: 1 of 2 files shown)


ThomDietrich:
Hey Ben, great work! I've merged this right away to give it a test drive. (Testing right now)

I still have one requests...
Could you add the additional whiptail dialog to inform the user of the details before starting the process? I've done that in almost all routines to avoid accidental clicks... [see here](https://github.com/ThomDietrich/openhabian/blob/master/openhabian-setup.sh#L382-L390)

ThomDietrich:
I've added a few small improvements to the printouts. Why did you decide to divide read and echo for the domain?

Another thing I just realized: One can not cancel while being in the prompts because I've trapped SIGINT. Would agree it would be a good idea to present all given information and ask for yet another confirmation before the setup process begins?

ThomDietrich:
I encountered the first real problem: My domain is resolved though a CNAME entry pointing to a dyndns provider domain: `openhab.mydomain.de -> abcd.no-ip.biz -> 88.66.68.21`

Now `dig +short openhab.mydomain.de` produces:
```
abcd.no-ip.biz.
88.66.68.21
```

which as a string is not equal to my wanip ("88.66.68.21"). Sadly `dig +short openhab.mydomain.de A` will not solve this problem as a CNAME is always part of the answer (if you know better, let me know). How do you think we can solve this?

1. We could extend the command to always use the last line of the output
2. We could extend [the else part](https://github.com/ThomDietrich/openhabian/blob/master/openhabian-setup.sh#L784) with a "I don't care, just use the given domain" option.

BClark09:
> Could you add the additional whiptail dialog to inform the user of the details before starting the process? I've done that in almost all routines to avoid accidental clicks

Sure thing!

---
> Why did you decide to divide read and echo for the domain?

Er.... Good question! I should have used a read prompt for this.

---
>Would agree it would be a good idea to present all given information and ask for yet another confirmation before the setup process begins?

Yes this would be a safe bet, otherwise some weird default nginx configurations might occur. I'll move the install of nginx to after this prompt too?

---
> CNAME entry pointing to a [..] domain

This is a normal way of doing this and for some reason didn't think of it! I don't think the "I don't care" option would be a good idea because if a domain can't eventually be resolved to the ip of the wan at the time of the setup, then this may cause problems with the proxy forwarding if the issue persists. Perhaps a better way of doing this is:
```
domainip=$(host $domain | awk '/has address/ { print $4 ; exit }')
```

ThomDietrich:
> Yes this would be a safe bet, otherwise some weird default nginx configurations might occur. I'll move the install of nginx to after this prompt too?

As you think. Sounds reasonable. ;)

>I don't think the "I don't care" option would be a good idea because ...

I was thinking of the tech-savvy user, knowing what he's doing but maybe it's better to not tempt the normal end user to try and occupy a domain he doesn't own.

Why not `dig +short $domain |tail -1`? This should always return the A entry

BClark09:
I've just been reading on this at it looks like the last line will always be an ip address. My hesitancy was if the last line was an ipv6 address if it ever becomes a default (rather than depending on AAAA type). But thinking about it changing $(dig +short myip.opendns.com @resolver1.opendns.com |tail -1) will always output the same IP and type.

BClark09:
@ThomDietrich Would the input be better served through whiptail dialogue? That way we can keep to the same input type instead of suddenly shifting the user to reading the command line.

ThomDietrich:
I've resorted to whiptail dialogs whenever possible to make things as "creamy" as possible for the end user but you already got your dialogs with read. So I would say it'ss totally up to you ;) As you know there is an example in the Karaf console function.

PR #27, opened 2016-11-01
Fixed initial nginx problems

BClark09:
- User input is now done by whiptail.
- Checks for failures are put into place for key generation.
- A confirmation of settings chosen is presented before any changes are made.
- dig retrieves the last IP in it's lookup. Which should solve problems for CNAME > IP routes

### DIFF
--- a/openhabian-setup.sh
+++ b/openhabian-setup.sh
@@ -734,114 +734,159 @@ influxdb_grafana_setup() {
 }
 
 nginx_setup() {
-function comment {
-  sed -i "$1"' s/^/#/' "$2"
-}
-function uncomment {
-  sed -i "$1"' s/^ *#//' "$2"
-}
+  introtext="This will enable you to connect to openHAB through a HTTP port and optionally secure it with a username/password and an SSL certificate."
+  failtext="Sadly there was a problem setting up the selected option. Please report this problem in the openHAB community forum or as a openHABian GitHub issue."
 
-echo "Installing NGINX and DNS Utilities..."
-apt -y -q install nginx dnsutils
+  if [ -n "$INTERACTIVE" ]; then
+    if ! (whiptail --title "Description, Continue?" --yes-button "Continue" --no-button "Back" --yesno "$introtext" 15 80) then return 1; fi
+  fi
 
-AUTH=false
-SECURE=false
-VALIDDOMAIN=false
-matched=false
+  function comment {
+    sed -i "$1"' s/^/#/' "$2"
+  }
+  function uncomment {
+    sed -i "$1"' s/^ *#//' "$2"
+  }
 
-echo -n "Obtaining Public IP Address... "
-wanip=$(dig +short myip.opendns.com @resolver1.opendns.com)
-echo "$wanip"
+  echo "Installing DNS Utilities..."
+  apt -y -q install dnsutils
 
-read -r -p "Would you like to secure your openHAB interface with username and password? [y/N]: " response
-if [[ $response =~ ^([yY][eE][sS]|[yY])$ ]]; then
-  read -p "Enter a username to sign into openHAB: " username
-  while [ "$matched" = false ]; do
-    read -s -p "Enter a password for $username: " password
-    echo ""
-    read -s -p "Please confirm the password: " secondpassword
-    echo ""
-    if [ "$password" = "$secondpassword" ]; then
-      matched=true
-      echo "Password accepted!"
+  AUTH=false
+  SECURE=false
+  VALIDDOMAIN=false
+  matched=false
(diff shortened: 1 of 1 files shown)


BClark09:
Also, royally screwed up the Diff with formatting changes. Really sorry about that!

ThomDietrich:
>Also, royally screwed up the Diff with formatting changes. Really sorry about that!
Not a problem, do you know `w=1` ;)
https://github.com/ThomDietrich/openhabian/pull/27/files?w=1

ThomDietrich:
Looks great! I'll merge and test this evening. Thanks!

Btw. I saw that you had to delete your fork and fork anew. Do you know that it's best to work in branches for PRs?

Again thanks for your contribution! I hope to see more from you in the future ;)

BClark09:
I'm yet to take a self driven crash course in the wonders of git. So all of this is fairly new to me and have just been pushing buttons until something reasonable happens :see_no_evil: . Thanks a lot for the advice!

ThomDietrich:
Hehehe. Been there, done that.

I found it very useful to use a GUI in the beginning. Many might argue that only by using the command line you will really understand the git internals but that's not what you are looking for in the beginning. I can recommend **GitKraken** as an easy to use and intuitive solution and my personal favorite is **SmartGit** (powerful, free, platform independent). I can not recommend the famous SourceTree as it will just throw the command line output at you and you'll have to figure out what's going on... in this case you are better off with the command line to begin with... But that is just my opinion!

Example for what you did here, maybe it's useful to kickstart your crash course:

* The first thing you do is **create a fork** of my repository on GitHub. Neither of the mentioned clients can do that for you (there are yet other tools out there).
* Next you would **clone your repository** in SmartGit and make sure to tick **"Create upstream remote"**. This way you will not only clone your repo but also see changes in the upstream repo, i.e. mine. (the upstream repo is just the name if a second "git remote")
* Now the first thing you should always do is (visit the "Log" view and) **create a new branch** on top of the latest commit.
* Next you can do your changes, commit to your new branch and at last **push your changes** back to your repository. GitHub will now invite you to **create a PR** based on that branch.
* After the **branch was merged** you probably want to bring your repository up to date (that's the point where you deleted your fork).
* **Pull** latest changes from the upstream repository master and **Push** them to your repository master. This way you will receive the merge commit and all other changes since you created your fork.
* You can now start to create yet another branch based on this one and start from the beginning.

After a while your Log could look like this:

2016-11-01 16_01_29-openhab-docs - log for

BClark09:
Wow, thanks for the heads up. Hopefully my next PR will be the start of proper usage!

ThomDietrich:
Amazing. Worked like a charm!
I added two improvements:

1. There are now three options: domain, static ip or no servername ("localhost")
2. Certbot will fail if port 80 of the RPi is not reachable. I've added a warning

I think we have reached a finished version of your feature. Amazing!! 🎆 🎉 I'll write a changelog posting the next few days.

ThomDietrich:
Okay I've one last question, why this complicated?
```
$ cat /etc/cron.d/certrenew
24 7-23/12 * * * root certbot renew --quiet
```

BClark09:
Nice additions! Certbot recommend performing an update check twice a day.  But let's encrypt ask that cron expressions are applied at a random time.  Since we're providing the cron,  then we wouldn't want to hardcode a timestamp,  otherwise a mass of raspberry pis will try to update at the same time.

ThomDietrich:
I got the idea, I was just confused about the twice a day because last time I checked a certificate is valid three months and it's safe to check once per week... Seems this was updated https://certbot.eff.org/all-instructions/#debian-8-jessie-nginx

BClark09:
Its just to stop systems being unavailable. 

Reading the docs again now,  I see that they do this with a debian install anyway.  We should probably our cron in this case from our script.

ThomDietrich:
Oh, it does indeed
```shell
[11:12:12] pi@openHABianPiW66a:~$ cat /etc/cron.d/certbot
# /etc/cron.d/certbot: crontab entries for the certbot package
#
# Upstream recommends attempting renewal twice a day
#
# Eventually, this will be an opportunity to validate certificates
# haven't been revoked, etc.  Renewal will only occur if expiration
# is within 30 days.
SHELL=/bin/sh
PATH=/usr/local/sbin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin

0 */12 * * * root test -x /usr/bin/certbot && perl -e 'sleep int(rand(3600))' && certbot -q renew

[11:13:43] pi@openHABianPiW66a:~$ cat /etc/cron.d/certrenew
24 7-23/12 * * * root certbot renew --quiet
```

PR #30, opened 2016-11-02
Remove certbot renewal from setup script

BClark09:
This is now already done when installing ngnix on Debian Jessie.


ThomDietrich:
:+1:

Issue #52, opened 2016-12-18
[Feature request] One click password changer


EliasGabrielsson:
I miss a menu which could handle all password changes for the system.

* pi user -> `passwd`
* Samba user, openhab -> ?
* HTTP basic authentication for nginx HTTPS setup.

I also wonder if it would be easier for the end user if the Samba account (openhab) and local admin (pi) would have the same username, eg. `openhab-admin`. Then there could be one "set admin password promt" which changes both passwords.

ThomDietrich:
Hey Elias,
I've been thinking about that too. I guess your idea came after you saw the "Please change the default password" dialog, which is actually only a few days old. The possibility to change all passwords right from the menu would be nice indeed.

* system users (pi) -> `passwd`
* samba users (openhab) -> `smbpasswd` ([Example](https://github.com/openhab/openhabian/blob/master/openhabian-setup.sh#L285-L286))
* karaf ssh login -> menu entry "06 | Karaf Console"
* nginx HTTP(S) auth -> `htpasswd` ([Example](https://github.com/openhab/openhabian/blob/master/openhabian-setup.sh#L877))

Regarding the samba user: that would be possible. Would probably make a few things easier. I will need to look into this. In the end I would like to see one menu entry to change all passwords. 

Do you want to help in any of these parts?

EliasGabrielsson:
I can look into it and see what I could manage to come up with. I am currently working to get a install script for some telldus parts as well.

ThomDietrich:
Thanks again!

PR #55, opened 2016-12-26
[RFC] Issue:#52 Password change function

EliasGabrielsson:
First draft for a password change function. See commit msg for overview.
Some more tweaking my be needed, especially to make output more pretty.

Notice that the pull repo in `post-install.txt` need to be changed from `openhab` to `EliasGabrielsson` to get the right code when installing a new instance.

### DIFF
--- a/includes/rc.local
+++ b/includes/rc.local
@@ -25,7 +25,7 @@ if [ ! -f $FLAG ]; then
   systemctl enable openhab2.service
   systemctl start openhab2.service
 
-  if [ -f "/home/pi/openHABian-install-failed" ]; then
+  if [ -f "/home/`ls /home`/openHABian-install-failed" ]; then
     echo "[openHABian] Indicating failure through green LED fast blinking." >> $LOG
     modprobe ledtrig_timer
     echo "timer" >/sys/class/leds/led0/trigger
--- a/installer-config.txt
+++ b/installer-config.txt
@@ -8,7 +8,7 @@ userpw=raspberry
 boot_volume_label=openHABian
 disable_root=1
 user_is_admin=1
-username=pi
+username=admin
 
 # Language and time
 timeserver=0.pool.ntp.org
--- a/openhabian-setup.sh
+++ b/openhabian-setup.sh
@@ -45,6 +45,25 @@ else
   echo "OK"
 fi
 
+# Load installer configuration
+if [ -a $SCRIPTDIR/installer-config.txt.netinstalled ]; then # When installed from raspbian-ua-netinst
+  source $SCRIPTDIR/installer-config.txt.netinstalled
+  NETINST=1
+elif [ -a $SCRIPTDIR/installer-config.txt.installed ]; then # Non raspbian installation and fallback
+  source $SCRIPTDIR/installer-config.txt.installed
+else # Make a copy for non raspbian-ua-netinst
+  cp $SCRIPTDIR/installer-config.txt $SCRIPTDIR/installer-config.txt.installed
+  if (whiptail --title "Administrative user" --yesno "This installer script is designed to bind to an administrative system user. Shall the script use current user: $SUDO_USER ?" 8 78) then
+    sed -i "/username=.*/c\username=$SUDO_USER" installer-config.txt.installed
+  else
+      whiptail --title "Correct user required" --msgbox "Please rerun this script while logged on with the correct user" 12 60
+      rm -f $SCRIPTDIR/installer-config.txt.installed
+      echo "[openHABian] Script canceled by user" 1>&2
+      exit 1
+  fi
+  source $SCRIPTDIR/installer-config.txt.installed
+fi
+
 # script will be called with 'unattended' argument by post-install.txt
 if [[ "$1" = "unattended" ]]
 then
@@ -91,6 +110,10 @@ is_pitwo() {
 is_pithree() {
   grep -q "^Revision\s*:\s*[ 123][0-9a-fA-F][0-9a-fA-F][0-9a-fA-F]08[0-9a-fA-F]$" /proc/cpuinfo
   return $?
+}
+is_pi() {
+  if is_pizero || is_pione || is_pitwo || is_pithree; then return 1; fi
+  return 0
 }
 
 calc_wt_size() {
@@ -153,14 +176,23 @@ basic_packages() {
   if [ $? -eq 0 ]; then echo "OK"; else echo "FAILED"; exit 1; fi
 }
 
-needed_packages() {
+# Packages relevant to platforms
+needed_packages_generic() {
   # install raspi-config - configuration tool for the Raspberry Pi + Raspbian
   # install apt-transport-https - update packages through https repository (https://openhab.ci.cloudbees.com/...)
   # install samba - network sharing
   # install bc + sysstat - needed for FireMotD
   # install avahi-daemon - hostname based discovery on local networks
   echo -n "[openHABian] Installing additional needed packages... "
-  cond_redirect apt -y install raspi-config oracle-java8-jdk apt-transport-https samba bc sysstat avahi-daemon
+  cond_redirect apt -y install oracle-java8-jdk apt-transport-https samba bc sysstat avahi-daemon
+  if [ $? -eq 0 ]; then echo "OK"; else echo "FAILED"; exit 1; fi
+}
+
+# Pi specific packages
+needed_packages_raspberry() {
+  # install raspi-config - configuration tool for the Raspberry Pi + Raspbian
+  echo -n "[openHABian] Installing additional raspberry specific packages... "
+  cond_redirect apt -y install raspi-config
   if [ $? -eq 0 ]; then echo "OK"; else echo "FAILED"; exit 1; fi
 }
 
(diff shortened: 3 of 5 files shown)


EliasGabrielsson:
Nginx is deliberately not dealt with here because it's likely to have different passwords and username for manage openhab and use it as a user. Next step would be to make a menu 31 or similar which could manage nginx users.

EliasGabrielsson:
Note to self: `readme.md` need to be updated to reflect new username & password before merge.

ThomDietrich:
@BClark09 @jupe76 Please ignore the mention-bot. Including him was a test ;)
@EliasGabrielsson Great contribution, I like the new ideas you came up with and everything looks well thought through. Give me a couple of days to get back at you with a few comments ;)

EliasGabrielsson:
I will squash commits together before final merge.

ThomDietrich:
Hello Elias, sorry I'm getting back to you only now. I've been distracted more than I expected over the holidays.
I've just started to review your PR. Sadly I had to stop after 5-6 comments. I recently read an article by a full-time open source maintainer, who stated "I barely accept PRs bigger than 80 lines". Looking at your PR I have no other choice but to agree. 😅 
Here's what I would like to suggest. Please create a new PR only submitting the password change routine and nothing else. We can discuss this one, merge, and move on from there. I hope this is okay for you!?

EliasGabrielsson:
Hi,
Checked out my PR today and have missed that my shuffling of some code to make it more readable made  a lot of removes and adds, not my intention at all. Sorry for that.

I reverted to the old order, I would say more chaotic for a new reader but easier to review for sure! 
(Let's discuss code chunks in another issue as time permits. I think it's make sense to split things apart when non pi users want to use the post install scripts.)

Anyway I can't give you a 80 line PR because user accounts and passwords is so integrated in all parts of the script  but I would say it's not as "huge" as it looked before. Mainly a new function and some rewrites of the current ones. ( eg. replace hardcoded `pi` with dynamic code.) 

Have a good one!

ThomDietrich:
Okay great, looks much better now. Thanks @EliasGabrielsson 
There are just a few things I'm not quite happy with, let me do a full review later today.

openhabian-setup.sh
```diff
@@ -327,42 +359,39 @@ misc_system_settings() {
...
   echo "OK"
 }
 
-openhab_shell_interfaces() {
+change_karaf_default_username(){
+  echo -n "[openHABian] Changing karaf user from \"openhab\" to \"$username\"... "
+  # Enable encryption of password within openhab2 karaf console
+  cond_redirect sed -i "s/encryption.enabled = false/encryption.enabled = true/g" /var/lib/openhab2/etc/org.apache.karaf.jaas.cfg
+  cond_redirect sed -i "s/encryption.name =/encryption.name = basic/g" /var/lib/openhab2/etc/org.apache.karaf.jaas.cfg
+  cond_redirect sed -i "s/encryption.algorithm = MD5/encryption.algorithm = SHA-256/g" /var/lib/openhab2/etc/org.apache.karaf.jaas.cfg
```
ThomDietrich: This should be proposed at OH/ESH, not here

post-install.txt
```diff
@@ -12,13 +12,17 @@ chroot /rootfs /usr/bin/git clone -b master https://github.com/openhab/openhabia
 if [ $? -eq 0 ]; then echo "OK"; else echo "FAILED"; exit 1; fi
 ln -s /opt/openhabian/openhabian-setup.sh /rootfs/usr/local/bin/openhabian-config
 
-echo "[openHABian] === Start executing 'openhabian-setup.sh' ==="
+echo -n "[openHABian] Backup installation configuration..."
+cp /rootfs/boot/installer-config.txt /rootfs/opt/openhabian/installer-config.txt.netinstalled
```
ThomDietrich: This needs to be done in your function above, otherwise an existing system will be without config.

post-install.txt
```diff
@@ -12,13 +12,17 @@ chroot /rootfs /usr/bin/git clone -b master https://github.com/openhab/openhabia
...
+echo -n "[openHABian] === Start executing 'openhabian-setup.sh' ==="
 chroot /rootfs /bin/bash /opt/openhabian/openhabian-setup.sh unattended
 if [ $? -eq 0 ]; then
-  rm /rootfs/home/pi/openHABian-install-failed
-  touch /rootfs/home/pi/openHABian-install-successful
+  rm /rootfs/home/`ls /rootfs/home`/openHABian-install-failed
+  touch /rootfs/home/`ls /rootfs/home`/openHABian-install-successful
 else
-  touch /rootfs/home/pi/openHABian-install-failed
+  touch /rootfs/home/`ls /rootfs/home`/openHABian-install-failed
```
ThomDietrich: This file is only used by raspian-ua-netinst. These changes are not needed

installer-config.txt
```diff
@@ -8,7 +8,7 @@ userpw=raspberry
 boot_volume_label=openHABian
 disable_root=1
 user_is_admin=1
-username=pi
+username=admin
```
BClark09: Also, "admin" is the most targeted username for brute-force attempts, much more so than "Pi".

ThomDietrich: True that ;)

EliasGabrielsson: It doesn't matter for me.  Changed from `pi` because a) it is standard for RPi and more prone for brute-forces but as BClark09 mention `admin` is not that good either. b) would be in my opinion illogical to have username `pi` for managing openhab.

But in the end of the day it's just preferences. The ones (like me) who wants a different name can just change it.

openhabian-setup.sh
EliasGabrielsson: Does ESH even use Karaf?
Is not that a part of OH?

I also wonder if it's really a OH-distro "responsibility" to manage this kind of security. Hardening of a production system is more of a post install issue.

For me personally I would love to have it in the distro but what is the practice of the project?
Which repo does this code belong to, openhab-distro?

@kaikreuzer

kaikreuzer: Is there any drawback of these settings? If not, why not harden the setup for everybody (instead of treating it as a post install issue).
The Karaf files of the distro are maintained here: https://github.com/openhab/openhab-distro/tree/master/distributions/openhab/src/main/resources/userdata/etc

post-install.txt
EliasGabrielsson: Already taken care of. The script differ native installed systems (`installer-config.txt.netinstalled`) from post install systems (`installer-config.txt.installed`) with tool only. 

This gives the capability to only check the default_password for net installed systems and force the user to add a username for folders when runned as a script only.

post-install.txt
EliasGabrielsson: But what if the username changes which this PR implements then we will have an regression bug in the codebase.

openhabian-setup.sh
EliasGabrielsson: I don't see any drawbacks. I will add above patch set to `openhab-distro` and remove these parts from this PR.

post-install.txt
EliasGabrielsson: or is this this moved to `/rootfs/opt/openHABian-install-failed`?

includes/rc.local
```diff
@@ -25,7 +25,7 @@ if [ ! -f $FLAG ]; then
   systemctl enable openhab2.service
   systemctl start openhab2.service
 
-  if [ -f "/home/pi/openHABian-install-failed" ]; then
+  if [ -f "/home/`ls /home`/openHABian-install-failed" ]; then
```
EliasGabrielsson: Do you mean it will be changed to: `/rootfs/opt/openHABian-install-failed`?

EliasGabrielsson:
@ThomDietrich So have some parts to clear out before I can refactor for next iteration. See comments on comments.

includes/rc.local
ThomDietrich: Not for now. I will look into that with a later PR. We should really stay with "Password change function" in this one. ;)

installer-config.txt
ThomDietrich: Exactly. You've introduced the non-static username use, which is the important part 👍

openhabian-setup.sh
ThomDietrich: The PR got merged already. I like this solution way better ;)

post-install.txt
ThomDietrich: Yes but you are missing the point, that an already installed system pulling this PRs results will not own a `/opt/openhabian/installer-config.txt.netinstalled`. Hence please move this as a conditional action inside your other routine.

post-install.txt
ThomDietrich: Yes but later, as stated before.

post-install.txt
ThomDietrich: Thinking one step further: You need to create this copy outside the openhabian folder. This folder is git controlled and cleaned on every `sudo openhabian-config` -> Update.

I would also name the file differently. How about `/opt/openhabian.conf` or even `/etc/openhabian.conf` I would not use two different filenames. Just store whatever information you need inside the config, that's what it is for.

ThomDietrich:
Hey @EliasGabrielsson, I've created a few more comments. Do you now get why I asked you to create smaller PRs ? ;) In fact I would still ask you to do so. Most of my comments are related to the username/config file idea. Would you outsource that? It's not related to Password change anyhow and I didn't touch that big chunk yet.

post-install.txt
EliasGabrielsson: Good points, understand what you mean now.
I think it would make more sense to store the configuration file in `/etc`so it will be backed up if the user choose to use `etckeeper`.

EliasGabrielsson:
@ThomDietrich
Sorry for not reached back for a while have not found time to complete this.
Looks like this will be on hold for another week or two, time will tell.

PR #102, opened 2017-02-24
Fix typo and improve serial port function

SuneRadich:
The if should test against 'is_pi', and not is pi

Because of this, part of the Serial configuration to make Serial work with a Razberry does not work


ThomDietrich:
Of course, a mistake from just a few days ago. Thanks for helping out!

Does this solve your problem here: https://community.openhab.org/t/rpi3-razbeery-issues/23768 ?

ThomDietrich:
Another thing you criticized is, that the selection should represent the current state of your system. I can look into that today. You are the second one to ask about that.

ThomDietrich:
@SuneRadich I've added the detection of current settings. Could you have a look?

Issue #113, opened 2017-03-16
Link/mount all openHAB folders in one place


ThomDietrich:
for a better overview and backups

https://community.openhab.org/t/openhabian-hassle-free-rpi-image/13379/793?u=thomdietrich

PR #120, opened 2017-03-21
openHAB bind mounts, Samba improvements

ThomDietrich:
Make accessing the RPi files easier:
* All openHAB folders are now bind mounted subdirectories to `/srv/` #113 
* Folder includes README for clarity
* The folder is shared via samba
* Samba now also shares the user home
* Widened the write permissions of the user to everything besides /usr/share/openhab2 (user shouldn't need that)
* Changed a few samba details while I was at it

Also in here:
* Added an empty ~/.ssh/authorized_keys for ease of use
* Security: The initial user password is now removed from openhabian.conf after setup
* Hostname function is now also called during setup, in case the user set a hostname via config file beforehand

### DIFF
--- a/includes/smb.conf
+++ b/includes/smb.conf
@@ -26,18 +26,18 @@
 ## Browsing/Identification ###
 
 # Change this to the workgroup/NT-domain name your Samba server will part of
-   workgroup = WORKGROUP
+  workgroup = WORKGROUP
 
 # Windows Internet Name Serving Support Section:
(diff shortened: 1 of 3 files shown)


ThomDietrich:
Test driving now.

Comments would be appreciated...

ThomDietrich:
Everything looks great.

Permissions might still be a problem for freshly created files by openhab. We need group write access. The only solution I see here is setting umask for openhab. Wdyt?

Issue #168, opened 2017-06-30
System root not on SD card?


hr3:
I setup the SD-Card with Etcher and the actual image. The USB-HDD is ready, but "Move" does't work. There are no other USB-devices.
```
###############################################################################
################# openHABianPi #################################### 7.05.170509
###############################################################################
##        Ip = 192.168.178.33
##   Release = Raspbian GNU/Linux 8 (jessie)
##    Kernel = Linux 4.9.28+
##  Platform = Raspberry Pi Model B Rev 2
##    Uptime = 0 day(s). 0:16:50
## CPU Usage = 0 % avg over 2 cpu(s)  (1 core(s) x 1 socket(s))
##  CPU Load = 1m: 0.47, 5m: 0.35, 15m: 0.44
##    Memory = Free: 0.19GB (40%), Used: 0.27GB (60%), Total: 0.47GB
##      Swap = Free: 0.09GB (100%), Used: 0.00GB (0%), Total: 0.09GB
##      Root = Free: 12.36GB (88%), Used: 1.62GB (12%), Total: 14.61GB
##   Updates = 0 apt-get updates available.
##  Sessions = 1 sessions
## Processes = 93 running processes of 32768 maximum processes
###############################################################################

              Welcome to            __  _____    ____  _
            ____  ____  ___  ____  / / / /   |  / __ )(_)___ _____
           / __ \/ __ \/ _ \/ __ \/ /_/ / /| | / __  / / __ `/ __ \
          / /_/ / /_/ /  __/ / / / __  / ___ |/ /_/ / / /_/ / / / /
          \____/ .___/\___/_/ /_/_/ /_/_/  |_/_____/_/\__,_/_/ /_/
              /_/
                  openHAB 2.1.0-1 (Build )
```
```
┌────────────────────┤ Move system root to '/dev/sda1' ├─────────────────────┐
│                                                                            │
│ DANGEROUS OPERATION, USE WITH PRECAUTION!                                  │
│                                                                            │
│ This will move your system root from your SD card to a USB device like an  │
│ SSD or a USB stick to reduce wear and failure or the SD card.              │
│                                                                            │
│ 1.) Make a backup of your SD card                                          │
│ 2.) Remove all USB massstorage devices from your Pi                        │
│ 3.) Insert the USB device to be used for the new system root.              │
│     THIS DEVICE WILL BE FULLY DELETED                                      │
│                                                                            │
│                                                                            │
│ Do you want to continue on your own risk?                                  │
[...]

PR #178, opened 2017-07-09
fixed "move system root to USB" when PARTUUID is used

jupe76:
* fixed "move system root to USB" when PARTUUID is used

* closes [#168](https://github.com/openhab/openhabian/issues/168) 

Signed-off-by: Peter Juenger <peter.juenger@web.de>


ThomDietrich:
Excellent fix! Seems like you thought of everything ;)

PR #201, opened 2017-08-17
Make all shell scripts executable.

uwehermann:
Signed-off-by: Uwe Hermann <uwe@hermann-uwe.de>


ThomDietrich:
Hey! Thanks for the contribution! 👏

Fun fact: I've deliberately held those scripts down on non-executable because I didn't want to confuse the normal user or encourage meaningless actions...

uwehermann:
Alright, in that case the patch can be ignored.

ThomDietrich:
I think that's for the better. Thanks. Feel free to drop by again later ;)

PR #223, opened 2017-10-28
Amanda docs and install update

mstormi:
based on user feedback

### DIFF
--- a/openhabian-setup.sh
+++ b/openhabian-setup.sh
@@ -1473,14 +1473,14 @@ amanda_setup() {
 
   if [ -n "$INTERACTIVE" ]; then
     if (whiptail --title "Create Amazon S3 based backup" --yes-button "Yes" --no-button "No" --yesno "Setup a backup mechanism based on Amazon Web Services. You can get 5 GB of S3 cloud storage for free on https://aws.amazon.com/. See also http://wiki.zmanda.com/index.php/How_To:Backup_to_Amazon_S3" 15 80) then
-        config=openhab-AWS
-        S3accesskey=$(whiptail --title "S3 access key" --inputbox "Enter the S3 access key you obtained at S3 setup time:" 10 60 3>&1 1>&2 2>&3)
-        S3secretkey=$(whiptail --title "S3 secret key" --inputbox "Enter the S3 secret key you obtained at S3 setup time:" 10 60 3>&1 1>&2 2>&3)
-	tapes=15
-	capacity=$(whiptail --title "Storage capacity" --inputbox "How much storage do you want to dedicate to your backup in megabytes ? Recommendation: 2-3 times the amount of data to be backed up." 10 60 3>&1 1>&2 2>&3)
-	let size=${capacity}/${tapes}
+      config=openhab-AWS
+      S3accesskey=$(whiptail --title "S3 access key" --inputbox "Enter the S3 access key you obtained at S3 setup time:" 10 60 3>&1 1>&2 2>&3)
+      S3secretkey=$(whiptail --title "S3 secret key" --inputbox "Enter the S3 secret key you obtained at S3 setup time:" 10 60 3>&1 1>&2 2>&3)
+      tapes=15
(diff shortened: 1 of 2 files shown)


mstormi:

mstormi:
@ThomDietrich 
most of this is docs changes only
but pls have a quick look at the code changes and merge this

openhabian-setup.sh
```diff
@@ -1439,6 +1441,8 @@ amanda_setup() {
             fi
       done
   fi
+  
+  /usr/sbin/usermod -a -G backup openhabian
   /usr/sbin/chpasswd <<< "${backupuser}:${password}"
   /usr/bin/chsh -s /bin/bash ${backupuser}
```
ThomDietrich: You didn't change that last line but could you please explain the need for it?

openhabian-setup.sh
```diff
@@ -1470,8 +1476,9 @@ amanda_setup() {
         config=openhab-AWS
         S3accesskey=$(whiptail --title "S3 access key" --inputbox "Enter the S3 access key you obtained at S3 setup time:" 10 60 3>&1 1>&2 2>&3)
         S3secretkey=$(whiptail --title "S3 secret key" --inputbox "Enter the S3 secret key you obtained at S3 setup time:" 10 60 3>&1 1>&2 2>&3)
-        tapes=$(whiptail --title "Number of virtual storage containers in rotation" --inputbox "How many virtual containers will you setup inside the S3 bucket ?  Note that #container x container size will need to fit into your S3 bucket." 10 60 3>&1 1>&2 2>&3)
-        size=$(whiptail --title "Container size" --inputbox "How large do you want one virtual container to be ? Specify the size in megabytes." 10 60 3>&1 1>&2 2>&3)
+	tapes=15
+	capacity=$(whiptail --title "Storage capacity" --inputbox "How much storage do you want to dedicate to your backup in megabytes ? Recommendation: 2-3 times the amount of data to be backed up." 10 60 3>&1 1>&2 2>&3)
+	let size=${capacity}/${tapes}
```
ThomDietrich: You are mixing tabs and spaces here. Please correct that.
For the future I'd recommended to use an editor that is capable to handle whitespaces properly.

ThomDietrich:
Hey @mstormi, here's a short review. The README looks good. On a more general level I still think it consists of too much running text. I'd suggest to work with paragraphs, sub-headlines, enumerations and similar to easy the reading experience...

openhabian-setup.sh
mstormi: which line?
chsh ? User needs to have a proper shell.
usermod? User can use default openhabian user to see Amanda config, browse storage

openhabian-setup.sh
mstormi: done

openhabian-setup.sh
ThomDietrich: I mean the chsh line. Why exactly does a non-interactive system account need a shell? That's normally not needed or if it is, other solutions are available.

mstormi: You need to switch to the backup user in case of restore, so while you normally wouldn't use it interactively, in this case you need to.

ThomDietrich: I see. I don't particularly like that we are exposing yet another user+password combination via SSH. Wdyt?

ThomDietrich:
There still many tabs left. Could you please have another look? Besides that I'm ready to merge.

grafik

openhabian-setup.sh
mstormi: Well, Amanda requires this, so little chance to work around that without breaking anything.
If you're truely worried, you can add a 'DenyUsers backup' line to /etc/sshd/sshd_config.

ThomDietrich: Well, wether I am worried or not is not the point. The system should still be secure for the end user. For now I'd be happy if you simply mentioned that "the password has to be safe because it will be able to log in remotely" in the enter password dialog.

ThomDietrich:
That's it!
Btw. Signing off is something that should ideally/normally happen per commit. To cope with problems of beginners it's acceptable to make an exception and put it into a comment in the PR thread. I realized it a few seconds too late but please sign every commit from here on out. Most git clients (like SmartGit) provide an option to do that automatically, on the console you can use the "git commit -s" option.

Thanks for your continued contribution!

PR #237, opened 2017-11-22
FIND install: extract ALL required files

### LINKED ISSUE #236: Find server installation, data and templates folders missing
I tried the Find server install option yesterday.
The script ran through successfully, but the findserver did not start. The service was in failed state.
Further investigation pointed to missing data and templates folders. I checked in the install zip and these should indeed be in the current directory when executing the find server. Copying these from the zip file to the /var/lib/findserver directory solved this issue. However, I am not sure these should be there. Maybe there is an environment variable to put it in a more appropriate place.
I believe the install script should be extended to put these directories and file in the appropriate place.

mstormi:
Fixes #236 

Signed-off-by: Markus Storm <markus.storm@gmx.net> (github: mstormi)

### DIFF
--- a/openhabian-setup.sh
+++ b/openhabian-setup.sh
@@ -1092,7 +1092,7 @@ HABian GitHub issue."
   /bin/sed -e "s|%MQTTSERVER|${MQTTSERVER}|g" -e "s|%MQTTPORT|${MQTTPORT}|g" -e "s|%FINDPORT|${FINDPORT}|g" -e "s|%FINDSERVER|${FINDSERVER}|g" ${SCRIPTDIR}/includes/findserver >${FIND_DEFAULT}
 
 
-  /bin/rm -f ${FIND_TMP} ${CLIENT_TMP} ${FIND_DSTDIR}/findserver
+  /bin/rm -f ${FIND_TMP} ${CLIENT_TMP}
   cond_redirect /bin/systemctl enable findserver.service || FAILED=1
   cond_redirect /bin/systemctl restart findserver.service || FAILED=1
   if [ $? -eq 0 ]; then echo "OK"; else echo "FAILED"; exit 1; fi
--- a/openhabian-setup.sh
+++ b/openhabian-setup.sh
@@ -1074,7 +1074,7 @@ HABian GitHub issue."
   /usr/bin/wget -O ${CLIENT_TMP} ${CLIENT_SRC}
   /usr/bin/unzip ${FIND_TMP} findserver -d /usr/sbin
   /usr/bin/unzip ${FIND_TMP} fingerprint -d /usr/sbin
-  /usr/bin/unzip ${FIND_TMP} static\* -d ${FIND_DSTDIR}
+  /usr/bin/unzip ${FIND_TMP} -d ${FIND_DSTDIR}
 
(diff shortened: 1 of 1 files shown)


mstormi:
not sure why this LF breaker patch is about to be committed again ? It's supposed to be in since 2 days.

openhabian-setup.sh
```diff
@@ -1060,39 +1060,38 @@ HABian GitHub issue."
...
+    FINDADMINPASS=$(whiptail --title "findserver MQTT Setup" --passwordbox "Enter a password for the FIND admin user on your MQTT broker:" 15 80 $DEFAULTFINDPASS 3>&1 1>&2 2>&3)
     cond_redirect /usr/bin/mosquitto_passwd -b ${MOSQUITTO_PASSWD} ${FINDADMIN} ${FINDADMINPASS} || FAILED=1
     if [ $? -ne 0 ]; then echo "FAILED"; exit 1; fi
   fi
   /bin/sed -e "s|%MQTTSERVER|${MQTTSERVER}|g" -e "s|%MQTTPORT|${MQTTPORT}|g" -e "s|%FINDADMIN|${FINDADMIN}|g" -e "s|%FINDADMINPASS|${FINDADMIN_PASS}|g" -e "s|%FINDPORT|${FINDPORT}|g" -e "s|%FINDSERVER|${FINDSERVER}|g" ${SCRIPTDIR}/includes/findserver.service >${FIND_SYSTEMCTL}
   /bin/sed -e "s|%MQTTSERVER|${MQTTSERVER}|g" -e "s|%MQTTPORT|${MQTTPORT}|g" -e "s|%FINDPORT|${FINDPORT}|g" -e "s|%FINDSERVER|${FINDSERVER}|g" ${SCRIPTDIR}/includes/findserver >${FIND_DEFAULT}
 
 
-  /bin/rm -f ${FIND_TMP} ${CLIENT_TMP}
+  /bin/rm -f ${FIND_TMP} ${CLIENT_TMP} ${FIND_DSTDIR}/findserver
```
ThomDietrich: Also "fingerprint" ?

ThomDietrich:
Maybe because your repo master is still on the old revision

openhabian-setup.sh
mstormi: no

ThomDietrich: But then it's available twice.
May I suggest something instead of the above. That is how I'd do something like this:

```shell
unzip ${FIND_TMP} -d ${FIND_DSTDIR}
ln -s ${FIND_DSTDIR}/findserver /usr/sbin/findserver
ln -s ${FIND_DSTDIR}/fingerprint /usr/sbin/fingerprint
```

mstormi: no, fingerprint isn't contained in the archive.
Then again, $FIND_TMP  in the unzip line is bad, it must read $CLIENT_TMP
Ok, will change that and do a ln for the findserver binary

ThomDietrich: Ah sorry, didn't see that detail. Now remove the `/bin/rm -f ${FIND_DSTDIR}/findserver` and we are good to go ;)

mstormi: here you are

ThomDietrich:
Looks good! Thanks!

PR #241, opened 2017-11-26
Break down monolithic  script file into sourced function files

### LINKED ISSUE #225: Split up single script file
To this point the file `openhabian-setup.sh` is one big collection of functions. It has grown out of it's size since long ago and I believe it is time to split it up into smaller junks.

* Menus
* Helper functions
* Essential openHAB related functions
* Optional components
* ...

It would also be interesting to make some script files / functions (e.g. the Nginx reverse proxy) executable on their own.

Any comments or ideas we should consider during that step?

ThomDietrich:
This PR basically outsources the numerous functions available in the main file `openhabian-setup.sh` into different files, like `openhab.sh` with all openHAB specific functions. These files are then re-included using the `source` command. The individual script files are not intended to be executed on their own (even though it would now be easier to accomplish that).

@BClark09 your opinion would be appreciated.

Resolves #225 


BClark09:
Looks good! I won't be able to test it until later but I don't see why it wouldn't work as is. 

I like the simplicity of including all files in the function folder within the loop and this makes sense to me for unattended installs, but is there a necessity to source it all at once before the whiptail menu? Can't the relevant file be sourced just before the function is called?

ThomDietrich:
I've tested it and everything worked so far.
Thanks for looking over it ;)


> Can't the relevant file be sourced just before the function is called?

Excellent question... That would be quite clever, I thought about that some more and still ask myself "Why?" 😄 Each file solely contains functions and no implicitly executed logic. Hence the only reason would be some kind of resource concern and I doubt we are anywhere near critical sizes... wdyt?

ThomDietrich:
@BClark09 I need these changes for further tests but please add your comment and we can discuss a subsequent PR

BClark09:
You have a good point, I think the limited resource saving wouldn't justify the effort needed to implement it.

Issue #188, opened 2017-07-26
[Proposal] Integrate Yeoman openHAB generator


kubawolanin:
This suggestion is a shameless plug of mine :smile: 

Since we have a Node.js in place, we could integrate my [Yeoman openHAB generator](https://community.openhab.org/t/openhab-generator-design-your-smart-home-in-no-time/32250) into openHABian. It'll allow the newcomers to start off easily.

Here's the repository:
https://github.com/kubawolanin/generator-openhab

@ThomDietrich thoughts?

ThomDietrich:
> This suggestion is a shameless plug of mine :smile:

:astonished: how could you!? :smile: Sure thing. Looks pretty neat!

Looking at the result of your generator I think we should seriously consider defining an openHAB style guide for names and config files. I've already mentioned that over in the vscode repo. Your generator result looks quite similar to how I'd organize my items but the mixture with groups and the intention is quite strange to me (this is an example and not bad). On the other side [I am a big fan of table style indention](https://github.com/ThomDietrich/openhab-config/blob/master/items/homematic-heizung.items) and would love to see a vscode beautify extension with it one day.

kubawolanin:
I've been following the following format:

* `[Room]_[Device]` if a house has just one floor
* `[GF|FF|F2|F3...]_[Room]_[Device]` for multi-floor house ;-)

> On the other side I am a big fan of table style indention and would love to see a vscode beautify extension with it one day.

Hey, just check out [Alignment](https://marketplace.visualstudio.com/items?itemName=annsk.alignment) plugin to VSCode.
By simply selecting your text block and hitting `Alt + -` it transforms this:
```
Group Bedroom "Bedroom" <bedroom> (Home)
    Switch Bedroom_Light "Light" <light> (Bedroom, Lights) { channel="" }
    Contact Bedroom_Window "Window" <window> (Bedroom, Windows) { channel="" }
    Contact Bedroom_Door "Door" <door> (Bedroom, Doors) { channel="" }
    Rollershutter Bedroom_Shutter "Shutter" <rollershutter> (Bedroom, Shutters) { channel="" }
```

To this:
```
Group Bedroom       "Bedroom"       <bedroom> (Home)
      Switch        Bedroom_Light   "Light"   <light>         (Bedroom, Lights)   { channel="" }
      Contact       Bedroom_Window  "Window"  <window>        (Bedroom, Windows)  { channel="" }
[...]

PR #243, opened 2017-11-26
Fix multiple issues

ThomDietrich:
@kubawolanin I've removed the yo-generator as one of the apps installed by default. Was that correct?


kubawolanin:
Hey Thomas,
Sure!
I think we can soon ditch it altogether when [HomeBuilder](https://github.com/openhab/openhab-core/pull/172) will find its way to master branch ;-)

Thanks for the heads up.

kubawolanin:
BTW

>Was that correct?

https://github.com/openhab/openhabian/issues/188#issuecomment-318068954 😉

ThomDietrich:
I know, I know. That's why I've asked you. I also think that with the HomeBuilder on the verge, we don't need to keep the generator as a per-default component and I was in the process of cleaning up :) Best!

PR #261, opened 2017-12-10
Add automatic update check, update and restart

ThomDietrich:
A PR I wanted to work out for a long time. It contains two things. First a check while starting the configuration tool to find out if updates (commits) are available, secondly a query if the update should be executed and finally I've worked around the need to restart the script manually after the update.


ThomDietrich:
grafik

PR #276, opened 2017-12-14
Update nodejs-apps.sh

Wolfmanxxx:
Removed nodejs_setup from nodered_setup() as line 41 installs ARMv7 or ARMv6 depending on your ARM version.
Not tested 100% but worked for me on both pi 0 and pi 3

### DIFF
--- a/functions/nodejs-apps.sh
+++ b/functions/nodejs-apps.sh
@@ -35,7 +35,6 @@ frontail_setup() {
 }
 
 nodered_setup() {
-  nodejs_setup
   echo -n "$(timestamp) [openHABian] Installing Node-RED... "
   FAILED=0
   cond_redirect wget -O /tmp/update-nodejs-and-nodered.sh https://raw.githubusercontent.com/node-red/raspbian-deb-package/master/resources/update-nodejs-and-nodered || FAILED=1


Blakko:
I confirm this works on a raspberry model b too. 
But i consider it a workaround (don't get me wrong, it's better than not working :P), I guess we'll end up with the same error if a user tries to install the log viewer first.
Maybe we could tweak the code in the `nodejs_setup` function and use the same one node-red is using during the install

ThomDietrich:
ah.
Yeah that's not a solution... We need a solution that works for nodejs in general. I knew that the node-red installer does that but that is not an option as we are going to have more and more node applications. It shouldn't be too complicated.

1. Use one of the helper functions to identify the system: https://github.com/openhab/openhabian/blob/master/functions/helpers.sh
2. Execute a different setup routine for ARMv6 hardware

Would you (or @Blakko) be up to the task? Either way thanks for your contribution!!

Blakko:
I'll try to do some tests on the following days, but it's my first time on this kind of scripts

ThomDietrich:
Great!
It should be rather easy to get going. All functions are small and you can derive your solution from the existing ones.

ThomDietrich:
Hey guys! With #277 merged, the question remains if this PR here should still be merged. It seems like Node-RED will check, install or update Node.js anyhow...

Blakko:
Yes, Node-Red will install it anyway so it should be safe to just remove the function call

ThomDietrich:
👍 here we go.

Thanks @Wolfmanxxx and @Blakko for adding value to openHABian for us all!

PR #284, opened 2017-12-17
Rename openhablog aliases

ThomDietrich:
Whoever reads this, wdyt?

Signed-off-by: Thomas Dietrich <Thomas.Dietrich@tu-ilmenau.de>

### DIFF
--- a/includes/bash.bashrc
+++ b/includes/bash.bashrc
@@ -92,9 +92,9 @@ alias ..='cd ..'
 alias ...='cd ../..'
 alias ....='cd ../../..'
 
-alias openhablog='tail -n 100 -f /var/log/openhab2/events.log -f /var/log/openhab2/openhab.log'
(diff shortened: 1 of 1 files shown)


mstormi:
Didn't know those, and feel to use numbers to distinguish isn't intuitive, so it's a reasonable change.
I would shorten it, though (oh instead of openhab)
Also, AFAIK multitail does not follow filename but inode, so if logs rotate, tail'ing is stuck, or am I mistaken there ? I'm using tail -F (instead of -f) for that purpose.

ThomDietrich:
I thought so too ;)
I'm not so sure about the second suggestion. openhablog is quickly typed and the function is clear. "oh" is not an official abbreviation.

mstormi:
what's "official" anyway ?
I feel annoyed by havin to type anything longer than it has to be, but YMMV.

PR #297, opened 2017-12-22
Fix #286 - Homegear won't install on Openhabian 1.4

### LINKED ISSUE #286: Homegear won't install because list file points to Jessie repository
When trying to install Homegear it fails on Openhabian 1.4 with these errors:

> The following packages have unmet dependencies:
 homegear : Depends: libgnutls-deb0-28 (>= 3.3.0) but it is not installable
            Depends: libmysqlclient18 (>= 5.5.24+dfsg-1) but it is not installable
            Depends: libssl1.0.0 (>= 1.0.1) but it is not installable
E: Unable to correct problems, you have held broken packages.

The problem seems to be in `/etc/apt/sources.list.d/homegear.list` which points to Jessie. Looks like this comes from here `https://github.com/openhab/openhabian/blob/master/functions/packages.sh` but I'm not too sure. Happy to submit a Pull Request to update that file but I'm not sure if you've fully migrated to Stretch or if there's still Jessie support in there.

I guess other list files also need to be updated for Stretch

lfrancke:
This adds support for all OSes supported by Homegear at the moment and should fix #286 

I've added the functions to helper.sh but ended up not using them. I'm an absolute bash newbie so I hope this solution is fine for you.
  
Signed-off-by: Lars Francke <mail@lars-francke.de> (github: lfrancke)

### DIFF
--- a/functions/helpers.sh
+++ b/functions/helpers.sh
@@ -87,7 +87,23 @@ is_debian() {
   [[ $(lsb_release -d) =~ "Debian" ]]
   return $?
 }
+is_raspbian() {
+  [[ "$(lsb_release -si)" = "Raspbian" ]]
+  return $?
+}
 is_jessie() {
   [[ $(lsb_release -c) =~ "jessie" ]]
+  return $?
+}
+is_stretch() {
+  [[ $(lsb_release -sc) =~ "stretch" ]]
+  return $?
+}
+is_trusty() {
+  [[ $(lsb_release -sc) =~ "trusty" ]]
+  return $?
+}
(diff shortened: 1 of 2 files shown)


ThomDietrich:
Sorry for the holidays related delay.
Excellent solution. Thanks @lfrancke !

lfrancke:
@ThomDietrich thanks! Sorry, I missed the contributors guidelines. I've added the sign off to the PR description.

ThomDietrich:
No problem. Thanks for your contribution!

PR #309, opened 2018-01-12
Improved PINE64 board detection

EliasGabrielsson:

### DIFF
--- a/functions/helpers.sh
+++ b/functions/helpers.sh
@@ -64,8 +64,10 @@ is_pi() {
   return 1
 }
 is_pine64() {
-  [[ $(uname -r) =~ "pine64-longsleep" ]]
-  return $?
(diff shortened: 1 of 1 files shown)


ThomDietrich:
Hey Elias,
so the A64 is the only one with this chip? :)

leonsio:
hi i have created a "little" script to detect different types of SBCs. it cat detect pine64 and pine64+ boards, and many other. you can include it in your project 

https://github.com/leonsio/arm-board-detect

i will try to add a detection of any new boards, which I can find.

EliasGabrielsson:
@leonsio This is an amazing work.
I think we definitely should use your work.

I will close this for now and we can do a major refraction further on instead.

ThomDietrich:
I agree.
@leonsio would you like to add a PR?

leonsio:
ok, i will create one

PR #322, opened 2018-01-24
Added PINE64 support for system binding and gpio closes #310

### LINKED ISSUE #310: System binding need 32-bit patch on PINE64 platform
### Symptom

Currently the system binding will throw errors regarding `udev` because of missing 32 bit library on PINE64.

### Fix

Apply the fix described in [this post](https://community.openhab.org/t/systeminfo-initialization-error-on-pine64/11726/9) for pine64. This will maybe be fixed upstream later on so put it in a special "fix function" for easy removal later on.

EliasGabrielsson:


functions/system.sh
```diff
@@ -162,6 +162,9 @@ permissions_corrections() {
     echo "FAILED (please execute after openHAB was installed)"
     exit 1
   fi
+
+  if is_pine64; then pine64_gpio_permissions;fi
```
ThomDietrich: Would it be easier to just move the four lines here?

functions/system.sh
```diff
@@ -217,6 +220,19 @@ pine64_platform_scripts() {
...
+  cond_redirect cp $BASEDIR/includes/PINE64-80-gpio-noroot.rules /etc/udev/rules.d/80-gpio-noroot.rules
+  cond_redirect sed -i -e '$i \chown -R root:gpio /sys/class/gpio \n' /etc/rc.local
+  cond_redirect sed -i -e '$i \chmod -R ug+rw /sys/class/gpio \n' /etc/rc.local
+} 
+
+pine64_fix_system_binding() { # This will maybe be fixed upstreams some day. Keep an eye open.
+  cond_redirect apt-get install -y udev:armhf
+  cond_redirect ln -s /lib/arm-linux-gnueabihf/ /lib/linux-arm
+  cond_redirect ln -s /lib/linux-arm/libudev.so.1 /lib/linux-arm/libudev.so
+}
```
ThomDietrich: Please a logging line, the usual "timstamp, openHABian, Doing this and that... " + "OK" would be fine.
Also please add some error checking and testing. Those `ln` calls will e.g. fail if the files already exist. I'm looking at future additions in the base system here...

Please replace apt-get by apt. Not that it makes a difference but all other functions use apt and it's a matter of consistency.

functions/system.sh
```diff
@@ -217,6 +220,19 @@ pine64_platform_scripts() {
   fi
 }
 
+pine64_gpio_permissions() {
+  cond_redirect groupadd gpio
+  cond_redirect cp $BASEDIR/includes/PINE64-80-gpio-noroot.rules /etc/udev/rules.d/80-gpio-noroot.rules
+  cond_redirect sed -i -e '$i \chown -R root:gpio /sys/class/gpio \n' /etc/rc.local
+  cond_redirect sed -i -e '$i \chmod -R ug+rw /sys/class/gpio \n' /etc/rc.local
```
ThomDietrich: why so complicated?

functions/system.sh
EliasGabrielsson: It dosen't matter for me I can move them. I thought if we add more specific boards later it's easy to get an overview by having functions instead of inline code.

functions/system.sh
EliasGabrielsson: As I understand it from [this post](https://forum.pine64.org/showthread.php?tid=2233) permission is lost every reboot due to a virtual filesystem. They need to be reapplied every boot, correct me if I am wrong. This works for me at least.

EliasGabrielsson:
Updated PR. Ready for another review.

functions/system.sh
ThomDietrich: Sorry, I didn't express myself clearly. I was confused about the fairly odd sed line but now I understood their meaning. 
Okay then, seems like a good solution ;)

includes/PINE64-80-gpio-noroot.rules
```diff
@@ -0,0 +1,7 @@
+# /etc/udev/rules.d/80-gpio-noroot.rules
+# Corrects sys GPIO permissions on the Pine64 so non-root users in the gpio group can manipulate bits
+#
+# Change group to gpio
+SUBSYSTEM=="gpio", PROGRAM="/bin/sh -c '/bin/chown -R root:gpio /sys/devices/soc.0/*pinctrl/gpio'"
+# Change user permissions to ensure user and group have read/write permissions
+SUBSYSTEM=="gpio", PROGRAM="/bin/sh -c '/bin/chmod -R ug+rw /sys/devices/soc.0/*pinctrl/gpio'"
```
ThomDietrich: Please add a newline after this one

ThomDietrich:
Thanks! 🏅

PR #325, opened 2018-02-03
Changed Nginx script to explicit use external DNS

### LINKED ISSUE #324: Ngnix configuration does not explicit use external DNS
## Symtom
While configuring nginx for https the entered dns (eg. openhab.besteva.com) is checked to point to the local instance. I use a NAT:ed ip range (eg. 10.0.0.34) which a local DNS is serving. When the external IP is compared to the local DNS there is mismatch and the script will stop.

## Expected result
The script will explicit use an external DNS server which not serving NAT:ed IP:s (local ip).

EliasGabrielsson:
Fixes #324

### DIFF
--- a/functions/packages.sh
+++ b/functions/packages.sh
@@ -440,7 +440,7 @@ nginx_setup() {
 
   while [ "$VALIDDOMAIN" = false ] && [ ! -z "$domain" ] && [ "$domain" != "IP" ]; do
     echo -n "Obtaining domain IP address... "
-    domainip=$(dig +short $domain |tail -1)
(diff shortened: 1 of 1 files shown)


ThomDietrich:
Makes sense and looks good to me. Thanks!

ThomDietrich:
@EliasGabrielsson I'm still seeing the issue with your mail address. The one you are using to create commits is unknown to GitHub as we can see here: 

grafik

You need to add this mail address to your GitHub settings.

EliasGabrielsson:
Hmm, this is strange. It's added. :-/ I will look into it.

PR #346, opened 2018-03-17
Restore ability to choose new conffile or not

BClark09:
I think this option was a good measure to prevent logging failure. But if there is a need to publish changes in linuxpkg that will overwrite other files like `/etc/default/openhab2`, then this becomes problematic.

Hopefully most people are on the latest stable/unstable by now.


ThomDietrich:
I agree, this was a good workaround back then but will do more harm than good in the long run. Thanks

PR #344, opened 2018-03-04
Fixed function call typo

EliasGabrielsson:
Fixes #341

Signed-off-by: Elias Gabrielsson <elias@benefactory.se> (github: EliasGabrielsson)

PR #348, opened 2018-03-17
Fix unstable function call naming

BClark09:
Similar to PR: #344 


ThomDietrich:
Ouch... Thanks!

PR #354, opened 2018-03-27
fixing failing drive0 and drive1 symlink creation

mues-lee:
Hi,
the target folder for
`ln -s ${storage}/slots ${storage}/slots/drive0;ln -s ${storage}/slots ${storage}/slots/drive1`
doesn't exist, to avoid the creation fail the need folder is created in before

Best
mues-lee

### DIFF
--- a/functions/backup.sh
+++ b/functions/backup.sh
@@ -20,7 +20,7 @@ create_backup_config() {
       if ! (whiptail --title "Storage container creation" --yes-button "Continue" --no-button "Back" --yesno "$introtext" 15 80) then echo "CANCELED"; return 0; fi
   fi
   # create virtual 'tapes'
-  mkdir ${storage}/slots # folder needed for following symlinks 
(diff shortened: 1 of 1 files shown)


ThomDietrich:
Seems correct. @mstormi 

@mues-lee the mail address you provided is correct?

mues-lee:
@ThomDietrich yes the mail address was correct, but since you asked I changed it to a more matching address ;) the new mail address of course also is working

mstormi:
to play it safe, you should add a `/bin/chown ${backupuser}:${backupuser} ${storage}/slots`

mues-lee:
@mstormi 
thanks for this hint, folder has now the correct owner and group

ThomDietrich:
Thanks!

@mstormi what's the deal with those two symlinks anyhow? Looks like a rather dirty hack to me 🙈

mstormi:
yes it's dirty. amlabel (called during installation) or amdump (latest) should create them so it shouldn't be necessary, but as I (and obviously others, too) have observed, this does not happen always so it's a hack to play things safe anyway.

Issue #269, opened 2017-12-12
Combine toolchain for RPi and for PineA64 to one with conditional actions.


ThomDietrich:
The files for both images are mostly identical. For the next image release I plan to merge inside
https://github.com/openhab/openhabian/tree/master/build-rpi-raspbian and
https://github.com/openhab/openhabian/tree/master/build-pine64-image

EliasGabrielsson:
Fixed in: https://github.com/openhab/openhabian/pull/358

PR #358, opened 2018-03-30
merge buildchain and add custom repositories

EliasGabrielsson:
This patch:
* merge `first-boot.sh` files to a common one for all platforms with conditional actions
* adds two build options for custom repositories, `dev-git` and `dev-url`
* adds `test build` watermark to builds created with `dev-git` and `dev-url`
* extract common build code to a new file, `build-common.sh`

Things to note:
* Adds dependency to `functions/helpers.sh`

Closes issue #269

Signed-off-by: Elias Gabrielsson elias@gabson.se (github: EliasGabrielsson)


ThomDietrich:
Interesting addition. I am not against it but may I ask why so complicated? You could have simply provided the full URL as a parameter or simply change URL after flashing...

EliasGabrielsson:
I do agree with you that the code is a bit more complicated then URL/branch but I think it's worth it.
It's really convenient to just typ `git push` followed by `sudo bash build-pine64.sh`. It's remove another human made error prone step out of the equation. Testing script by a new installation takes time so have a typo there "cost" a lot of time.

As a refactoring I think a file `build-common.sh` would be nice to have shared build code. It's not unreasonable to think that more specific platforms will be supported given how many new ones are released every year. Then the git url identification code can be over there. 

Having just an URL would be a case as well, totally agree.
```
build-pine64.sh go
build-pine64.sh dev-git
build-pine64.sh dev-url "branch" "URL"
```

EliasGabrielsson:
I make another try on this and we will see if we are happy enough after that. I take #269 into account as well. Better to review things an extra round before going into merge. 👍

ThomDietrich:
Didn't quite understand you. Is this PR ready to merge? It's content looks good to me.

#269 is something you can of course look into, otherwise I'll do that for images v1.6.

EliasGabrielsson:
This is not ready to merge yet.

EliasGabrielsson:
Please have an initial look on this @ThomDietrich when you have time.
Before merging I need to to one more complete installation from scratch to verify functionality from latest changes.

EliasGabrielsson:
Finished up this work today. 
Since last commit:
* I choose to keep `$clone_string` much cleaner then having to variable all over.
* Merged all build script into a single file. Don't know why I didn't this in the first place, obvious when you think about it.
* Merged `webif.sh` to the common folder to both platforms.

Built images for both Pine64 and Raspberry Pi.
Install tested on Pine64. Had a strange error first run with web interface not loading hostname for iFrame. Next test run worked as intended, let's keep an eye on it.

I will merge this now and are ready to do hot fixes in the unlikely that something else breaks.

build-image/first-boot.sh
```diff
@@ -33,7 +33,11 @@ else
   echo "OK"
 fi
 
-userdef="pi"
+userdef="openhabian"
+if is_pi; then
```
ThomDietrich: @EliasGabrielsson did I overlook the function definition?

build-image/first-boot.sh
```diff
@@ -57,11 +61,8 @@ elif grep -q "openHABian" /etc/wpa_supplicant/wpa_supplicant.conf; then
     rfkill unblock wifi
     ifconfig wlan0 up
     if iwlist wlan0 scanning 2>&1 | grep -q "Interface doesn't support scanning"; then
-      echo "FAILED"
-      echo ""
-      echo "I was not able to turn on the wifi"
-      echo "Here is some more information"
-      echo ""
+      echo "FAILED" 
+      echo -n "$(timestamp) [openHABian] I was not able to turn on the wifi \n Here is some more information: \n"
```
ThomDietrich: Everywhere you wrote " \n Here" you'll have a whitespace at the beginning of the line. I'd suggest to fix that.

build-image/first-boot.sh
```diff
@@ -86,14 +92,9 @@ until ping -c1 9.9.9.9 &>/dev/null || [ "$(wget -qO- http://www.msftncsi.com/ncs
...
-      echo ""
-      echo "I was not able to connect to the configured Wi-Fi."
-      echo "Please check your signal quality. Reachable Wi-Fi networks are:"
+      echo -n "$(timestamp) [openHABian] I was not able to connect to the configured Wi-Fi. \n Please check your signal quality. Reachable Wi-Fi networks are: \n"
       iwlist wlan0 scanning | grep "ESSID" | sed 's/^\s*ESSID:/\t- /g'
-      echo ""
-      echo "Please try again with your correct SSID and password."
-      echo "The following Wi-Fi configuration was used:"
-      echo ""
+      echo -n "$(timestamp) [openHABian] Please try again with your correct SSID and password. \n The following Wi-Fi configuration was used: \n"
```
ThomDietrich: You did change the number of newlines in those two blocks. They had their esthetic purpose. I'd suggest to restore them.

ThomDietrich:
Hey, besides my comments the PR looks goo to me.

> I will merge this now and are ready to do hot fixes in the unlikely that something else breaks.

It is a rather big PR and I'd have preferred another review by me or someone else. Anyhow, please check my comments. Thanks for your effort!!

build-image/first-boot.sh
EliasGabrielsson: Sorry but I don't understand what you mean?

build-image/first-boot.sh
EliasGabrielsson: Will fix!

build-image/first-boot.sh
ThomDietrich: Where is `is_pi` defined? Sorry if I overlooked it.

build.sh
```diff
@@ -0,0 +1,231 @@
...
+  git clone -b master https://github.com/longsleep/build-pine64-image.git $buildfolder
+  
+  echo_process "Downloading aditional files needed by \"longsleep/build-pine64-image\" project... "
+  wget -nv -P $buildfolder/ https://www.stdin.xyz/downloads/people/longsleep/pine64-images/simpleimage-pine64-latest.img.xz
+  wget -nv -P $buildfolder/ https://www.stdin.xyz/downloads/people/longsleep/pine64-images/linux/linux-pine64-latest.tar.xz
+
+  echo_process "Copying over 'rc.local' and 'first-boot.sh' for image integration... "
+  cp $sourcefolder/rc.local $buildfolder/simpleimage/openhabianpine64.rc.local
+  cp $sourcefolder/first-boot.sh $buildfolder/simpleimage/openhabianpine64.first-boot.sh
+  sed -i -e '1r functions/helpers.sh' $buildfolder/simpleimage/openhabianpine64.first-boot.sh # Add platform identification
```
EliasGabrielsson: @ThomDietrich `is_pi` is added here

build-image/first-boot.sh
EliasGabrielsson: No worries, it's in the commit message:

"Adds dependency to `functions/helpers.sh`"

referred you on the specific places

PR #359, opened 2018-03-30
Re-add rpi-update to basic packages

ThomDietrich:
CC @mues-lee 

### DIFF
--- a/functions/system.sh
+++ b/functions/system.sh
@@ -18,10 +18,6 @@ system_upgrade() {
 
 basic_packages() {
   echo -n "$(timestamp) [openHABian] Installing basic can't-be-wrong packages (screen, vim, ...)... "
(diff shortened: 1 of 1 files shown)


mues-lee:
the whole `if is_pi is unnessesary
- `rpi-update` is still there and working before it was removed by 'cond_redirect rm -f /usr/bin/rpi-update'
- `rpi-update` is in the apt repository, so it got already to the newest version, when we fired apt update and & apt upgrade in the first boot

ThomDietrich:
I remember there were times when it wasn't part of the raspbian image (maybe the ua-netinst image). So you are telling me it is redundant because it's available on a Raspbian image by default?

mues-lee:
yes exactly

ThomDietrich:
👍

PR #360, opened 2018-03-30
fixing failing git installation

mues-lee:
handle repository changes in `Updating repositories and upgrading installed packages...` before git install


ThomDietrich:
Until #269 we need to keep both files in sync. (Background: In the beginning they were pretty different. Since then I've synchronized as much steps as possible. By now merging both shouldn't be a big issue.

ThomDietrich:
@mues-lee I saw that you are working in master. Not good :) You should always work in a feature branch. See: https://guides.github.com/introduction/flow/

mues-lee:
@ThomDietrich  thank you for that hint, i will do so in future

PR #365, opened 2018-04-03
updating grafana repository to use stretch builds

mues-lee:

### DIFF
--- a/functions/packages.sh
+++ b/functions/packages.sh
@@ -389,6 +389,10 @@ influxdb_grafana_setup() {
   successtext="Setup successful. Please continue with the instructions you can find here:\n\nhttps://community.openhab.org/t/13761/1"
 
   echo "$(timestamp) [openHABian] Setting up InfluxDB and Grafana... "
+  distributor=$(lsb_release -si | awk '{print tolower($0)}')
+  if ! [ $distributor = "ubuntu" ]; then #since we only have ubuntu and debian as supported OS at the moment
+        distributor="debian"             # replacing e.g. raspbian with debian
+  fi
   codename=$(lsb_release -sc)
   if [ -n "$INTERACTIVE" ]; then
     if ! (whiptail --title "Description, Continue?" --yes-button "Continue" --no-button "Back" --yesno "$introtext" 15 80) then echo "CANCELED"; return 0; fi
@@ -398,7 +402,7 @@ influxdb_grafana_setup() {
   cond_echo ""
   echo "InfluxDB... "
   cond_redirect wget -O - https://repos.influxdata.com/influxdb.key | apt-key add - || FAILED=1
-  echo "deb https://repos.influxdata.com/debian $codename stable" > /etc/apt/sources.list.d/influxdb.list || FAILED=1
+  echo "deb https://repos.influxdata.com/$distributor $codename stable" > /etc/apt/sources.list.d/influxdb.list || FAILED=1
   cond_redirect apt update || FAILED=1
   cond_redirect apt -y install influxdb || FAILED=1
   cond_redirect systemctl daemon-reload
@@ -414,7 +418,10 @@ influxdb_grafana_setup() {
     cond_redirect apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 379CE192D401AB61 || FAILED=2
   else
     cond_redirect wget -O - https://packagecloud.io/gpg.key | apt-key add - || FAILED=2
-    echo "deb https://packagecloud.io/grafana/stable/debian/ $codename main" > /etc/apt/sources.list.d/grafana.list || FAILED=2
+    echo "deb https://packagecloud.io/grafana/stable/$distributor/ $codename main" > /etc/apt/sources.list.d/grafana.list || FAILED=2
+    if [ $distributor = "ubuntu" ]; then #remove this loop when grafana ubuntu packages are available
+        echo "deb https://packagecloud.io/grafana/stable/debian/ jessie main" > /etc/apt/sources.list.d/grafana.list || FAILED=2
+    fi
   fi
   cond_redirect apt update || FAILED=2
   cond_redirect apt -y install grafana || FAILED=2
--- a/functions/packages.sh
+++ b/functions/packages.sh
@@ -409,11 +409,11 @@ influxdb_grafana_setup() {
   echo "Grafana... "
   if is_pi; then
     if is_pione || is_pizero || is_pizerow; then GRAFANA_REPO_PI1="-rpi-1b"; fi
-    echo "deb https://dl.bintray.com/fg2it/deb${GRAFANA_REPO_PI1} jessie main" > /etc/apt/sources.list.d/grafana-fg2it.list || FAILED=2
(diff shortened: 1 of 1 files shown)


EliasGabrielsson:
What do you think about this? https://github.com/openhab/openhabian/issues/367
More job yes but longterm stability will be better.

mues-lee:
@EliasGabrielsson please review the changes, hope this was what you meant in #367

ThomDietrich:
Is this routine safe?
Are repository branches available for jessie, stretch, trusty, xenial, ...?

BClark09:
The [influxdata repo](https://repos.influxdata.com/debian/dists) contains the major ones. However, the script will fail for Ubuntu users because Grafana only contains `jessie`, `stretch` and `wheezy`. 

Grafana [recommend in the docs](http://docs.grafana.org/installation/debian/) using Jessie regardless of distro used.

mues-lee:
it's true the docs on grafana still recommend using jessie, but i think this page is just outdated, of course i don't have any prove of my thinking

about ubuntu, yes it would fail for them, but as i understood the function, it was never meant to work on ubuntu, since there wasn't a distinguish between debian and ubuntu at all
sure this can be added, but i believe a central platform recognition would be much better then doing this in each function

functions/packages.sh
```diff
@@ -397,7 +398,7 @@ influxdb_grafana_setup() {
   cond_echo ""
   echo "InfluxDB... "
   cond_redirect wget -O - https://repos.influxdata.com/influxdb.key | apt-key add - || FAILED=1
-  echo "deb https://repos.influxdata.com/debian jessie stable" > /etc/apt/sources.list.d/influxdb.list || FAILED=1
+  echo "deb https://repos.influxdata.com/debian $codename stable" > /etc/apt/sources.list.d/influxdb.list || FAILED=1
```
EliasGabrielsson: Don't have an answer to this but which of these two could we use?
`https://repos.influxdata.com/debian/dists/*` or `https://repos.influxdata.com/ubuntu/dists/*`

Both seams to have all major releases for both Ubuntu and Debian.

functions/packages.sh
```diff
@@ -409,11 +410,11 @@ influxdb_grafana_setup() {
   echo "Grafana... "
   if is_pi; then
     if is_pione || is_pizero || is_pizerow; then GRAFANA_REPO_PI1="-rpi-1b"; fi
-    echo "deb https://dl.bintray.com/fg2it/deb${GRAFANA_REPO_PI1} stretch main" > /etc/apt/sources.list.d/grafana-fg2it.list || FAILED=2
+    echo "deb https://dl.bintray.com/fg2it/deb${GRAFANA_REPO_PI1} $codename main" > /etc/apt/sources.list.d/grafana-fg2it.list || FAILED=2
     cond_redirect apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 379CE192D401AB61 || FAILED=2
   else
     cond_redirect wget -O - https://packagecloud.io/gpg.key | apt-key add - || FAILED=2
-    echo "deb https://packagecloud.io/grafana/stable/debian/ stretch main" > /etc/apt/sources.list.d/grafana.list || FAILED=2
+    echo "deb https://packagecloud.io/grafana/stable/debian/ $codename main" > /etc/apt/sources.list.d/grafana.list || FAILED=2
```
EliasGabrielsson: Stretch seams to be deploy and used for pkg (https://github.com/grafana/grafana/issues/8648) but the ubuntu releases is not available as @BClark09 mention. Some clever filtering is needed here, like a fallback for Ubuntu.

functions/packages.sh
EliasGabrielsson: Maybe just do an `if´ case for the Ubuntu case?

EliasGabrielsson:
@mues-lee Great changes. Made two comments in the code which sums up reasoning in this thread. When they are solved I would say this is ready for merge.

mues-lee:
@EliasGabrielsson please review my recent changes

functions/packages.sh
```diff
@@ -414,7 +418,10 @@ influxdb_grafana_setup() {
     cond_redirect apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 379CE192D401AB61 || FAILED=2
   else
     cond_redirect wget -O - https://packagecloud.io/gpg.key | apt-key add - || FAILED=2
-    echo "deb https://packagecloud.io/grafana/stable/debian/ $codename main" > /etc/apt/sources.list.d/grafana.list || FAILED=2
+    echo "deb https://packagecloud.io/grafana/stable/$distributor/ $codename main" > /etc/apt/sources.list.d/grafana.list || FAILED=2
+    if [ $distributor = "ubuntu" ]; then #remove this loop when grafana ubuntu packages are available
+        echo "deb https://packagecloud.io/grafana/stable/debian/ jessie main" > /etc/apt/sources.list.d/grafana.list || FAILED=2
+    fi
```
EliasGabrielsson: Correct me if I am wrong but isn't something like this what we are after as there are no `ubuntu` packages available?

```
if [ $codename = "jessie" ]; then
    echo "deb https://packagecloud.io/grafana/stable/debian/ jessie main" > /etc/apt/sources.list.d/grafana.list || FAILED=2
else
    echo "deb https://packagecloud.io/grafana/stable/debian/ stretch main" > /etc/apt/sources.list.d/grafana.list || FAILED=2
fi
```

EliasGabrielsson: Have not tested above code, could contain syntax error etc.

mues-lee: well this is what my changes do, with the added benefit that once when there will be the packages for Ubuntu will be there, the if loop can be removed and we end up with a cleaner code with less points to adjust when a new supported stable release from, let's say from debian, is released.

But of course we can do it now withe the if-else loop, this should not be my decision

EliasGabrielsson: Totally misread your if statement. It's brilliant and I like it a lot!

EliasGabrielsson: But doesn't we end up with an invalid entry in `sources.list` for `Ubuntu` then which we can't resolve?
Eg. `"deb https://packagecloud.io/grafana/stable/ubuntu/ xenial  main"` will produces sources can't be found.

mues-lee: well no ... line 421 would insert `deb https://packagecloud.io/grafana/stable/ubuntu/ xenial main` but than it will be overwritten by the if loop with `deb https://packagecloud.io/grafana/stable/debian/ jessie main´ 
i have planed this for the future, when there maybe will be ubuntu packages for grafana, then we can just remove the if loop and we will be done

EliasGabrielsson:
@mues-lee Thanks for your patience when in reviewing this.
Now we will have correct packages with for all OS:es. 🥁🥁🥁

mues-lee:
@EliasGabrielsson you are welcome

mues-lee:
that is correct, hadn't think about this, but it would be know problem to exchange my special checks, just let me know if i should do this

ThomDietrich:
If I remember correctly you wanted to add the additional configuration steps as described in the tutorial. If this is the case, just do it while at that ;)

mues-lee:
not sure about which 
> additional configuration steps as described in the tutorial
you mean

ThomDietrich:
https://github.com/openhab/openhabian/issues/366

PR #375, opened 2018-04-10
added is_pithreeplus() - closes #374

### LINKED ISSUE #374: Detect RPi3B+
https://community.openhab.org/t/raspberry-3-b-wifi-on-openhabian/43417

@mues-lee could you troubleshoot this problem for me?

My guess: This routine doesn't match the new Pi, hence that condition isn't true:

https://github.com/openhab/openhabian/blob/master/functions/helpers.sh#L55
https://github.com/openhab/openhabian/blob/master/functions/wifi.sh#L5

mues-lee:
closes #374

could been done also by just adjusting the regex form is_pithree() and the differentiation in pi3b and pi3b+ is not really needed at the moment, but is "a nice to have"-feature  for future use


functions/system.sh
```diff
@@ -276,7 +276,7 @@ Finally, all common serial ports can be made accessible to the openHAB java virt
 
   # Find current settings
   if is_pi && grep -q "enable_uart=1" /boot/config.txt; then sel_1="ON"; else sel_1="OFF"; fi
-  if is_pithree && grep -q "dtoverlay=pi3-miniuart-bt" /boot/config.txt; then sel_2="ON"; else sel_2="OFF"; fi
+  if ( is_pithree || is_pithreeplus ) && grep -q "dtoverlay=pi3-miniuart-bt" /boot/config.txt; then sel_2="ON"; else sel_2="OFF"; fi
```
ThomDietrich: why braces here?

ThomDietrich:
I'm also not sure when and how this could be useful. Now that you went the extra mile I' okay to do it this way.

mues-lee:
most likely will will never need it

also i noticed that I somehow managed to not branch the master, but my own webif branch, sorry for this
maybe it will be better to discard this PR, to keep everything separated and clean

functions/system.sh
ThomDietrich: Ah I see. In bash this is in fact not needed.

https://unix.stackexchange.com/a/88851

ThomDietrich:
Yes I saw that. A new PR would be the easiest solution, yeah

mues-lee:
PR will be back in a clean state

functions/system.sh
mues-lee: i will change this in the new PR

PR #383, opened 2018-04-22
Update installer to use Node.js 8 (from v7)

### LINKED ISSUE #379: Installation failing on Ubuntu Server 17.10 "Artful Aardvark"
I ran across an issue installing OpenHABian on Ubuntu 17.10.  It doesn't appear to be an OpenHABian issue directly, but I wanted to log it here because it caused my installer to fail and it may be useful to others.

When installing Node.js, it appears the installer queries the node release servers and looks for a matching release rname.  At this time, only the Trusty, Xenial, Yakkety, and Zesty Ubuntu releases are supported.

I believe this is an upstream issue.  I don't think there's anything to be done directly by OpenHABian (aside from wait for Artful to be supported, and for me to use a different distro for this install).

brianwarner:
Resolves #379

Installation of fronttail was failing on Ubuntu distributions newer than Zesty
because they were not supported by Node v7. This patch changes the installed
version of Node to 8 in the openhabian-setup.sh installer.

Signed-off-by: Brian Warner <brian@bdwarner.com>

### DIFF
--- a/functions/nodejs-apps.sh
+++ b/functions/nodejs-apps.sh
@@ -15,8 +15,8 @@ nodejs_setup() {
     cond_redirect rm /tmp/nodejs-armv6l.tar.gz 2>&1
   else
     echo -n "$(timestamp) [openHABian] Installing Node.js (prerequisite for other packages)... "
-    cond_redirect wget -O /tmp/nodejs-v8.x.sh https://deb.nodesource.com/setup_8.x || FAILED=1
-    cond_redirect bash /tmp/nodejs-v8.x.sh || FAILED=1
+    cond_redirect wget -O /tmp/nodejs-setup.sh https://deb.nodesource.com/setup_8.x || FAILED=1
+    cond_redirect bash /tmp/nodejs-setup.sh || FAILED=1
     if [ $FAILED -eq 1 ]; then echo "FAILED (nodejs preparations)"; exit 1; fi
     cond_redirect apt -y install nodejs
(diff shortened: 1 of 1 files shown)


ThomDietrich:
Thanks!

All, @brianwarner, @EliasGabrielsson, @mues-lee and @Blakko,
how do you feel about changing the armv6l branch as well? It currently installs baron (v6), even though a [carbon](https://nodejs.org/download/release/latest-carbon/) (v8) binary is available.

mues-lee:
@ThomDietrich i think upgrading to carbon would be a good idea, but aren't the RPi3B and 3B+ not armv7l or do i mix up something?

brianwarner:
I'm maybe not the best person to have a well-informed opinion on upgrading as I don't use Node for anything other than fronttail, but I'm happy to check if it works with the default setup.  My RPi2 will be free for experimentation shortly (once I finish a production setup on x86 Bionic), and I'll tweak the installer and give it a shot with Carbon.

Also, I looked into the question posed by @mues-lee.  It appears that:
* RPi Model B was ARMv6 (the armv6l build)
* RPi2 is an ARMv7 (the armv7l build)
* RPi3 is an ARMv8 (the arm64 build)

ARMv6 binaries should generally run on ARMv7.  I don't know enough about ARM backwards compatibility to say whether the ARMv6 binary will work with the 64 bit chips on the RPi3, though.  Maybe someone better informed than me could weigh in?

ThomDietrich:
The armv6l architecture can be found inn the RPi1 and more importantly the RPi0W. The latter is imho more important. I'll open an issue.

PR #377, opened 2018-04-10
Update wifi function for stretch, resolves #374

ThomDietrich:
Resolves #374 

@mues-lee @haroldh I just realized that we missed this change in #376 

Signed-off-by: Thomas Dietrich <thomas.dietrich@tu-ilmenau.de> (github: ThomDietrich)

PR #384, opened 2018-04-22
Fixes wifi country for wireless regulatory domain

### LINKED ISSUE #382: wireless regulatory domain is not set correctly
wireless regulatory domain informs the wifi controller what channels can be used, and this varies from country to country, and affects router connectivity.

see:  `/etc/default/crda`

```
[15:33:56] openhabian@openHABianPi:~$ cat /etc/default/crda
# Set REGDOMAIN to a ISO/IEC 3166-1 alpha2 country code so that iw(8) may set
# the initial regulatory domain setting for IEEE 802.11 devices which operate
# on this system.
#
# Governments assert the right to regulate usage of radio spectrum within
# their respective territories so make sure you select a ISO/IEC 3166-1 alpha2
# country code suitable for your location or you may infringe on local
# legislature. See `/usr/share/zoneinfo/zone.tab' for a table of timezone
# descriptions containing ISO/IEC 3166-1 alpha2 country codes.

REGDOMAIN=
```

Currently the country value is hard coded to "US"
see: https://github.com/openhab/openhabian/blob/master/build-rpi-raspbian/first-boot.sh#L48
and: https://github.com/openhab/openhabian/blob/master/functions/wifi.sh#L24

The users country code could be sourced and inserted where required.

jscottnz:
This fixes #382

This PR sets the wireless regulatory domain to the user set value, rather than the hardcoded `US`


ThomDietrich:
Hey @jscottnz,

there are a few things I'm not quite comfortable with. The first thing is that I would love (see #377) to auto-detect the country code. It's pretty easy:

```shell
curl -s http://ipinfo.io | jq -r '.country'
```

Maybe we should query at least two services to get reliable and continuous results.

The second comment is that a manually set country code in the configuration file is a good idea but you should still support a default. I'll have a more thorough review after you had the chance to comment on the geolocation idea.

jscottnz:
I think that geolocation is a grand idea - except that you need an internet connection to do that!
You MUST start with the config file - I guess that we can default to the US and then detect the location and then reset the configs

ThomDietrich:
You are right of course. Maybe it would be best to not change the first-boot script at all. We would gain more freedom if the step was part of `openhabian-setup.sh` and would not have to have a duplicate copy inside the wifi function.

ThomDietrich:
This was a mixup of thoughts. Of course we can add the config option and apply it, otherwise choose US (or any other suited choice). What I meant is, that we should move the geolocation based steps into `openhabian-setup.sh`.

ThomDietrich:
You know what, never mind. Let's first get this PR ready to merge, then we can discuss the second part. So to summarize:

* Part 1: (this PR) set the wifi country to the setting preconfigured by the user in `openhabian.conf`. If not, configure `country=US`
* Part 2: Add a new step to the unattended execution of `openhabian-setup.sh` to detect the users wifi region via geolocation, then set accordingly.

jscottnz:
@ThomDietrich I have updated the PR with defaults to US, and some other niceties.

functions/wifi.sh
```diff
@@ -13,6 +13,8 @@ wifi_setup() {
     if [ $? -ne 0 ]; then return 1; fi
     PASS=$(whiptail --title "Wifi Setup" --inputbox "What's the password for that Wifi?" 10 60 3>&1 1>&2 2>&3)
     if [ $? -ne 0 ]; then return 1; fi
+    WIFICOUNTRY=$(whiptail --title "Wifi Setup" --inputbox "What is the 2 letter country code? ie. US DE NZ AU...\nSee https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2" 10 60 3>&1 1>&2 2>&3)
```
ThomDietrich: "the" seems out of context.
How about:

> Please enter the two-letter country code matching your region. E.g. ...

ThomDietrich: Also I'd suggest to check for a two-letter uppercase input after this step.

build-rpi-raspbian/openhabian.raspbian.conf
```diff
@@ -17,4 +17,13 @@ system_default_locale="en_US.UTF-8"
...
 wifi_psk=""
 
+# Governments assert the right to regulate usage of radio spectrum within
+# their respective territories so make sure you select a ISO/IEC 3166-1 alpha2
+# country code suitable for your location or you may infringe on local
+# legislature. See `/usr/share/zoneinfo/zone.tab' 
+# For a table of timezone descriptions containing ISO/IEC 3166-1 alpha2 country codes.
+# See https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2 
+# eg: US DE AU NZ...
+wifi_country=""
```
ThomDietrich: How about we reduce this to a simple one/two-liner? Most details are irrelevant for the end user. We just need the "local legislature", wikipedia article and examples. Also mention US default

build-rpi-raspbian/first-boot.sh
```diff
@@ -44,9 +44,25 @@ elif grep -q "openHABian" /etc/wpa_supplicant/wpa_supplicant.conf; then
...
+  fi
+
   echo -e "# config generated by openHABian first boot setup" > /etc/wpa_supplicant/wpa_supplicant.conf
-  echo -e "country=US\\nctrl_interface=DIR=/var/run/wpa_supplicant GROUP=netdev\\nupdate_config=1" >> /etc/wpa_supplicant/wpa_supplicant.conf
+  echo -e "country=$wifi_country\\nctrl_interface=DIR=/var/run/wpa_supplicant GROUP=netdev\\nupdate_config=1" >> /etc/wpa_supplicant/wpa_supplicant.conf
   echo -e "network={\\n\\tssid=\"$wifi_ssid\"\\n\\tpsk=\"$wifi_psk\"\\n\\tkey_mgmt=WPA-PSK\\n}" >> /etc/wpa_supplicant/wpa_supplicant.conf
+  
+  echo -e "# config generated by openHABian first boot setup"
+  sed -i '$ d' /etc/default/crda # removes the last line
+  echo -e "REGDOMAIN=$wifi_country" >> /etc/default/crda # adds the country code
```
ThomDietrich: I don't like this method. It is doomed to create issues, I guarantee it...
Go with this instead: https://github.com/openhab/openhabian/blob/master/functions/system.sh#L245

build-rpi-raspbian/first-boot.sh
```diff
@@ -44,9 +44,25 @@ elif grep -q "openHABian" /etc/wpa_supplicant/wpa_supplicant.conf; then
   echo "$(timestamp) [openHABian] Setting up Wi-Fi connection... OK"
 else
   echo -n "$(timestamp) [openHABian] Setting up Wi-Fi connection... "
+
+  # check the user input for the country code
+  # check: from the start of line, the uppercased input must be followed by a whitespace
+  if grep -q "^${wifi_country^^}\s" /usr/share/zoneinfo/zone.tab; then
+    wifi_country=${wifi_country^^}
```
ThomDietrich: Good call. I didn't know the ^^ trick.

build-rpi-raspbian/first-boot.sh
```diff
@@ -44,9 +44,25 @@ elif grep -q "openHABian" /etc/wpa_supplicant/wpa_supplicant.conf; then
...
+  else
+    echo "${wifi_country} is not a valid country code found in /usr/share/zoneinfo/zone.tab"
```
ThomDietrich: You missed the case in which the user left ´wifi_country´ blank

build-rpi-raspbian/first-boot.sh
```diff
@@ -44,9 +44,25 @@ elif grep -q "openHABian" /etc/wpa_supplicant/wpa_supplicant.conf; then
...
+    echo "Defaulting to US"
+    wifi_country=US
+  fi
```
ThomDietrich: I'd suggest this:

```bash
wifi_country=${wifi_country^^}
if [ -z "$wifi_country" ]; then
  wifi_country=US
elif ! grep -q "^$wifi_country\s" /usr/share/zoneinfo/zone.tab; then
  echo "${wifi_country} is not a valid country code found in /usr/share/zoneinfo/zone.tab"
  wifi_country=US
fi
```

build-rpi-raspbian/openhabian.raspbian.conf
jscottnz: I've replaced the empty string with US, which will effectively make it default

jscottnz:
hmm, re: geolocation based detection, im not sure of the value. How many people would setup wifi while connected to a wired network? Especially since openhabian is installed via an image there is no need for a wired network.

ThomDietrich:
Imho this is unrelated to a wired network. Anyhow, I'll create a new issue.

PR #396, opened 2018-05-16
Retrieve gpg key unconditionally in openHAB setup

### LINKED ISSUE #393: openhabain should set gpg key for repository on stable and unstable releases
I setup openhabian with a manual install and the use the option fresh install. While using the unstable openhab option i was not able to install openhab unstable, becase the gpg key is not available. Tjis key is only set if you select the stable version.

It should be set in any case.

ThomDietrich:
Resolves #393


ThomDietrich:
Thanks @BClark09

PR #405, opened 2018-05-25
Update first-boot.sh

Alluuu:
If updating repositories and upgrading installed packages fails, give it another chance with reinstating dpkg

Signed-off-by: Alari Truuts<Alari.Truuts@gmail.com> (github: Alluuu)

### DIFF
--- a/build-rpi-raspbian/first-boot.sh
+++ b/build-rpi-raspbian/first-boot.sh
@@ -101,7 +101,14 @@ echo "OK"
 echo -n "$(timestamp) [openHABian] Updating repositories and upgrading installed packages... "
 apt update &>/dev/null
 apt --yes upgrade &>/dev/null
-if [ $? -eq 0 ]; then echo "OK"; else echo "FAILED"; fail_inprogress; fi
+if [ $? -eq 0 ]; then
+  echo "OK";
+else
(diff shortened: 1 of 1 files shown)


ThomDietrich:
Hey! great work and thanks for the PR. We could merge as is but I think this step would be better suited as part of the routine before this one in line 98. Wdyt?

ThomDietrich:
My main point is, we should be able to do this more elegantly. Should be is key here. I am not aware how it plays exactly and how we could improve this.

Fun fact: we had this extra step in the code base at some point but removed it when it wasn't needed anymore.

Anyhow, nothing speaks against merging the PR as is. Thanks for your contribution!!

ThomDietrich:
Ah one thing you missed, please sign your work. See contributing. The result has to look similar to:

Alluuu:
@ThomDietrich  Thanks for the input. I'm sure there could be a more elegant solution - the current solution at least quickly solves the issue for the time being.

Without changing code I can't write a commit message, any ideas on how to sign it off so it'd count?

ThomDietrich:
Just put it in the first message this time.

ThomDietrich:
Great. Done

PR #409, opened 2018-05-28
Generalise stable version in menu

BClark09:

### DIFF
--- a/functions/menu.sh
+++ b/functions/menu.sh
@@ -15,7 +15,7 @@ show_main_menu() {
   "" "" \
   "01 | Update"                  "Pull the latest revision of the openHABian Configuration Tool" \
   "02 | Upgrade System"          "Upgrade all installed software packages to their newest version" \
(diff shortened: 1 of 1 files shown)


BClark09:
I appear to have left off the > in the commit sign-off. :( Should I retry?

ThomDietrich:
No worries. Doesn't change the validity

ThomDietrich:
Regarding this specific entry: Previously I included it shortly after every new release. Do we want to keep it or just remove it forever?
There are the entries in "openHAB related".

Besides, How about we remove the "X"?

BClark09:
Sure thing, I think the entry fits well on the front page, but probably should be like the entries in the "openHAB related" section.

There is the problem that the script won't actually switch to 2.3.0 stable if 2.4.0-SNAPSHOT has already been installed, but I think that may be another issue.

ThomDietrich:
👍

Issue #337, opened 2018-02-12
Optional AddOn Node-Red installs and runs with root


freaky-m0:
I installed the nodered option with openhabian-config. This script needs to run with root (sudo).

In the installation process following question comes along:
```
$ bash /tmp/update-nodejs-and-nodered.sh 

Root user detected. Typically install as a normal user. No need for sudo.

Are you really sure you want to install as root ? (y/N) ? n
```

If you say "n" nodered isn't installed after it. If you say "y" nodered processes are running as "root"

I don't think that is right? It should run with a separate user or with openhabian at least or?

skatehouse:
Same issue here. Wanted to install via sudo openhabian-config. 
After choosing the install script for nodered the same as mentioned above.

`$ bash /tmp/update-nodejs-and-nodered.sh 

Root user detected. Typically install as a normal user. No need for sudo.

Are you really sure you want to install as root ? (y/N) ? n

$ npm install -g node-red-contrib-bigtimer 
npm WARN checkPermissions Missing write access to /usr/lib/node_modules/node-red-contrib-bigtimer
npm ERR! path /usr/lib/node_modules/node-red-contrib-bigtimer
npm ERR! code ELOOP
npm ERR! errno -40
npm ERR! syscall access
npm ERR! ELOOP: too many symbolic links encountered, access '/usr/lib/node_modules/node-red-contrib-bigtimer'


I can see that some directories and files have been created in /usr/lib/node_modules
Owner is nobody / Group is root 
It it possible to delete the whole directory?

freaky-m0:
Installing manually shouldn't be a problem. But the method from the openhabian-config wizard is not correct.

ThomDietrich:
Hey you two!
That sounds like a valid argument to me 👌 I did not implement this routine, it was provided by another developer who is not around any longer. Would you guys like to provide a Pull Request? From what I read it would be a good idea to create a separate user for nodered or to use the openhabian, possibly he openhab user. For some other services like frontail we are using openhab, this might be the easiest and best choice.

OTrumpf:
Now I understand the reason for this issue:  #213 

Perhaps, someone can explain, how to deinstall node-red (installed via openhabian-config under /root/.node-red) and how to perform a regular install with openHab while preserving all flows...

ThomDietrich:
@OTrumpf just uninstall and install following the normal instructions.

But seriously, it would be easier if someone provided a fix for the openHABian routine.

[...]

PR #420, opened 2018-06-13
Node-RED Install

msw1970:
Updated shell script file to install Node-RED as the openhabian user rather than root


EliasGabrielsson:
https://github.com/openhab/openhabian/issues/337 @ThomDietrich could you screen this?

ThomDietrich:
Theoretically the use of `sudo` is over the top as we are already root. However, if the command is able to solve the issue that would be fine by me. One thing though: The main user is not necessarily called "openhabian". The varible `$username` always holds the correct user name (most likely "openhabian").

@msw1970 would you be so kind to try `runuser -l $username -c "/tmp/update-nodejs-and-nodered.sh"`? Thanks!!

msw1970:
I agree that $username should be used. (Not sure how to update the PR though)

Tested the 3 options you suggested all of them fail with the error "sudo: no tty present and no askpass program specified"

Not sure what you mean with regard to #337 and updating addons from within the web UI??

PR #429, opened 2018-08-11
Removes /opt/openbian directory if it exist then attempts to clone again

jmcgrath207:
On the first install, the step "Cloning myself..." failed. I found it was an issue where the  /opt/openbian directory already existed and returned a code of 128. I added a conditional that removes the directory and attempts to reclone from GitHub again.

### DIFF
--- a/build-image/first-boot.sh
+++ b/build-image/first-boot.sh
@@ -145,7 +145,8 @@ if [ $? -eq 0 ]; then echo "OK"; else echo "FAILED"; fail_inprogress; fi
 
 echo -n "$(timestamp) [openHABian] Cloning myself... "
 /usr/bin/git clone -b master https://github.com/openhab/openhabian.git /opt/openhabian &>/dev/null
-if [ $? -eq 0 ]; then echo "OK"; else echo "FAILED"; fail_inprogress; fi
+if [ $? -eq 128 ]; then rm -Rf /opt/openbian && /usr/bin/git clone -b master https://github.com/openhab/openhabian.git /opt/openhabian &>/dev/null;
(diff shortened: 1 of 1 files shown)


EliasGabrielsson:
Hey John!
Need some clarification on what we are trying to solve. I currently perceived the situation as following. Please fill me in if I misunderstood anything.

1. Starting normal installation, copy in sd card and so on...
2. Installation got stuck in step `Cloning myself` 
3. By terminal you did manual clone eg. `git clone https://github.com/openhab/openhabian/` to fix the error?
4. As there was a folder you got error 128.

jmcgrath207:
Correct.
When I ran the command `/usr/bin/git clone -b master https://github.com/openhab/openhabian.git /opt/openhabian &>/dev/null` it returned the error code 128 since the folder `/opt/openhabian` already existed. I noticed that the folder already had similar contents but I assume that it was older. However, I can't be certain since I didn't run a `git log` before I applied my fix.

EliasGabrielsson:
Okey in that case I will close your request. Not because your solution is technically wrong but it doesn't fit the recommend use case. In your case a complete reinstallation is preferred, you probably just where unlucky with connection issues at the moment of cloning. 

If you and other experience this issue often an automated retry is the way to go. Don't forget the mantra of the project: **"A home automation enthusiast doesn't have to be a Linux enthusiast!"**. The user should not need to use the terminal more then absolutely necessary, especially not with development tools like git.

But hey, thank for contribution and raising awareness of an eventual issue. 👍

ThomDietrich:
An imho more elegant solution would have been to just delete `/opt/openhabian` IF available before the actual clone.


This is the first time I hear or see this error and I'm currently unclear in which case this error does actually occur... If anyone has more info or runs into it, please contribute

PR #430, opened 2018-08-11
Added Exim4 as MTA and config generation

mstormi:
Added Exim4 and generation of proper configuration to allow for sending mails either as an openHAB mail action or from non-openHAB openHABian components to require email to work such as Amanda.
By its default config created via menu option 16, Exim will be relaying mails through a public SMTP service such as Gmail.


EliasGabrielsson:
Hi Marcus.
Some initial feedback:
* `Exim` is currently not part of base installation, eg. new installed system from scratch.
Therefor it must be moved to `menu 20`.

* Use `local` variable for all function internal stuff.
This is a long-term goal to reduce risk of strange script errors
http://tldp.org/HOWTO/Bash-Prog-Intro-HOWTO-5.html

* Also comments in the code directly

functions/backup.sh
```diff
@@ -49,10 +49,9 @@ create_backup_config() {
       let counter+=1
   done
 
-# no mailer configured for now
-#  if [ -n "$INTERACTIVE" ]; then
-#     adminmail=$(whiptail --title "Admin reports" --inputbox "Enter the EMail address to send backup reports to. Note: Mail relaying is not enabled in openHABian yet." 10 60 3>&1 1>&2 2>&3)
-#  fi
+  if [ -n "$INTERACTIVE" ]; then
+     adminmail=$(whiptail --title "Admin reports" --inputbox "Enter the EMail address to send backup reports to. Note: Mail relaying requires an MTA to be setup such as Exim (openHABian menu option 16)." 10 60 3>&1 1>&2 2>&3)
```
EliasGabrielsson: Check text in general (newline, lowercase, uppercase etc.)

functions/packages.sh
```diff
@@ -40,6 +40,50 @@ firemotd_setup() {
   fi
 }
 
+create_mta_config() {
+  TMP=/tmp/.$$
+  HOSTNAME=`/bin/hostname`
+  INTERFACES="`/usr/bin/dig +short $HOSTNAME | /usr/bin/tr '\n' ';'`;127.0.0.1;::1"
+  EXIM_CLIENTCONFIG=/etc/exim4/passwd.client
+
+  introtext1="We will guide you through the install of exim4 as the mail transfer agent on your system and configure it to relay mails through a public service such as Google gmail.\nPlease
```
EliasGabrielsson: Check text in general (newline, lowercase, uppercase etc.)

functions/packages.sh
```diff
@@ -40,6 +40,50 @@ firemotd_setup() {
   fi
 }
 
+create_mta_config() {
```
EliasGabrielsson: Add function description like the one below especially explaining difference in behaviour while executed INTERACTIVE vs UNATTENDED.

https://github.com/openhab/openhabian/blob/de039dac49462990992cf127b82316202e639700/functions/passwords.sh#L3

functions/packages.sh
```diff
@@ -40,6 +40,50 @@ firemotd_setup() {
...
+  TMP=/tmp/.$$
+  HOSTNAME=`/bin/hostname`
+  INTERFACES="`/usr/bin/dig +short $HOSTNAME | /usr/bin/tr '\n' ';'`;127.0.0.1;::1"
+  EXIM_CLIENTCONFIG=/etc/exim4/passwd.client
+
+  introtext1="We will guide you through the install of exim4 as the mail transfer agent on your system and configure it to relay mails through a public service such as Google gmail.\nPlease
+enter the data shown in the next window when being asked for. You will be able to repeat the whole installation if required by selecting the openHABian menu for MTA again."
+  introtext2="Mail server type: mail sent by smarthost; received via SMTP or fetchmail\nSystem mail name: FQDN (your full hostname including the domain art)A\nIPs should be allowed by the se
```
EliasGabrielsson: Sorry but I don't understand this text. I don't get what will follow if I accept this. What is an `smarthost` for example? Remember the problem openhabian tries to solve: **A home automation enthusiast doesn't have to be a Linux enthusiast!**

If you need to include application specific terms and concept refer the reader to relevant documentation in the dialog box.

EliasGabrielsson: Check text in general (newline, lowercase, uppercase etc.)

functions/backup.sh
EliasGabrielsson: Add a dialog box which invoke the MTA installation from the backup script if it's needed.

functions/packages.sh
mstormi: I believe you misunderstood this. Actually the exim install routine - once called - will prompt the user with these texts. Now we cannot change the exim package to do that, but we can tell the user what he'll have to enter. That's what my window here is telling the user: what to enter if he's (later on) asked for a selection. He doesn't need to understand what a smarthost is just to make that choice.

  introtext2="Mail server type: mail sent by smarthost; received via SMTP or fetchmail\nSystem mail name: FQDN (your full hostname including the domain art)A\nIPs should be allowed by the se
```

File: functions/packages.sh (hunk 3 above)
Comment (mstormi) [reply functions/packages.sh]: I believe this routine is never called unattended, only ever via menu selection. It requires user to input data.
So shouldn't I rather remove the checks for INTERACTIVE ? I just kept them for safety reasons.

File: functions/packages.sh [hunk 5]
```diff
@@ -40,6 +40,50 @@ firemotd_setup() {
   fi
 }
 
+create_mta_config() {
+  TMP=/tmp/.$$
```
BClark09: If I understand the next lines correctly, this is a [security concern](https://en.wikipedia.org/wiki/Symlink_race) when using a temporary files to store passwords or other sensitive information.

Use [`mktemp`](https://www.gnu.org/software/autogen/mktemp.html) instead:

e.g.


```suggestion
  TMP="$(mktemp /tmp/.XXXXXXXXXX)"
```

functions/backup.sh
```diff
@@ -65,7 +62,7 @@ create_backup_config() {
   
   mkdir -p ${confdir}
   touch ${confdir}/tapelist
-  hostname=`/bin/hostname`
+  local hostname=`/bin/hostname`
```
BClark09: The newer form `$(cmd)` is used elsewhere in this file, rather than the deprecated form `` `cmd` ``.

mstormi:
A mailer to work isn't compulsory for Amanda to work but running without has a number of drawbacks such as non-sent mails to queue up (to possibly result in problems in the future), plus the functional drawback that a user has to actively check backup results.

BClark09:
That's fair enough. I don't use the feature so don't have any strong opinions but I think it would be better to allow the user to choose this function. If they skip or elect not to have the mail sent each night or if this function fails then the amcheck cron should have the "-m" omitted right?

```
Do you want to setup Amanda?
  Yes - Setup base configuration
    Did you want to have a nightly backup report via email?
      Yes - mail setup and add -m to amcheck
      No - enable Amanda as is and exit
  No - exit
```

mstormi:
To be honest I don't know how to disable Amanda to send mails on amdump. So while backups will work without I think it is not a reasonable option to not have any working mailer (disadvantages see above).
I rather opted to remove the amcheck as it does not make much sense anyway, it's meant to remind you to put in the right removable medium such as tape no. X, but that doesn't apply to the way we use it.
With this patch I'm checking for exim to be setup on Amanda install and if not I inform the user he should install it from the menu before returning to Amanda installation. But I also offer to continue with Amanda installation instead as it may be the user's intention to not install exim because he either does not want to have any MTA to relay mails (but deliver locally), or he already has a different MTA working. Remember openHABian is supposed to run on any UNIX and it's pretty much best practice to have a mailer in place.

functions/packages.sh
mstormi: @EliasGabrielsson now that benjy had his look, could you also have have another one please?

functions/backup.sh
```diff
@@ -111,28 +107,36 @@ create_backup_config() {
...
+  local successtext="Setup was successful. Amanda backup tool is now taking backups at 01:00. For further readings, start at http://wiki.zmanda.com/index.php/User_documentation."
 
   if [ -n "$INTERACTIVE" ]; then
     if ! (whiptail --title "Amanda backup installation" --yes-button "Yes" --no-button "No, I'll go read it" --defaultno --yesno "$querytext" 10 80) then return 0; fi
     whiptail --msgbox "$introtext" 25 132
   fi
+  
+  /usr/sbin/exim --version 2>/dev/null >/dev/null
+  if [ $? -ne 0 ]; then
+     if ! (whiptail --title "No exim mail transfer agent" --yes-button "Continue" --no-button "Back" --defaultno --yesno "Seems exim is not installed as a mail transfer agent.\nAmanda needs to be able to send emails. You should stop installing Amanda here and install exim from the openHABian menu (option 2C) first, then return here.\nOnly continue if you know there's a working mail transfer agent other than exim on your system.\nDo you want to continue with Amanda installation ?" 15 80) then echo "CANCELED"; return 0; fi
```
EliasGabrielsson: Please call the EXIM script directly. eg. `exim_setup()` if you need it installed.
Suggestionally, you change the text to or similar:
```

whiptail --title "No exim mail transfer agent" --yes-button "Install EXIM4" --no-button "MTA already exist, ignore installation" --defaultyes --yesno ...
```

functions/packages.sh
EliasGabrielsson: Then just write out that it will be ignored while called UNATTENDED.

mstormi: ok

mstormi:
I think remaining issues are resolved now.
But I had trouble merging, so better give it a second look please, I might have overlooked something.

EliasGabrielsson:
Thanks Marcus!
Nice that you put the effort into rewriting the code to directly call the `exim` function.
This ensure a more consistent end-user experience with less hassle. 👍

PR #449, opened 2018-10-02
refactored Amanda AWS backup option

mstormi:
refactored all of Amanda install routine

setting up AWS S3 backup didn't work at all, now it does (hopefully)

Signed-off-by: Markus Storm markus.storm@gmx.net (github: mstormi))


mstormi:
thanks for the quick merge.
Reminds me of #430  (as Amanda should be able to send mail to users as well but there's still no MTA).
Anything wrong with adding that ?

PR #455, opened 2018-10-23
Fix sed expression to remove blank lines from /etc/fstab

rohanl:
When first-boot.sh is run, I see an error in the log:

```
2018-10-23_11:50:18_NZDT [openHABian] Preparing openHAB folder mounts under /srv/... sed: -e expression #1, char 5: unterminated address regex
OK
```

I don't quite understand what is going on, but it seems "$#" is being parsed as something special, so the closing '#' is not recognised.

We can just revert to using the standard '/' delimiter


ThomDietrich:
Introduced in https://github.com/openhab/openhabian/commit/31043c9141fb2135995a4b130e5b22ff38033c55

I wonder why the issue comes up only now. You are right, the combination `$#` is parsed special. Not by sed but by bash itself.

> `$#` is the number of positional parameters.

Good catch. Thanks

Issue #457, opened 2018-10-26
Zulu install dependency on Ubuntu repo


mstormi:
I noticed 'on the go' that Zulu installation routine retrieves keys from an Ubuntu repo.
Now as unavailability of DNS, inet access or repos or changed keys may (and for some people do) result in installation failures, I wonder whether this is really needed or if it can be removed.
I don't know why it is in there (I'm guessing it's a leftover of earlier days when the packages were not yet directly downloaded from Azul Systems) and I'm unsure if there's another dependency I don't see.

BClark09:
The repository packages are signed, apt requires a key to validate the package.

Alternatively, the key can be obtained from the Azul site rather than a key repository with a specific hash.

mstormi:
Ok. Now given we're a little behind on Java (see also #453) and the fact that the Azul repo does not seem to get updated, shouldn't we move to a different installation method (wget download from https://www.azul.com/downloads/zulu-embedded/ instead of apt) ?
AFAIK those packages aren't signed, so we could remove validation, too.

BClark09:
Absolutely, looking at the documentation on Zulu, it seems that zulu-embedded dropped its support for using a repository some time ago.

Ideally, we'd need to change this so that it downloaded, extracted and set as default using `update-alternatives` or `update-java-alternatives`. The harder part, is pointing towards Azul's terms and conditions and grabbing the latest version of Java 8 programmatically.

mstormi:
Agree.
Right on the download page, Azul just says "By downloading software from this page, you agree to the Terms of Use." with "Terms of Use" being a link.
So we could simply prompt the user if he wants us to display the license and if so present him with that URL or its contents.

mstormi:
Created PR#459.

mstormi:
@BClark09 mind to check [PR #459](https://github.com/openhab/openhabian/pull/459)

BClark09:
Will do, but will only have time to properly check it in the weekend.

EliasGabrielsson:
@mstormi Do you have a source on the deprecation of the deb repository from Azul?
Otherwise, I assume that the rep is still maintained and we can close this issue and related PR:s.

mstormi:
No real source but the repo has not been updated in ages (6+ months) meanwhile they did create and offer new packages for download. I think it's reasonably safe to assume they don't maintain it any more and even if they did it would make sense to switch to package downloads as these are way more up to date and reliable.

EliasGabrielsson:
[...]

Issue #453, opened 2018-10-20
javax.net.ssl.SSLHandshakeException - handshake_failure


Markkuuss:
The Zulu Embedded Java (1.8.0_152) included with openhabian causes some problems with TLS handshake. There is a newer version (1.8.0_181) on the homepage https://www.azul.com/downloads/zulu-embedded/ meanwhile. This fixes these bugs, so this version should also be included in the openhabian package.

See also: https://community.openhab.org/t/solved-http-binding-javax-net-ssl-sslhandshakeexception-handshake-failure/54253/3?u=mas

mstormi:
We currently use the Azul repo for install, and seems _152 is the latest in there so we would need for that to get updated or rework the install system. Not sure if we wanna do the latter as there is no real need for _181. The TLS issue can also be fixed [by installing JCE](https://github.com/openhab/openhabian/issues/447) as well.

EliasGabrielsson:
fixed by updated java packages

Issue #450, opened 2018-10-03
Java 11


mstormi:
openHAB is to move to Java 11 soon, see [this](https://github.com/openhab/openhab-distro/issues/768) for current status.

We need to offer this to at least be an install option.

EliasGabrielsson:
Notice license change from oracle when doing this.

kaikreuzer:
> openHAB is to move to Java 11 soon

Not at all "soon" - we are just evaluating and trying to fix issues, so that IF someone uses Java 11 by accident can still use it. The recommendation will be Java 8 for the foreseeable future.

For openHABian, we use Zulu Embedded JVM, so Oracle licensing does not impact us.
There are no Zulu Embedded builds of Java 11 available, so this is another reason to stick to Java 8.

I'd suggest to only open an issue here once there is a decision to upgrade from 8 to anything newer.

mstormi:
I disagree a little here and I think you shouldn't close it as fast as you did.
First, the issue is meant to provide Java 11 just as an install _option_ (not default) and I think it makes sense to have that available in openHABian *before* OH moves to Java 11 so people can start testing early.

Also remind you openHABian runs (and is supposed to run) on non-ARM hardware as well, and there's Java 11 Zulu on x86 already available.
Also, until now Oracle is a supported alternative on both, ARM and non-ARM, so either we decide to discontinue support for Oracle or we mustn't ignore licensing issues.

kaikreuzer:
This issue made the impression as if there can/should be something done in openHABian. I don't agree that we should add features where we know that it is NOT working, just for the sake of testing.

> Also remind you openHABian runs (and is supposed to run) on non-ARM hardware as well

Ok, good point - I was so far only aware that we package Zulu embedded in the images (and I cannot see anything in the [features](https://github.com/openhab/openhabian/blob/master/docs/openhabian.md#optional-components) that mentions x86 JVMs as an install option).

> so either we decide to discontinue support for Oracle

Not sure who and why this has ever been added, but I'd be all for removing it indeed.

ThomDietrich:
Hey guys!

>> so either we decide to discontinue support for Oracle
>
> Not sure who and why this has ever been added, but I'd be all for removing it indeed.

The routine was working when we decided to switch over to Zulu. At this point I decided to offer the Oracle solution as an alternative, in case Zulu created issues for some. Never removed it since.
@mstormi feel free to remove the menu entry.

PR #459, opened 2018-11-02
Download and install Zulu Embedded Java from .tar

mstormi:
We need to change the installation method of Zulu Java as Azul systems does not maintain Java repos any more.

### DIFF
--- a/functions/java-jre.sh
+++ b/functions/java-jre.sh
@@ -17,7 +17,33 @@ java_webupd8() {
   if [ $? -eq 0 ]; then echo "OK"; else echo "FAILED"; exit 1; fi
 }
 
-java_zulu_embedded() {
+java_zulu() {
+  FILE="$(mktemp /tmp/.XXXXXXXXXX)"
+
+  if is_arm; then
+    JAVA=zulu8.31.1.122-jdk1.8.0_181-linux_aarch32hf
+  else
+    JAVA=zulu8.31.0.1-jdk8.0.181-linux_x64
+  fi
+  querytext="Downloading Zulu implies to agree to Azul Systems' Terms of Use. Display them now or proceed with downloading."
+  while ! (whiptail --title "Download Zulu Java" --yes-button "Download and Install" --no-button "Read Terms of Use" --defaultno --yesno "$querytext" 10 80) ; do
+    whiptail --textbox /opt/openhabian/docs/azul-zulu-license.md --scrolltext 27 116
+  done
+
+  wget -nv -O $FILE http://cdn.azul.com/zulu-embedded/bin/${JAVA}.tar.gz
+  cd /usr/lib/jvm; tar xpzf $FILE
+  cond_redirect update-alternatives --install /usr/bin/java java /usr/lib/jvm/${JAVA}/bin/java 1083000
+  cond_redirect update-alternatives --install /usr/bin/javac java /usr/lib/jvm/${JAVA}/bin/javac 1083000
+  rm -f $FILE
+
+  if [[ is_pi || is_pine64 ]]; then
+    cond_echo "Optimizing Java to run on low memory single board computers... "
+    sed -i 's#^EXTRA_JAVA_OPTS=.*#EXTRA_JAVA_OPTS="-Xms400m -Xmx512m"#g' /etc/default/openhab2
+  fi
+}
+
+# Unused
+java_zulu_embedded_archive() {
   echo -n "$(timestamp) [openHABian] Installing Zulu Embedded OpenJDK... "
   cond_redirect apt -y install dirmngr
   if is_arm; then arch="[arch=armhf]"; fi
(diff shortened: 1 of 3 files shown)


mstormi:
To fix https://github.com/openhab/openhabian/issues/457 and https://github.com/openhab/openhabian/issues/453.
See also https://github.com/openhab/openhabian/issues/450.

functions/java-jre.sh
```diff
@@ -17,7 +17,28 @@ java_webupd8() {
...
 }
 
-java_zulu_embedded() {
+java_zulu() {
+  FILE="$(mktemp /tmp/.XXXXXXXXXX)"
+
+  if is_arm; then
+    JAVA=zulu8.31.1.122-jdk1.8.0_181-linux_aarch32hf
+  else
+    JAVA=zulu8.31.0.1-jdk8.0.181-linux_x64
```
BClark09: Should we assume here that anything not arm is 64-bit?

functions/java-jre.sh
```diff
@@ -17,7 +17,28 @@ java_webupd8() {
...
+  fi
+  querytext="Downloading Zulu implies to agree to Azul Systems' Terms of Use. Display them now or proceed with downloading."
```
BClark09: Quick check from @kaikreuzer, is copying the terms of use and displaying them here sufficient in your opinion?

functions/java-jre.sh
```diff
@@ -17,7 +17,28 @@ java_webupd8() {
   if [ $? -eq 0 ]; then echo "OK"; else echo "FAILED"; exit 1; fi
 }
 
-java_zulu_embedded() {
+java_zulu() {
+  FILE="$(mktemp /tmp/.XXXXXXXXXX)"
```
BClark09: I don't think there is need to use the temporary file storage here. You make sure to delete the file anyway.

Some opt to make /tmp/ tmpfs, and the archive is > 100MB

functions/java-jre.sh
```diff
@@ -17,7 +17,28 @@ java_webupd8() {
...
+  else
+    JAVA=zulu8.31.0.1-jdk8.0.181-linux_x64
+  fi
+  querytext="Downloading Zulu implies to agree to Azul Systems' Terms of Use. Display them now or proceed with downloading."
+  while ! (whiptail --title "Download Zulu Java" --yes-button "Download and Install" --no-button "Read Terms of Use" --defaultno --yesno "$querytext" 10 80) ; do
+    whiptail --textbox /opt/openhabian/docs/azul-zulu-license.md --scrolltext 27 116
+  done
+
+  wget -nv -O $FILE http://cdn.azul.com/zulu-embedded/bin/${JAVA}.tar.gz
+  cd /usr/lib/jvm; tar xpzf $FILE
```
BClark09: I would usually use `/opt/jdk/` here as the archive contains binaries and libraries. 

You should also make sure that the path exists or is created before before running this command.

EliasGabrielsson:
Thanks @BClark09  for the reviewing!
I have some time this upcoming weekend for test and merge PR:s.

functions/java-jre.sh
mstormi: note  that as part of the install the user also has to confirm he read

functions/java-jre.sh
mstormi: We have no check for 32/64 bit and latest Zulu is not available in 64 bit on x86 so I think it is somewhat safe to assume that we _want_ this to be 64bit always, wdyt?

BClark09:
@mstormi would you be willing to rebase your PR or resolve the conflicts?

ThomDietrich:
Thanks ;) @BClark09 @EliasGabrielsson  anything else you want to comment on?

functions/java-jre.sh
```diff
@@ -17,7 +17,32 @@ java_webupd8() {
...
+  mkdir ${INSTALLROOT}
+  mkdir ${TEMPROOT}
+  
+  # Latest version, check https://www.azul.com/downloads/zulu-embedded
+  if is_arm; then
+    local JAVA=zulu8.33.0.134-jdk1.8.0_192-linux_aarch32hf
+  else
+    local JAVA=zulu8.33.0.1-jdk8.0.192-linux_x64
+  fi
+  whiptail --textbox $BASEDIR/includes/azul-zulu-license.md --scrolltext 27 116
```
ThomDietrich: "docs"? Please test your code before committing changes

mstormi: I don't understand what you mean ?

mstormi: Yes it was working, pointing to docs which is where I've put the license.
I moved it to includes now, that's how I understood your comment.

Do you want me remove Oracle as part of this PR, too ?

EliasGabrielsson:
The code does not take into account the case of `x86` architecture. Its need to be one method for installing `zulu8` and one concern `zulu8-embedded`.

Please refactor out two methods:

```
java_zulu() {
    main method deciding which package to use.
}
java_zulu_enterprise() {
   "old code" with embedded part removed
}
java_zulu_embedded() {
   Your new code introduced in this PR
}
```

mstormi:
it does, it's using the x86 file if not is_arm()
you still want me to refactor that ?

EliasGabrielsson:
> it does, it's using the x86 file if not is_arm()
> you still want me to refactor that ? Wouldn't make sense to me, that's more lines of code.

Compared to the `zulu8-embedded` software is `zulu8` still updated at the apt repository. That means we shall keep the current routine as that ensures updates are pushed out.

So there must be two methods even if it's more line of codes. 
I can see if I will get some time next weekend to code on this and help you out with that refactoring / part!

mstormi:
> Compared to the `zulu8-embedded` software is `zulu8` still updated at the apt repository.

Did you verify that ?

ThomDietrich:
@mstormi I just realized that this should be a good tradeoff solution:

```
curl -s https://www.azul.com/downloads/zulu-embedded  | grep -Eo "http://[a-zA-Z0-9./?=_-]*aarch32hf.tar.gz"
```
Of course you need to add a simple check to see if the result is an empty string and then this should be fine to use also for future releases

mstormi:
> @mstormi I just realized that this should be a good tradeoff solution:

Using the latest code isn't always the best choice as it can be buggy. Using older Java on the other hand does not do harm (every openHABian user has been installing u152 with the existing code).
I feel my current code is the safer bet.

ThomDietrich:
> Using the latest code isn't always the best choice as it can be buggy. 

😳 if Azul is confident enough to publish a new release on their webpage, we have to trust in their judgement and follow through. That is the way we follow with every apt repository we install packages from. Be it the deprecated azul repo or any other. Please reconsider.

mstormi:
I'm a bit more conservative there, maybe due to my job where I keep seeing this can often be a problem.
So no I don't want to change that so please don't ask me to. Why not merge my PR and if you really really want to change this then make another followup PR yourself.

EliasGabrielsson:
I also think Thomas's script is the way to go for `zulu8-embedded`. This reduce code maintenance similar to using an apt repository. As our voluntary resources are few we should devote them to more important stuff. I even think we shall push it even further and call this java installation function from the menu option `02 | Upgrade system` as well.

ThomDietrich:
Good point Elias.

@mstormi I understand your concern and your request for a merge. I agree, we should not keep this important PR waiting any longer. I'll give it a last review.

EliasGabrielsson:
A hint for next time please `rebase` on master instead of `merge` the code.
I also noticed that the script is broken for x86 installation as the download link is wrong, pointing to zulu-embedded instead of zulu.

ThomDietrich:
Could you clarify? In general it's almost always the better choice to "Squash merge" a PR, especially when it consists of multiple commits that you want to squash for the master branch.

About x86: 😕 Could one of you please create a PR?

EliasGabrielsson:
Don't meant to be picky when reviewing stuff but look at this commit message: https://github.com/openhab/openhabian/commit/6f6154dba39d5e28e3f4fc5be87631332bc305cf
Of course could we as maintainers go over everything by hand but I would prefer that the authors of PRs just do a `rebase` on master and squash theirs' commits.

ThomDietrich:
Ah now I get it. I admit, the history of this branch/PR is a bit messed up and I could have cleaned the commit message. On the other side I have to say that I hate it when contributors rewrite the history of a branch in the middle of the review process. Rewriting git history is evil - doing it during the merge into master is the best time for it.

If I would have to do it again, I'd again choose the squash merge option but actively reduce the commit message to its essence. Agreed?

PR #472, opened 2018-11-28
Change from apt to apt-get in build

EliasGabrielsson:


ThomDietrich:
why? Which archaic system are you on?? :D

EliasGabrielsson:
The API for apt wasn't stable so the script didn't work. The changed `--yes` option somehow :-)

PR #480, opened 2018-12-17
Update packages.sh to refer to current (v2) MQTT Binding

stefan-muc:


ThomDietrich:
Totally correct. Thanks for your addition! 👏 
I'm wondering if we should also advertise the new MQTT broker built into openHAB. Seems to me as this should be the preferred way. What do the others think?

stefan-muc:
You're welcome :)

Oh, there's an included brooker? Didn't know that :)
Generally yes I would advertise internal/included tools, but as far as I can see from https://community.openhab.org/t/mqtt-binding-version-2-4-pre-release/53811/258 it's not yet stable enough - I personally would wait for next release to have bugs fixed.
(Can somebody tell about which broker is used and which features it brings?)

EliasGabrielsson:
Shouldn't we deprecate this feature if there is an included broker within openhab?

ThomDietrich:
I agree with @stefan-muc, these are two solutions and in the current phase of potential transition we can offer both.

@stefan-muc https://github.com/andsel/moquette

ThomDietrich:
I believe this topic will become more actionable in the near future. The change in this PR is a clear improvement, no need to wait

PR #488, opened 2018-12-19
Tweaking wait for network in first-boot.sh

### LINKED ISSUE #441: add a delay and/or check the network is active prior to issue the apt-update
When installing using wifi, I have found that the install fails at apt-update.  I modified the first-boot.sh on line 80 adding a sleep and everything seemed to work great.  

Prior to adding the sleep, this command would routinely fail.  Hardware this failed and was tested on was Raspberri Pi 3 model b.  

[first-boot.log](https://github.com/openhab/openhabian/files/2384857/first-boot.log)

[first-boot.sh.txt](https://github.com/openhab/openhabian/files/2384868/first-boot.sh.txt)

It looks as if the logic here: 
`echo -n "$(timestamp) [openHABian] Waiting for dpkg/apt to get ready... "
until apt update &>/dev/null; do sleep 1; done
echo "OK"
`
Is this supposed to possibly check for network to be online?  If so there seems to be some error doing so.  However putting a sleep in solved the problem.  **Please let me know** if you need more information or would like me to make any modifications.  I attached my modified file that worked for me.

EliasGabrielsson:
resolves #441

### DIFF
--- a/build-image/first-boot.sh
+++ b/build-image/first-boot.sh
@@ -123,6 +123,7 @@ echo "OK"
 
 echo -n "$(timestamp) [openHABian] Waiting for dpkg/apt to get ready... "
 until apt update &>/dev/null; do sleep 1; done
+sleep 10
 echo "OK"
 
 echo -n "$(timestamp) [openHABian] Updating repositories and upgrading installed packages... "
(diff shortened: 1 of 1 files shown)


ThomDietrich:
Merging with the option/hope to remove at a later point when a better solution was found

Issue #495, opened 2018-12-30
Grafana setup always tries to install armhf instead of actual architecture


GruensFroeschli:
Clean debian stretch installation.

GruensFroeschli:
functions.sh
influxdb_grafana_setup()
Lines 608 and 614
The checks on the architecture do nothing because this is not a bash "if" statement but a value assignment.

EliasGabrielsson:
Thanks again for fixing this 👍 
Closing this now.

PR #496, opened 2018-12-30
Fix bash syntax errors (#495)

GruensFroeschli:
Because of this bug, other architectures (in the debugged case amd64)
are not able to install because the code always tries to install armhf
which is not possible.

Signed-off-by: Matthias May <github@may.nu>


EliasGabrielsson:
Thanks for fix @GruensFroeschli

PR #497, opened 2019-01-01
Update openhabian.md

kobold81:
Hello,

due to a little crash of my raspberry, I was forced to re-read the installation manuals. Afterwards, I was beyond confused and missed that openHABian is not only available on the pre-packagedSD images for the raspberry but can be installed an only linux machine.

Currently, the "Installation Guide" is structured by 
- OS (Windows, Linux, macOS),
- device (Raspberry, PINE, QNAP)
- container platform
- openhab package type (openHABian, package repository version, manual install)

Either you have time or are smart enough to understand this 4-dimensional matrix or you just wave the white flag.

The objective of the changes in my proposal is to shift the attention of the end user to the most simple way to install openhab without rewriting the whole manual.

Best regards
Marc

### DIFF
--- a/docs/openhabian.md
+++ b/docs/openhabian.md
@@ -68,7 +68,7 @@ Additionally the **openHABian Configuration Tool** [`openhabian-config`](#openha
 
 Here you'll find supported and tested installation platforms and instructions.
 
-### Raspberry Pi
+### Raspberry Pi (Prepackaged SD-Card Image)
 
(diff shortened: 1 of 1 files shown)


EliasGabrielsson:
Hi Marc and thanks for your feedback.
I am looking into reviewing the documentation for the 1.5v release coming up this spring, soon™.
If you have any more thoughts you are more then welcome to share them. 👍

PR #502, opened 2019-01-05
Allow to add necessary headers for CORS to work

### LINKED ISSUE #501: [Nginx][CORS] Allow to add necessary headers for CORS to work
For [Cross-Origin Resource Sharing](http://www.w3.org/TR/cors/) to work (access of the openHAB REST API from another origin) NGINX need some additional headers:

```
server {
  add_header 'Access-Control-Allow_Credentials' 'true';
  add_header 'Access-Control-Allow-Headers' 'Authorization,Accept,Origin,DNT,X-CustomHeader,Keep-Alive,User-Agent,X-Requested-With,If-Modified-Since,Cache-Control,Content-Type,Content-Range,Range';
  add_header 'Access-Control-Allow-Methods' 'GET,POST,OPTIONS,PUT,DELETE,PATCH';
  ...
}
```

**Why is this important**: If you develop Web GUIs (Paper UI) etc, you usually start them on an embedded node based webserver, so the port differs to the one from openHAB. This results in a CORS access error by modern browsers.

davidgraeff:
Fixes #501

Signed-of-by: davidgraeff <david.graeff@web.de>

### DIFF
--- a/includes/nginx.conf
+++ b/includes/nginx.conf
@@ -16,6 +16,12 @@ server {
     server_name                     DOMAINNAME;
 #   add_header                      Strict-Transport-Security "max-age=31536000; includeSubDomains";
 
+    # Cross-Origin Resource Sharing.
+    # add_header 'Access-Control-Allow-Origin' 'http://localhost:8080/rest';
+    add_header 'Access-Control-Allow_Credentials' 'true';
+    add_header 'Access-Control-Allow-Headers' 'Authorization,Accept,Origin,DNT,X-CustomHeader,Keep-Alive,User-Agent,X-Requested-With,If-Modified-Since,Cache-Control,Content-Type,Content-Range,Range';
+    add_header 'Access-Control-Allow-Methods' 'GET,POST,OPTIONS,PUT,DELETE,PATCH';
+    
 ## Secure Certificate Locations
 #   ssl_certificate                 CERTPATH;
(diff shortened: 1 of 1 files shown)


EliasGabrielsson:
Thanks David!

BClark09:
Hi Both, sorry for being quite late to seeing this PR. I have noticed that this PR will have broken NGINX setup for authentication and SSL configuration.

Since there are new lines in this file, pointers to the commented lines should have been changed on lines [793](https://github.com/openhab/openhabian/blob/master/functions/packages.sh#L793), [820](https://github.com/openhab/openhabian/blob/master/functions/packages.sh#L820), and [836](https://github.com/openhab/openhabian/blob/master/functions/packages.sh#L836) of `packages.sh`.

@davidgraeff would you be able to alter these numbers in your new PR?

There's definitely a better way of dynamically changing the template rather than using magic line numbers. I can find a better way at a later date.

davidgraeff:
Hm. I think its better to revert this commit instead and change to another solution in `packages.sh` first.

There might be more coming for the nginx config file and I doubt that we will always think of adapting all line numbers.

BClark09:
I'm working on the solution which currently doesn't include the changes in this PR:

The change uses comment "tags" rather than relying on magic line numbers.

https://github.com/openhab/openhabian/compare/master...BClark09:nginx-config?expand=1

PR #507, opened 2019-01-08
reduce Java mem settings

mstormi:

### DIFF
--- a/functions/openhab.sh
+++ b/functions/openhab.sh
@@ -71,7 +71,7 @@ Check the \"openHAB Release Notes\" and the official announcements to learn abou
 
   if [[ is_pi || is_pine64 ]]; then
     cond_echo "Optimizing Java to run on low memory single board computers... "
-    sed -i 's#^EXTRA_JAVA_OPTS=.*#EXTRA_JAVA_OPTS="-Xms400m -Xmx512m"#g' /etc/default/openhab2
+    sed -i 's#^EXTRA_JAVA_OPTS=.*#EXTRA_JAVA_OPTS="-Xms250m -Xmx350m"#g' /etc/default/openhab2
   fi
 
   if [ -n "$INTERACTIVE" ]; then


mstormi:
Looking at people to run additional program next to openHAB, we should pre-allocate enough but as little as possible memory to Java.
I've successfully been running OH2.4 with these settings 200/256 was too few and made me run into garbage collection again which this setting is meant to avoid.

EliasGabrielsson:
Great to see you put effort into profiling and optimizing this. 
Just for curiosity how much does things and items influence the memory footprint?

ThomDietrich:
I suspect most openHABian users without additional (demanding) software aside openHAB, If that is the case, why would we restrict memory for it? HOw would a user identify issues resulting out of this restriction?
I like the idea here but we should be careful to not over-optimize. Could you provide the method you used to get to these numbers? Thanks ;)

mstormi:
Essentially I played with my own installation. It's quite large and complex as you can imagine, so a mem setting to make that run smoothly will likely do so for the vast majority of people.
The first parameter is the preallocated heap size. I start seeing impact (garbage collection, i.e. rarely executed rules take a couple of seconds to reload) when I reduce that by 50M.
For the second parameter Xmx, 256M is the JVM default limit, i.e. Java will crash when hit and we should have heard of complaints if that was the case for some users but I didn't hear any. And my setting now means another safety margin of 100M.
So I think these are reasonable values for a default. I bet the number of openHABian users to also co-install Grafana or similar memory hogs and to then run into trouble because of that is a lot higher than the number of people to likely run into trouble with these settings. Let's not make a science of it, anyone can still increase that if needed.

mstormi:
>  Just for the sake of curiosity how much does things and items influence the memory footprint?

I didn't really notice any. It's rather the rules and eventually BLOBs such as images or MP3 sounds to have a noticeable impact.

PR #512, opened 2019-01-11
Display Zulu license in interactive mode only

mstormi:
next try ...

### DIFF
--- a/functions/java-jre.sh
+++ b/functions/java-jre.sh
@@ -31,7 +31,9 @@ java_zulu() {
     # Latest version check https://www.azul.com/downloads/zulu-linux
     local JAVA=zulu8.33.0.1-jdk8.0.192-linux_x64
   fi
-  whiptail --textbox $BASEDIR/includes/azul_zulu_license.md --scrolltext 27 116
+  if [ -n "$INTERACTIVE" ]; then
+    whiptail --textbox $BASEDIR/includes/azul_zulu_license.md --scrolltext 27 116
+  fi
   
   cond_redirect wget -nv -O $FILE http://cdn.azul.com/zulu-embedded/bin/${JAVA}.tar.gz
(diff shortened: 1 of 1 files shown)


EliasGabrielsson:
Is this actually tested?

EliasGabrielsson:
There is definitely need for improvement in the space of testing. 
I have though if it would be possible to do a two step testing pipeline including both unit and integration tests. The unit test could consist of shellcheck and testing of individual install functions. For this to work we need to unify a parameter structure and document it for the functions. Kind of what UNATTENDED does today.

I have started to elaborate a little on this in: https://github.com/openhab/openhabian/blob/b65b2034d580ce1b75214e0ee08438d2df4d1abe/functions/packages.sh#L409 and https://github.com/openhab/openhabian/blob/b65b2034d580ce1b75214e0ee08438d2df4d1abe/functions/packages.sh#L855

We also need to have a "unit" test machine that we can somehow build up every time we do the test and testcases of course. There are actually some framework dealing with bash testing. A good start would be to test against a vanilla debian stretch images. Maybe use docker as a start which could speed up things a bit, I used https://drone.io/ before for that case.

The integration test would be performed on real hardware using same configuration files for the unit test somehow. Eg. A regular installation which also perform a suit of test cases. This can be combined with the WebIF as well. Maybe a "regular" installation is testing itself when installed somehow using the same testcase we use for testing and report the result in WebIF. 

The integration testing also enables us to have a `production` branch beside `master` which all installation could pull from.

ThomDietrich:
Having a VM/container executed for a full setup test was also on my mind a few times.

I'm still in the middle of moving in but I'm coming back to the openHAB development and I wanted to initiate a discussion about the future of openHABian. This should be one aspect of it. Would you like to move your comment into its own issue?

Issue #524, opened 2019-01-29
Special device /dev/mapper/loop0p1 does not exist


uadnan:
If there is already some images mounted using `kpartx` build.sh fails with an error message

```
special device /dev/mapper/loop0p1 does not exist
```

PR #525, opened 2019-01-29
Dynamically select loopXX

uadnan:
Fix for #524


EliasGabrielsson:
Thanks for contributing!

PR #528, opened 2019-02-06
Another CORS fix

davidgraeff:
openHAB sometimes respond with a 400-type http error code (if you send a command to an item and the item type doesn't allow that command for instance). So CORS must also work for 400-type codes. Added "always" therefore.

Signed-off-by: davidgraeff <david.graeff@web.de>

### DIFF
--- a/includes/nginx.conf
+++ b/includes/nginx.conf
@@ -18,9 +18,9 @@ server {
 
     # Cross-Origin Resource Sharing.
     add_header 'Access-Control-Allow-Origin' '*' always; # "Always" makes sure that also a 400-type response works
-    add_header 'Access-Control-Allow_Credentials' 'true';
-    add_header 'Access-Control-Allow-Headers' 'Authorization,Accept,Origin,DNT,X-CustomHeader,Keep-Alive,User-Agent,X-Requested-With,If-Modified-Since,Cache-Control,Content-Type,Content-Range,Range';
-    add_header 'Access-Control-Allow-Methods' 'GET,POST,OPTIONS,PUT,DELETE,PATCH';
+    add_header 'Access-Control-Allow_Credentials' 'true' always;
+    add_header 'Access-Control-Allow-Headers' 'Authorization,Accept,Origin,DNT,X-CustomHeader,Keep-Alive,User-Agent,X-Requested-With,If-Modified-Since,Cache-Control,Content-Type,Content-Range,Range' always;
+    add_header 'Access-Control-Allow-Methods' 'GET,POST,OPTIONS,PUT,DELETE,PATCH' always;
     
 ## Secure Certificate Locations
 #   ssl_certificate                 CERTPATH;
--- a/includes/nginx.conf
+++ b/includes/nginx.conf
@@ -17,7 +17,7 @@ server {
 #   add_header                      Strict-Transport-Security "max-age=31536000; includeSubDomains";
 
     # Cross-Origin Resource Sharing.
-    # add_header 'Access-Control-Allow-Origin' 'http://localhost:8080/rest';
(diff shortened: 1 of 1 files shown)


includes/nginx.conf
```diff
@@ -17,10 +17,10 @@ server {
 #   add_header                      Strict-Transport-Security "max-age=31536000; includeSubDomains";
 
     # Cross-Origin Resource Sharing.
-    # add_header 'Access-Control-Allow-Origin' 'http://localhost:8080/rest';
-    add_header 'Access-Control-Allow_Credentials' 'true';
-    add_header 'Access-Control-Allow-Headers' 'Authorization,Accept,Origin,DNT,X-CustomHeader,Keep-Alive,User-Agent,X-Requested-With,If-Modified-Since,Cache-Control,Content-Type,Content-Range,Range';
-    add_header 'Access-Control-Allow-Methods' 'GET,POST,OPTIONS,PUT,DELETE,PATCH';
+    add_header 'Access-Control-Allow-Origin' '*' always; # "Always" makes sure that also a 400-type response works
```
BClark09: Is the wildcard here necessary? If so, shouldn't we have this off by default and configure this on the setup page?

davidgraeff: I thought about that. But the average non-techie user will struggle with new REST based interfaces if CORS is off by default.

CORS is a broken concept anyway in our case, you can always just use a server side access instead of a browser/fetch.
CORS is basically just preventing frontend developers to do stupid things, like storing credentials in the browser when the remote API was actually meant to be used by backend servers.

In our case though we want UIs to access openHABs REST interface, at any given time. Authentication is another feature that is unrelated to CORS.

BClark09: I read that we will also run into problems with some browsers and setups because of the two conflicting headers:

https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS/Errors/CORSNotSupportingCredentials

Can we construct the the header using regex within NGINX? I found: https://stackoverflow.com/questions/14499320/how-to-properly-setup-nginx-access-control-allow-origin-into-response-header-bas/29113949#29113949

Admittedly, this is not a strong area for me: Is there a possibility of setting this in the REST implementation?

davidgraeff: > Can we construct the the header using regex within NGINX?

Not sure. It is important that the CORS origin is not only "localhost" or "127.0.0.1" etc, but contains the real IP (IPs for multiple interfaces). And "*" is just the easy workaround. I didn't know that it at the same breaks HTTP Auth.

BClark09: It may be easier to have an option in `openhabian-config`, to add to a list of allowed origins. Are there any current examples of cross-origin interfaces?

davidgraeff: Sure. My work on the Paper UI NG design study. If you use it straight from the github hosted webpage, you need CORS on OH for it to work.

EliasGabrielsson:
Wouldn't it be possible to set this dynamically. Eg.

1. Set `localhost` and `127.0.0.1` as static config to start with.
2. Add local ip address such as `192.168.0.0` or similar
3. Get our `public IP` like we do here: https://github.com/openhab/openhabian/blob/c38c029fa722ff84198e9ee4a3a0e25cbd134b87/functions/packages.sh#L778 
(This should preferable be refactored to a helper function `get_public_ip()` so we can unit test it.)
4. Do a reverse lookup on public and local IP, add result to config file.
5. Display result for the user and ask if they miss any domain and give option to add it.

EliasGabrielsson:
I am all go on adding `always` to allow 400 answers.

davidgraeff:
Sounds like a solution to allow localhost as a start and dynamically extend it.

PR #535, opened 2019-02-23
fix adminmail in amanda_setup when exim already installed

mstormi:
Signed-off-by: Markus Storm [markus.storm@gmx.net](mailto:markus.storm@gmx.net) (github: mstormi)

### DIFF
--- a/functions/backup.sh
+++ b/functions/backup.sh
@@ -58,7 +58,6 @@ create_backup_config() {
     fi
   fi
 
-#   /bin/sed -e "s|%CONFIG|${config}|g" -e "s|%CONFDIR|${confdir}|g" -e "s|%BKPDIR|${bkpdir}|g" -e "s|%ADMIN|${adminmail}|g" -e "s|%TAPES|${tapes}|g" -e "s|%SIZE|${size}|g" -e "s|%TAPETYPE|${tapetype}|g" -e "s|%TPCHANGER|${tpchanger}|g" ${BASEDIR}/includes/amanda.conf_template >${confdir}/amanda.conf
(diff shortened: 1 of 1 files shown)


BClark09:
This solves the error and fixes the function to what looks like the intended functionality, but I had a quick question before merge:

Should the user be able to specify `adminmail` if exim is already installed? Currently if exim exists, then the script keeps `adminmail` to `root@<hostname>`.

functions/backup.sh
```diff
@@ -133,14 +133,13 @@ amanda_setup() {
   fi
 
   /usr/sbin/exim --version 2>/dev/null >/dev/null
+  local adminmail="root@$(/bin/hostname)"
```
BClark09: While `local` declarations and assignments on the same line should mask error codes. This line should be above the previous line to make absolutely clear that we are assessing the return code for `/usr/sbin/exim` and not `/bin/hostname` in the next few lines.

functions/backup.sh
```diff
@@ -133,14 +133,13 @@ amanda_setup() {
...
   if [ $? -ne 0 ]; then
      if [ -n "$INTERACTIVE" ]; then
         if (whiptail --title "No exim mail transfer agent" --yes-button "Install EXIM4" --no-button "MTA already exist, ignore installation" --defaultyes --yesno "Seems exim is not installed as a mail transfer agent.\nAmanda needs one to be able to send emails.\nOnly choose to ignore if you know there's a working mail transfer agent other than exim on your system.\nDo you want to continue with EXIM4 installation ?" 15 80) then
            exim_setup
         fi
-        local adminmail=$(whiptail --title "Admin reports" --inputbox "Enter the email address to send backup reports to." 10 60 3>&1 1>&2 2>&3)
-     else
-        local adminmail="root@$(/bin/hostname)"
+        adminmail=$(whiptail --title "Admin reports" --inputbox "Enter the email address to send backup reports to." 10 60 3>&1 1>&2 2>&3)
```
BClark09: Thought I'd write my comment here too, should the user be able to set `adminmail` if exim already existed?

mstormi: yes he should.
See changed code. That now will also avoid the error to happen if a user enters an empty mail address.

BClark09:
Could you sign off the commit?

mstormi:
forgot to also write admin address into amanda.conf => https://github.com/openhab/openhabian/pull/536

PR #541, opened 2019-03-09
add ip capabilities to arping

mstormi:
arping needs to be executed as either root or with these capabilities to have proper access to the network to work. openHAB network binding to use it won't work without.

### DIFF
--- a/functions/system.sh
+++ b/functions/system.sh
@@ -192,6 +192,7 @@ permissions_corrections() {
 misc_system_settings() {
   echo -n "$(timestamp) [openHABian] Applying miscellaneous system settings... "
   cond_redirect setcap 'cap_net_raw,cap_net_admin=+eip cap_net_bind_service=+ep' $(realpath /usr/bin/java)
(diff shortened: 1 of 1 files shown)


EliasGabrielsson:
Good addon Marcus.

Please rewrite it to keep existing permission, i.e using the "+" sign.
I think this line will do:
`cond_redirect setcap 'cap_net_raw,cap_net_admin=+eip cap_net_bind_service=+ep' /usr/sbin/arping`

mstormi:
Well it does not have any capabilities before so it should be the same effect.
But I don't mind and changed it.

PS: but please find & fix the bug in 'Marcus', that I do mind :-)

EliasGabrielsson:
It is correct that no capabilities are added in our installation but the script can be invoked from an already installed system as well.

Are there any reasons you omitted the `cond_redirect` part of the script?

mstormi:
err no, I've put it back in so you can merge

PR #543, opened 2019-03-10
[RFC] CI Pipeline + Updated Java package

EliasGabrielsson:
Added an initial CI pipeline and updated the java installation.
See individual commit messages for details. Would like some input on this PR so we get things right.
(Maybe should this PR be splitted up in two parts, CI + java, but it was nice to use the test framework while developing the java part.)

The Java part currently lacks:
* Method declaration
* Graceful shutdown and startup of openHAB while updating (or any Java application).
* Have not been tested on hardware which is important because that is the most changed code.
* Does not display the license for the end-user

Please have a look. 
 @mstormi @ThomDietrich @BClark09


build-image/first-boot.bash
```diff
@@ -129,36 +133,30 @@ echo "OK"
 echo -n "$(timestamp) [openHABian] Updating repositories and upgrading installed packages... "
 apt update &>/dev/null
 apt --yes upgrade &>/dev/null
-if [ $? -eq 0 ]; then
-  echo "OK";
-else
-  dpkg --configure -a
```
EliasGabrielsson: Needed to remove this code to get it running in the VM container at Travis, don't ask me why. 
Did this code solve a specific edge case? @ThomDietrich

EliasGabrielsson:
Have a look at the updated readme in the PR to get a grip of how the CI pipeline works.

EliasGabrielsson:
Superseded by https://github.com/openhab/openhabian/pull/547 to trigger Travis builds

mstormi:
Hi @EliasGabrielsson 

I tried to follow. First, bear with me since I'm new to Docker, Travis and BATS.
Right the first docker build command failed:
Step 2/7 : RUN apt-get update && apt-get install -y git locales
 ---> Running in a398a6f294cf
standard_init_linux.go:207: exec user process caused "exec format error"

The reason probably is
pi@devpi:~/openhabian $ docker inspect minimum2scp/systemd-stretch:latest|grep Arc
        "Architecture": "amd64",

docker is said to auto-select architecture and official images are multi-arch, but seems
minimum2scp/systemd-stretch is amd64 only.
So you are probably running it on x86? I tried on my RPi.

Would we need to select a different base system ?

PR #545, opened 2019-03-12
Add braces to 10th parameter on backup.sh

BClark09:
Bash interprets `$10` as `${1}0` which is not what we want here.


mstormi:
Duh. Good catch. One is never done learning.

PR #552, opened 2019-03-20
Fixed ln link

EliasGabrielsson:


EliasGabrielsson:
This will not handle new installed systems with the old images floating around. Closing

PR #567, opened 2019-05-05
Install 32-bit version of Java

EliasGabrielsson:
* Add support for x86 architecture for old hardware
* Install 32-bit version of Java which is recommended for OpenHAB

### DIFF
--- a/functions/java-jre.bash
+++ b/functions/java-jre.bash
@@ -49,7 +49,7 @@ java_zulu(){
     jdkInstallLocation="/opt/jdk"
     file="/var/tmp/.zulu.$$"
 
-    if $1; then # if argument 64-bit is true
+    if $1 && is_arm; then # if argument 64-bit is true
       echo -n "$(timestamp) [openHABian] Installing Zulu Embedded 64-Bit OpenJDK... "
       downloadPath=$(curl -s https://www.azul.com/downloads/zulu-embedded | grep -Eo "http://[a-zA-Z0-9./?=_-]*zulu8[a-zA-Z0-9./?=_-]*aarch64.tar.gz")
       archName=aarch64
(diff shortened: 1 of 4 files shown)


EliasGabrielsson:
$arch variable not set

functions/java-jre.bash
```diff
@@ -31,12 +49,35 @@ java_zulu(){
...
+    else
+      if is_x86_64; then 
+        dpkg --add-architecture i386
+        dpkg --add-architecture i686
+        cond_redirect apt update
+        cond_redirect apt -y install libc6:i386 libncurses5:i386 libstdc++6:i386
+      fi
+      echo -n "$(timestamp) [openHABian] Installing Zulu Enterprise 32-Bit OpenJDK... "
+      downloadPath=$(curl -s https://www.azul.com/downloads/zulu | grep -Eo "https://[a-zA-Z0-9./?=_-]*/zulu8[a-zA-Z0-9./?=_-]*i686.tar.gz" | head -n 1)
+      archName=i686
```
EliasGabrielsson: Is used as i386 in /etc/ld.so.conf.d case

openhabian-setup.sh
```diff
@@ -67,7 +67,7 @@ if [[ -n "$UNATTENDED" ]]; then
   bashrc_copy
   vimrc_copy
   firemotd_setup
-  java_zulu
+  java_zulu false
```
ThomDietrich: "false" isn't really intuitive. how about a verbal argument like "force-64bit" or similar

Besides I wonder if it wouldn't make sense to divide the huge function into smaller ones!?

EliasGabrielsson: Totally agree in both cases, period.

PR #572, opened 2019-05-10
Enable zram compression for swap

mstormi:

### DIFF
--- a/functions/system.bash
+++ b/functions/system.bash
@@ -190,6 +190,19 @@ permissions_corrections() {
   if [ $? -eq 0 ]; then echo "OK"; else echo "FAILED"; fi
 }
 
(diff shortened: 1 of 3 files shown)


mstormi:
Hi @EliasGabrielsson @ThomDietrich 

this patch is to enable zram swap so the box will rather compress pages and keep them in memory rather than swaping them to disk (SD card usually with the danger of corrupting that due to wearout))

mstormi:
Hi @EliasGabrielsson 
would you mind giving me write access to the openhabian repo ? So I could e.g. assign issues to myself.
best regards
Markus

mstormi:
I dropped this in favor of #576 . That one will also use zram for directories that are actively written to.

PR #575, opened 2019-06-12
run apt --fix-broken

mstormi:
See https://community.openhab.org/t/solved-openhabian-updating-repositories-and-upgrading-installed-packages-failed/36071/13

### DIFF
--- a/build-image/first-boot.bash
+++ b/build-image/first-boot.bash
@@ -132,6 +132,7 @@ echo "OK"
 
 echo -n "$(timestamp) [openHABian] Updating repositories and upgrading installed packages... "
 apt update &>/dev/null
+/usr/bin/apt --yes --fix-broken install &>/dev/null
 apt --yes upgrade &>/dev/null
 if [ $? -eq 0 ]; then echo "OK"; else echo "FAILED"; fail_inprogress; fi
 


mstormi:
should fix https://github.com/openhab/openhabian/issues/577

EliasGabrielsson:
Thanks Marcus!
When we touching this files I thinks it's a good idé to migrate to `apt-get` instead as `apt` is not designed to be stable for scripts, see cited man below:

```
 The apt(8) commandline is designed as an end-user tool and it may change behavior between
 versions. While it tries not to break backward compatibility this is not guaranteed either
 if a change seems beneficial for interactive use.

 All features of apt(8) are available in dedicated APT tools like apt-get(8) and apt-
 cache(8) as well.  apt(8) just changes the default value of some options (see apt.conf(5)
 and specifically the Binary scope). So you should prefer using these commands (potentially
 with some additional options enabled) in your scripts as they keep backward compatibility
 as much as possible.
```
http://manpages.ubuntu.com/manpages/xenial/man8/apt.8.html

EliasGabrielsson:
I do not know what is correspondent flag of `--fix-broken` is though.

build-image/first-boot.bash
```diff
@@ -132,6 +132,7 @@ echo "OK"
 
 echo -n "$(timestamp) [openHABian] Updating repositories and upgrading installed packages... "
 apt update &>/dev/null
+/usr/bin/apt --yes --fix-broken install &>/dev/null
 apt --yes upgrade &>/dev/null
```
ThomDietrich: Maybe this is just me but I believe that differences like the superfluous `/usr/bin/` make this code worse (error handling, readability, maintainability) than it needs to be.

mstormi: Yep that's just you :) It _should be_ superfluous but nevertheless it's best practice to use the full path  as $PATH isn't necessarily set or you can have multiple binaries of the same name in your path or even shell builtins such as 'echo'.

Siox0911:
A question: Why did you write **/usr/bin/apt** and not **apt** like the commands before and later?

build-image/first-boot.bash
ThomDietrich: My point was that we should keep it unified. Your single full path doesn't add any value as long as there are two short version "apt"s before and after it. It just messes up the code ;)

Regarding your argument I actually kind of disagree. I know this is an established best practice but I believe the following: openHABian either controls the system or works closely with it. Guessing where a certain command is located is redundant in that context and doesn't really add any value.
Anyhow I'm not making a strong argument here, nor would I be strictly against changing the way we do things in this project.
I am however against random deviations from the established style... :)
Jm2c, please don't take it the wrong way ;)

Issue #463, opened 2018-11-09
Inform user that reboot is required after openhabian.conf is changed


bodiroga:
Hi @ThomDietrich and guys!

I have noticed that, if I change the hostname in the openhabian.conf file right after flashing the SD card, the device doesn't change its hostname until a manual reboot is done. That's very confusing for new and inexperienced users, because they try to access their openHAB installation in the new URL: http://<new_hostname>.local:8080, but it doesn't work because the old http://openhabianpi.local:8080 url applies.

How could we address this? Could we store the "reboot requirement" information during the installation process and reboot the Raspberry Pi at the end of the process?

Many thanks and best regards,

Aitor

mstormi:
Mmh, I don't see this to be a bug. If a user _manually_ changes hostname (instead of using the openhabian-setup menu) he should know what he's doing and how to proceed.
We cannot cater for any off-mainstream modifications.

EliasGabrielsson:
You are technical correct Markus it's not a technical bug, but from a user perspective someone is confused about it.
What do you think off adding an information line in the config file?

mstormi:
ok, I went for b)
https://github.com/openhab/openhabian/pull/582

PR #582, opened 2019-06-24
add comments in openhabian.conf that it mustn't be used to change config

mstormi:
"fixes" #463 

### DIFF
--- a/openhabian.conf.dist
+++ b/openhabian.conf.dist
@@ -1,5 +1,7 @@
 # Modify your openHABian settings
 # Handle with care, only touch if you need to!!
+# This file will only be used on initial install.
+# If you want to change anything after install completed, use menu in openhabian-config.
 
 hostname=openHABianDevice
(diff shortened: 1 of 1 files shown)


EliasGabrielsson:
Aren't the hostname/ip updated on each reboot from this file?
This is not reflected in the information text.

Do you mind looking into how it exactly works out, ie. if the file have no technical functionality after first install it shall rather be deleted instead.

EliasGabrielsson:
The file is only used at first installation so it could technically be removed.
But maybe is there still a usage to keep it for debug purposes. 

I will merge this for now, it solves the problem at least. 👍

ThomDietrich:
The config file was definitely meant to hold various details about the setup relevant for openhabian-config even after the initial setup. See also https://github.com/openhab/openhabian/blob/master/functions/config.bash
Which variables from inside the file are actually used in another part of the tool, I would need to check.

PR #583, opened 2019-06-26
Use apt-get instead of apt where appropirate

### LINKED ISSUE #580: Migrate all usage of "apt" to "apt-get"
The `apt` command is not long-term stable according to manual:

```
 The apt(8) commandline is designed as an end-user tool and it may change behavior between
 versions. While it tries not to break backward compatibility this is not guaranteed either
 if a change seems beneficial for interactive use.

 All features of apt(8) are available in dedicated APT tools like apt-get(8) and apt-
 cache(8) as well.  apt(8) just changes the default value of some options (see apt.conf(5)
 and specifically the Binary scope). So you should prefer using these commands (potentially
 with some additional options enabled) in your scripts as they keep backward compatibility
 as much as possible.
```
http://manpages.ubuntu.com/manpages/xenial/man8/apt.8.html

holgerfriedrich:
Closes #580.


holgerfriedrich:
- building the rpi image checked.
- booting and first install on rpi - still open.
- building the pine64 image is broken on master (build-pine64-image.bash to be replaced by build-pine64-image.sh in build.bash)

EliasGabrielsson:
Hi! Nice refactoring 👍 
Have you reviewed every single line manually to check that eventual flags and argument are reflected in the `apt-get` case? 

Cheers

holgerfriedrich:
Hi @EliasGabrielsson, 
I have just briefly reviewed the calls to apt and the parameters. For me it looks pretty much like standard parameters. Not checked if there is some subtle difference.
Though, output of the scripts looks different (less verbose/user friendly).

Still, testing a fresh install on RPI not done yet.

Best, Holger

holgerfriedrich:
@EliasGabrielsson tested on RPI3B+ on top of PR #591. Firsttime install is fine.
Can we merge this? I think testing towards the new image for Buster would identify any remaining issues.

EliasGabrielsson:
Thanks so much for ironing out these issues and I warmly welcome you as contributor to openhabian.

ThomDietrich:
Thanks for this quick and nice PR!
I also welcome you to the openHABian team! I'm super happy to see more people joining the cause :)

PR #585, opened 2019-06-28
fix for building the pine64 image

holgerfriedrich:
pine64: change build-pine64.image.bash to .sh to match file from longsleep/build-pine64-image

### DIFF
--- a/build.bash
+++ b/build.bash
@@ -162,7 +162,7 @@ if [ "$hw_platform" == "pine64-xenial" ]; then
   fi
 
   echo_process "Hacking \"build-pine64-image\" build and make script... "
-  sed -i "s/date +%Y%m%d_%H%M%S_%Z/date +%Y%m%d%H/" $buildfolder/build-pine64-image.bash
+  sed -i "s/date +%Y%m%d_%H%M%S_%Z/date +%Y%m%d%H/" $buildfolder/build-pine64-image.sh
   makescript=$buildfolder/simpleimage/make_rootfs.sh
   sed -i "s/^pine64$/openHABianPine64/" $makescript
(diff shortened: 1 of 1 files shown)


EliasGabrielsson:
Thanks for fixing this. Some refactoring went to fast last time apparently ;-)

ThomDietrich:
What's the current status with the Pine64? I've invested quite some time into this image back when it first came out but no one really cared about it ever since - as far as I could tell. Do you believe it's still needed besides the manual setup option? Maybe instead of providing images for two devices, how about an extension of the documentation that mentions manual setup for all tested devices. I guess that would be just as easy for most users and it would highlight the wider support AND it would make the fact that device xyz can be used for openHABian/openHAB.

What do you guys think?

holgerfriedrich:
Hi @ThomDietrich,

I cannot comment on the Pine64, I have no hardware at hand.
Looking at the statistics at BinTray https://bintray.com/openhab/bin/openHABian#statistics there is a strong indication that most people are downloading the 1.4.1 Raspbian image (no 1.4.1 of the Pine64). Version 1.4 which contains both Raspbian and Pine64 images is downloaded just a few times a day.

Maybe it is worth to start a discussion in the community forum, since we do not see what is actually in use.

My personal opinion: I would propose to put the efforts towards the Raspbian image and get it running on then new Rpi4 with 4GB RAM. This would keep the image creation scripts a bit cleaner and better to maintain.
Best,

EliasGabrielsson:
I use the pine64 for my everyday server so it's in my interest to build an image.
A though I have had if it would be possible to use [armbian](https://docs.armbian.com/) as a base image for all non RPi boards and though support a lot of boards.

PR #590, opened 2019-07-03
only add openhabian user to backup group if it exists

### LINKED ISSUE #533: Setting up amanda without openHABian user fails
The line @ [backup.sh](https://github.com/openhab/openhabian/blob/master/functions/backup.sh#L166) will fail if the openhabian user does not exist. It is likely trying to set ownership to the default user's group for easy access.

We should check if the openhabian user exists first before executing such a command, and make the inputs to the usermod function variables.

A secondary question: Should the openhabian group exist for manual installs too?

See: https://community.openhab.org/t/fatal-error-when-installing-amanda-backup-via-openhabian-config/68229/4

mstormi:
fix #533 


EliasGabrielsson:
Should not this group be created if it not exist?
Will amanda work without it?

mstormi:
Amanda does not need it to run (as long as the "backup" user does exist) so it does not need to be created. This is just to answer Benjy's #533.

mstormi:
@BClark09  Benjy, are you ok with this fix, can we merge it ?

BClark09:
Yes, thanks @mstormi!

Issue #289, opened 2017-12-19
raspi-config should not be included (Stretch)


ThomDietrich:
With the latest stretch based image it seems like (1) raspi-config is back and (2) the alias once defined to inform the user to use openhabian-config doesn't work anymore.

Investigate and repair.

mstormi:
raspi-config long gone, also in buster.

PR #602, opened 2019-07-22
Removed raspi-config as a tool

EliasGabrielsson:
Fixes: #289

### DIFF
--- a/includes/bash_profile
+++ b/includes/bash_profile
@@ -23,7 +23,7 @@ elif [ -f "/opt/openHABian-install-inprogress" ]; then
   return 0
 fi
 
-if [ ! -f "/usr/bin/raspi-config" ]; then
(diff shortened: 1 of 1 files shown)


ThomDietrich:
👍 
Happy you found my old ticket and solved it. The change confuses me though. Could you explain why this change is the solution?

ThomDietrich:
If the command is not available: Show info text.
If the command is available: User explicitly installed it and we shouldn't mess with this!? Correct?

EliasGabrielsson:
Okey, then I misunderstood the feature. I thought we like to force user to not use the `raspi-config` as it would interfere with `openhabian-config`.

I do not know why it wasn't working before... Strange behavior

EliasGabrielsson:
What about just get rid of this removal?

Issue #534, opened 2019-02-21
Sort out correct user:group combinations


BClark09:
Related somewhat to #533

openHABian sets the permissions of the openHAB directories to `openhab:openhabian` which overwrites the package/service's default of `openhab:openhab`.

It seems weird to me to assign folder permissions created by a service to a user group so. I believe it would be best to keep the permissions rights as it is, and if necessary add the `openhabian` user (as well as any service users that need write permissions e.g. for backup) to the `openhab` group.

We can then call `openhab-cli reset-ownership` to deal with openHAB's own folder structure.

WDYT?

EliasGabrielsson:
I personally don't have any experience to add to the question, but if this is the de-facto standard for Linux I would prefer this design. @ThomDietrich do you have any thoughts to add on to this design decision?

mstormi:
Agree keeping the openhab group default is Linux standard and I'd prefer that, too.

ThomDietrich:
I agree. The openhabian group is not needed and as openhabian is member of the openhabian group, all stays the same for openhabian in case of `openhab:openhab` permissions

mstormi:
> I agree. The openhabian group is not needed and as openhabian is member of the openhabian group, all stays the same for openhabian in case of `openhab:openhab` permissions

I assume you mean the user openhabian to be member of the group **openhab** (without 'ian') ? Else with this change openhabian will not be usable any more for operating OH ?

ThomDietrich:
No you are correct.

PR #606, opened 2019-07-26
Give ownership of Openhab folders to "openhab"

EliasGabrielsson:
Fixes: #534

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ node_modules/
 *.img.xz
 *.img.bz2
 *.log
+.DS_Store
--- a/functions/system.bash
+++ b/functions/system.bash
@@ -177,10 +177,9 @@ permissions_corrections() {
 
   #
   openhab_folders=(/etc/openhab2 /var/lib/openhab2 /var/log/openhab2 /usr/share/openhab2)
-  cond_redirect chown openhab:$username /srv /srv/README.txt
+  cond_redirect chown openhab:openhab /srv /srv/README.txt /opt
   cond_redirect chmod ugo+w /srv /srv/README.txt
-  cond_redirect chown -R openhab:openhab /usr/share/openhab2 ${openhab_folders[@]}
-  cond_redirect chown -R openhab:$username /opt
+  cond_redirect chown -R openhab:openhab ${openhab_folders[@]}
   cond_redirect chmod -R ug+wX /opt ${openhab_folders[@]}
   cond_redirect chown -R $username:$username /home/$username
   #
--- a/functions/system.bash
+++ b/functions/system.bash
@@ -176,11 +176,11 @@ permissions_corrections() {
   cond_redirect adduser "$username" openhab
 
   #
-  openhab_folders=(/etc/openhab2 /var/lib/openhab2 /var/log/openhab2 /usr/share/openhab2/addons)
+  openhab_folders=(/etc/openhab2 /var/lib/openhab2 /var/log/openhab2 /usr/share/openhab2)
   cond_redirect chown openhab:$username /srv /srv/README.txt
   cond_redirect chmod ugo+w /srv /srv/README.txt
-  cond_redirect chown -R openhab:openhab /usr/share/openhab2
-  cond_redirect chown -R openhab:$username /opt ${openhab_folders[@]}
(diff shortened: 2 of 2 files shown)


EliasGabrielsson:
Help me think this through. As the user `openhabian` have access to the group `openhab` this ownership change should not make any issues right?

EliasGabrielsson:
I also wonder if we shall only make the folders and the config files readable by `openhab` and `openhabian` user?

These files can contain quite sensitive API keys etc. 
Does any other service then openhab (+ the openhabian user) needs to have access to the folders? 

Fronttrail what permission is that service using?

mstormi:
I'm a little confused. Changing permissions is a sensitive thing that easily can go wrong.
A (first, incomplete) number of questions to come to mind:
* Do we always have a openhabian user ? What if someone's installing but not using the image ?
* we mustn't expect users to use the openhab user for interactive works. It used to be "openhabian" or something else but not "openhab". 
* Your patch shows .... `chown -R openhab:$username` ... why, use the USER name for the GROUP ?

EliasGabrielsson:
> Do we always have a openhabian user ? What if someone's installing but not using the image ?
Yes, we always have a user as we ask for that in non-openhabian installations.

> we mustn't expect users to use the openhab user for interactive works. It used to be "openhabian" or something else but not "openhab".
Sorry I don't get what you mean. The patch changes ownership of openhab runtime folder etc. to the service account. Could you elaborate a bit more? 

> Your patch shows .... chown -R openhab:$username ... why, use the USER name for the GROUP ?
I think there is an "openhabian" group as well but this is clearly confusing. This would also be broken in the case of installing on a non-openhabian installation. I will look into it.

mstormi:
> The patch changes ownership of openhab runtime folder etc. to the service account. Could you elaborate a bit more?

openhab is the account OH runs as, but a user does not necessarily use it to configure it. That's "openhabian", "pi" or even something else.
We need files to be writeable for the OH group and "openhab" and (all) the service account(s) need to be members of that group as well.

.gitignore
```diff
@@ -5,3 +5,4 @@ node_modules/
 *.img.xz
 *.img.bz2
 *.log
+.DS_Store
```
ThomDietrich: I stopped to add these files to every repository individually and added them to my systems global gitignore instead. Still not against this, but maybe at all while at it? https://github.com/ebeshero/DHClass-Hub/blob/master/.gitignore

functions/system.bash
```diff
@@ -176,11 +176,11 @@ permissions_corrections() {
...
 
   #
-  openhab_folders=(/etc/openhab2 /var/lib/openhab2 /var/log/openhab2 /usr/share/openhab2/addons)
+  openhab_folders=(/etc/openhab2 /var/lib/openhab2 /var/log/openhab2 /usr/share/openhab2)
   cond_redirect chown openhab:$username /srv /srv/README.txt
   cond_redirect chmod ugo+w /srv /srv/README.txt
-  cond_redirect chown -R openhab:openhab /usr/share/openhab2
-  cond_redirect chown -R openhab:$username /opt ${openhab_folders[@]}
+  cond_redirect chown -R openhab:openhab /usr/share/openhab2 ${openhab_folders[@]}
+  cond_redirect chown -R openhab:$username /opt
```
ThomDietrich: I agree with @BClark09. Just merge both lines for `openhab:openhab` and get rid of all other `openhab:$username`.

EliasGabrielsson:
Take a look at this please. Are we settled?

mstormi:
There's a couple of places where files in the provisioning user's homedir are set group ownership $username like
functions/openhab.bash:114:  chown -R "$username:$username" /home/$username/.vim
functions/system.bash:213:  chown -R "$username:$username" /home/"$username"/.ssh
functions/system.bash:193:  cond_redirect chown -R "$username:$username /home/$username"

Would we want to keep that ? or change to group openhab ?

Working on PR #613 there's also a couple of places shellcheck marked where the $username variable eventually isn't set and I changed those to have a default now.
Wasn't always sure what to insert, e.g. the user and group to use when installing node-red ?
Maybe worth double checking those places for this issue here.

EliasGabrielsson:
Hmm, the above cases regards the homefolder so we shouldn't mess with a service account over there.

PR #601, opened 2019-07-22
Updated FireModD installation

EliasGabrielsson:
* Removed old hotfix for spikes on github repo
* Fixes: #425, #568

Signed-off-by: Elias Gabrielsson <elias@gabson.se>

PR #609, opened 2019-07-30
a number of fixes to the build system

mstormi:
This is a number of fixes that do NOT change functionality.
Most of these fix running openhabian-setup and BATS tests in Docker, some of the others are cosmetic.


build-image/first-boot.bash
```diff
@@ -156,18 +156,22 @@ ln -sfn /opt/openhabian/openhabian-setup.sh /usr/local/bin/openhabian-config
 echo "$(timestamp) [openHABian] Executing 'openhabian-setup.sh unattended'... "
 if (/bin/bash /opt/openhabian/openhabian-setup.sh unattended); then
 #if (/bin/bash /opt/openhabian/openhabian-setup.sh unattended_debug); then
-  systemctl start openhab2.service
+  service openhab2 start
```
EliasGabrielsson: What is the reason to change to `service` in place of `systemctl` ?
If there are reasons for it we should change the whole codebase and track it in an individual issue.

build-image/first-boot.bash
```diff
@@ -156,18 +156,22 @@ ln -sfn /opt/openhabian/openhabian-setup.sh /usr/local/bin/openhabian-config
...
 else
   fail_inprogress
 fi
 echo "$(timestamp) [openHABian] Execution of 'openhabian-setup.sh unattended' completed."
 
-echo -n "$(timestamp) [openHABian] Waiting for openHAB to become ready... "
-until wget -S --spider http://localhost:8080 2>&1 | grep -q 'HTTP/1.1 200 OK'; do
+echo -n "$(timestamp) [openHABian] Waiting for openHAB to become ready..."
+i=0
+until wget -S --spider -t 3 --waitretry=4 http://localhost:8080 2>&1 | grep -q 'HTTP/1.1 200 OK' || [ $i -ge 10 ]; do
```
EliasGabrielsson: I like this part 👍 We have plenty of places thought out the code which could use this pattern.
Could you make this to a separate function in the helper section?

Maybe something like this?
```
untilRdy(String checkCommand, Int tries) {
...
} 
```

build-image/webif.bash
```diff
@@ -49,7 +49,7 @@ fi
 
 if [ $1 = "cleanup" ]; then
   webifrunning=$(ps -ef | pgrep python3)
-  if [ -z $webifrunning ]; then
+  if ! [ -z $webifrunning ]; then
```
EliasGabrielsson: Why kill it when it not exist?

functions/influxdb+grafana.bats
```diff
@@ -3,16 +3,20 @@
 load influxdb+grafana
 load helpers
 
-@test "destructive-install_influxDB" {
+@test "destructive-influxDB_install" {
```
EliasGabrielsson: Will this improve sorting ?

functions/influxdb+grafana.bats
```diff
@@ -3,16 +3,20 @@
...
   run influxdb_install "Password1234"
   [ "$status" -eq 0 ]
-  run systemctl is-active --quiet influxdb.service
+  echo -e "# \e[32mInfluxDB installation successful." >&3
```
EliasGabrielsson: I would say this is implicit understood if the test passes. I do not want to flood the CI log, shellcheck does it well already 😆 
What do you have in mind?

functions/nodejs-apps.bash
```diff
@@ -35,13 +35,16 @@ frontail_setup() {
...
+  
   cp $BASEDIR/includes/frontail-preset.json $frontail_base/preset/openhab.json
   cp $BASEDIR/includes/frontail-theme.css $frontail_base/web/assets/styles/openhab.css
   cp $BASEDIR/includes/frontail.service /etc/systemd/system/frontail.service
   chmod 664 /etc/systemd/system/frontail.service
   cond_redirect systemctl daemon-reload
   cond_redirect systemctl enable frontail.service
-  cond_redirect systemctl restart frontail.service
+  cond_redirect service frontail restart
+  service frontail status >&/dev/null
```
EliasGabrielsson: I don't get this line. Do you send the status to null?

functions/system.bash
```diff
@@ -109,7 +109,9 @@ hostname_change() {
   hostnamectl set-hostname "$new_hostname" &>/dev/null
   hostname "$new_hostname" &>/dev/null
   echo "$new_hostname" > /etc/hostname
-  sed -i "s/127.0.1.1.*/127.0.1.1 $new_hostname/" /etc/hosts
+  local TMP="$(mktemp /tmp/.XXXXXXXXXX)"
+  sed "s/127.0.1.1.*/127.0.1.1 $new_hostname/g" /etc/hosts >${TMP} && cp $TMP /etc/hosts
+  rm -f $TMP
```
EliasGabrielsson: What does this solve?

build-image/webif.bash
mstormi: -z tests for "empty"

build-image/first-boot.bash
mstormi: Tricky. Actually systemctl does not (always) work in Docker but service does.
Found a different solution meanwhile though so will remove this change

build-image/first-boot.bash
mstormi: Will do midterm but for now I mustn' lose the overview so won't.
Still WIP with this inside Docker , the dashboard won't come up.

functions/influxdb+grafana.bats
mstormi: no but beautify the output

functions/influxdb+grafana.bats
mstormi: Actually when working on tests inside docker it's useful to have intermediate output.
This is just 2 lines. Shellcheck is yet another story of course

functions/system.bash
mstormi: a weird error in Docker testrun

functions/nodejs-apps.bash
mstormi: Yes but I reverted that to the original, reason same as in the first comment

functions/nodejs-apps.bash
```diff
@@ -35,13 +35,15 @@ frontail_setup() {
   cond_redirect npm update -g frontail
   #
   frontail_base="/usr/lib/node_modules/frontail"
+  mkdir -p ${frontail_base}/preset ${frontail_base}/web/assets/styles
```
EliasGabrielsson: Markus do you think that this will have any implications on this PR: https://github.com/openhab/openhabian/pull/601 ?

mstormi: sorry maybe because it's late but I don't see any linkage there ?

EliasGabrielsson:
Okey, this is ready to merge. Now I just want a squashed commit.
I recommend to set an git editor you are comfortable with before doing this. For example using VS Code `git config --global core.editor "code --wait"`

Then squash by doing:
1. Enter your branch: buildsystem
2. Squash commit history:
`git rebase -i -S HEAD~23`
Pick "s" on each commits expect the top most one, save file.
3. Update PR by `git push --force` which will overwrite your content in github.

mstormi:
> Okey, this is ready to merge. Now I just want a squashed commit.

Ok. Thought you would want to keep commits so we could selectively revert them in case of need.
But here you are.

openhabian-setup.sh
```diff
@@ -80,7 +80,6 @@ if [[ -n "$UNATTENDED" ]]; then
   misc_system_settings
   samba_setup
   clean_config_userpw
-  nodejs_setup
```
EliasGabrielsson: It is not possible to remove NodeJS from base installation, used by some packages.
Why was this removed?

functions/nodejs-apps.bash
EliasGabrielsson: Okey, nice

openhabian-setup.sh
mstormi: it's called (again) as 1st step in frontail_setup() anyway

EliasGabrielsson: Hmm you are correct. Let's call the node installation from there.

EliasGabrielsson:
@mstormi It is really hard to keep a clean commit history when you merge the upstream master branch into your local development branches. From now on you will need to _rebase_ your work on top of master.

When you sync your local branches to upstream openhabian do NOT use `git pull` or `git merge upstream/master` or similar. Use `git fetch upstream` and `git rebase upstream/master`.

mstormi:
mmh, ok. Still occasionally fighting with git but I think I'm getting there.

EliasGabrielsson:
Just ask if things get all **** up. GIT is hard because there are a lot of ways to achieve similar results. A good rule of thumb when working in a github context is ONLY use `git pull` on your local master branches.

PR #610, opened 2019-07-30
Update openHAB key on system upgrade

### LINKED ISSUE #607: Update keys with Upgrade option (02)
See https://community.openhab.org/t/apt-repo-public-key-expiration-update/78638.

Worth adding?

BClark09:
Closes #607 

### DIFF
--- a/functions/system.bash
+++ b/functions/system.bash
@@ -11,6 +11,15 @@ whiptail_check() {
 
 system_upgrade() {
   echo -n "$(timestamp) [openHABian] Updating repositories and upgrading installed packages... "
+  # TODO: Remove updating the key on each update after some time (in 2020)
+  cond_redirect wget -O openhab-key.asc 'https://bintray.com/user/downloadSubjectPublicKey?username=openhab'
+  cond_redirect apt-key add openhab-key.asc
+  if [ $? -ne 0 ]; then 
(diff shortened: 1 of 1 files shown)


functions/system.bash
```diff
@@ -11,6 +11,15 @@ whiptail_check() {
...
 system_upgrade() {
   echo -n "$(timestamp) [openHABian] Updating repositories and upgrading installed packages... "
+  # TODO: Remove updating the key on each update after some time (in 2020)
+  cond_redirect wget -O openhab-key.asc 'https://bintray.com/user/downloadSubjectPublicKey?username=openhab'
+  cond_redirect apt-key add openhab-key.asc
+  if [ $? -ne 0 ]; then 
+    echo "FAILED (updating OH key)"
+    echo -n "attempting to continue..."
+  else
+    rm -f openhab-key.asc
```
ThomDietrich: Unconditional?

BClark09: Do you mean we shouldn't need the else? I can create another PR if so?

ThomDietrich: Yes feels like this can potentially end in a file written to the current folder. We should in any case always try to use `/tmp` files 🤔
Don't bother, it's a temporary fix and should be okay.

PR #612, opened 2019-07-31
make python webserver start

mstormi:
seems that never worked properly that early in the process because 1) python was not installed and 2) some install/grep stuff used  "python" and some "python3"

### DIFF
--- a/build-image/first-boot.bash
+++ b/build-image/first-boot.bash
@@ -27,7 +27,9 @@ source /etc/openhabian.conf
 echo "OK"
 
 echo -n "$(timestamp) [openHABian] Starting webserver with installation log... "
-if hash python 2>/dev/null; then
+cond_redirect apt-get update
+cond_redirect apt-get install python3 python3-pip
+if hash python3 2>/dev/null; then
   bash /boot/webif.bash start
   sleep 5
   webifisrunning=$(ps -ef | pgrep python3)
@@ -37,7 +39,7 @@ if hash python 2>/dev/null; then
     echo "OK"
   fi
 else
-  echo "Python not found, SKIPPED"    
+  echo "Python 3 not found, SKIPPED"    
 fi
 
 userdef="openhabian"
--- a/build-image/first-boot.bash
+++ b/build-image/first-boot.bash
@@ -28,7 +28,7 @@ echo "OK"
 
 echo -n "$(timestamp) [openHABian] Starting webserver with installation log... "
 cond_redirect apt-get update
-cond_redirect apt-get install python3 python3-pip
+cond_redirect apt-get -y install python3 python3-pip
 if hash python3 2>/dev/null; then
   bash /boot/webif.bash start
   sleep 5


ThomDietrich:
but it worked back when we merged it.
@mues-lee wdyt?

EliasGabrielsson:
Isn't python a part `raspian-lite` image?
This feels like an path issue.

mues-lee:
yes python, as well as python3, was part of the 'raspian lite' image, it was the main reason to choose the python approach at all
I can't test it at the moment, if python is in the the lite-image, but i would be very surprised if they had removed it

In case the python packages are still part of the image, i think the changes are unnecessary

mstormi:
I'm working on getting the openhabian-setup unattended test to complete in Docker.
I'm not using stock Raspbian but the balenallib docker image which apparently does not have python3 available.
As it should work on both bases it makes sense to install it (does not harm if it's already installed).

The final openhabian-setup step, i.e. access to http://localhost:8080/ still doesn't work because no server is listening.
Debugging from there backwards I came to see that the Python webserver wasn't started
Hence this PR.
Now with that it does start on port 80 but there's still nothing to listen on 8080.

Could anyone quickly explain please what is supposed to run on 8080 and where it is started from ?

ThomDietrich:
Sounds reasonable. Let's do this. However it's probably worth to turn the logic around:

```
if python not available
   install python
fi
if python not available
  echo error message
fi
set up webserver
```

Wdyt?

EliasGabrielsson:
I am pushing to get a 1.5 release out this weekend and this feature needs to be tested on actual hardware. @BClark09 @holgerfriedrich do you mind to actually test this on a RPi4?

mstormi:
>  this feature needs to be tested on actual hardware

*this* one ? I don't think so. It makes sense to *always* apt-get install python3/python3-pip packages regardless of the HW and OS the install is executed on. Either they're already there (like on Raspbian) then these commands effectively do nothing (well, eventually this will update the packages to their latest version) or they're missing (like in my OS base image) and this will install them.

EliasGabrielsson:
Mange to get the webserver to work on a RPi3 using a freshly built Buster image. Closing this now.

mstormi:
The webservice still does not work if python3 ain't installed such as on older or non Raspbians OS images so I think we should rather merge my patch than to say we don't need it because it works on buster Raspbian (....only....).

EliasGabrielsson:
We do not support old images nor other boards then RPi, so no I will not merge this patch.

mstormi:
what about x86?

EliasGabrielsson:
Yea, but this feature only regards the automatic installation which only is targeted at the Raspbian OS.

ThomDietrich:
@EliasGabrielsson whenever possible, let's try to discuss things in a constructive way before directly closing a PR.

I did not dislike the improvement of this feature to additionally install python when not existent, HOWEVER I agree that this not needed in the current scope and we have the friendly "Python not found, SKIPPED" message for those developers that might run into this again. The web interface is only relevant in the RPi image and this one will contain python already. Did we miss anything @mstormi ?

All the best, th

mstormi:
Yes web if is relevant for RPi installs from image only but leaving dangling code for other OS doesn't feel right either.
Having python thus installing still makes sense even on non-RPIs as it's also used (and installed!) e.g. in timezone calculation so I see no downside in installing it earlier (which btw. only happens IF it wasn't contained in the image already).
And I still believe the concept is broken because we install (or expect to have in place) "python" but the webserver explicitly looks for and runs "python3".
Plus eventually there's a more up to date version available so we should "install"=update even if it was contained in the image.

EliasGabrielsson:
Sorry if I sounded rude and nonconstructive in my earlier comments. I do however stand my point,
I think we should try to keep the first boot file as lean as possible. 

Doing all heavy lifting in the code we run after cloning the openhabian repo enable us to better patch broken code and iterate faster. I would say a python installation should go into that category.

PR #614, opened 2019-08-02
Allow building RPI image without root

holgerfriedrich:
Allow building RPI image without root. guestmount from libguestfs-tools
is used instead of kpartx/loopback/mount which requires root.


EliasGabrielsson:
Hey Holger! Good patch 👍

PR #637, opened 2019-08-04
Ngnix certbot provided by Buster repo; closes #636

### LINKED ISSUE #636: Ngnix install fails on Buster if domain is specified
Ngnix install fails in case the user specifies a domain name.

The error leaves the apt system broken
`E: Malformed entry 2 in list file /etc/apt/sources.list.d/backports.list (Component)`
`E: The list of sources could not be read.`
until you drop the corrupted line from `backports.list`

Root cause is in `packages.bash` line 571, where certrepo is undefined for Buster.

holgerfriedrich:

### DIFF
--- a/functions/packages.bash
+++ b/functions/packages.bash
@@ -568,7 +568,9 @@ nginx_setup() {
             certbotrepo="stretch-backports"
           fi
           certbotoption="-t"
-          echo -e "# This file was added by openHABian to install certbot\ndeb http://ftp.debian.org/debian ${certbotrepo} main" > /etc/apt/sources.list.d/backports.list
+          if [ ! -z "$certbotrepo" ]; then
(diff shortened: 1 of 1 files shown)


EliasGabrielsson:
Good find.
But your fix will break installation for Stretch.
Why not just update the package file with `buster` ?

holgerfriedrich:
Both Stretch and old Jessie define certbotrepo. I think it will not break something...
Or did I make a mistake?

EliasGabrielsson:
I think we need this: ~

```
elif is_buster; then
            certbotrepo="buster"
fi
```

wdyt?

holgerfriedrich:
No, die Buster certbot ist not in backports anymore but included in the standard repo.

BClark09:
@EliasGabrielsson the travis build failed but I don't think it's related to this PR. How can we get it to rerun? Happy to merge anyway.

EliasGabrielsson:
I think we need to look over the java installation routine. Usually that which fails.

PR #640, opened 2019-08-05
fix MTA installation

holgerfriedrich:
remove additional line breaks


holgerfriedrich:
No clue why this was never flagged by shellcheck. I could not complete the run on packages.bash w/o fixing it....

Issue #642, opened 2019-08-06
Script fails to install zulu-embedded (due to new url maybe) 


libo:
This line https://github.com/openhab/openhabian/blob/master/functions/java-jre.bash#L38
tries to fetch the `tar.gz` from azul.com website by parsing the HTML code.

It looks like Azul changed something lately and the url the script tries to load:
`https://www.azul.com/downloads/zulu-embedded` actually redirects to `https://www.azul.com/products/zulu-embedded/`

(we don't use the `-L` option in curl so it doesn't follow redirections)

I have tried to update the URL in the script but with no success... looks like the page changed a lot.

teolairlasg:
The links apparently are now in:
https://www.azul.com/downloads/zulu-community/

Download links (I think): (**https**)
https://cdn.azul.com/zulu-embedded/bin/zulu8.40.0.178-ca-jdk1.8.0_222-linux_aarch64.tar.gz
https://cdn.azul.com/zulu-embedded/bin/zulu8.40.0.178-ca-jdk1.8.0_222-linux_aarch32hf.tar.gz

I tried to change the regular expression adding the https but still not working. 
Could it be because of AJAX?

libo:
Yes I inspected the code and can confirm that the list is populated asynchronously via a POST request to this end-point `https://www.azul.com/wp-admin/admin-ajax.php`

<img width="827" alt="Screenshot 2019-08-06 at 10 11 46" src="https://user-images.githubusercontent.com/37512/62522846-d270e380-b832-11e9-9218-951f2918a0b1.png">

@EliasGabrielsson I am wondering if it would make sense at this point to hard code the URLs to the JDK distros instead of trying to parse HTML with a regexp. That would also allow us to point to a specific version of the dist, and that might be not that a bad idea after all.

EliasGabrielsson:
Hi!
Thanks for your research of this issue. I will try to make a patch for it this evening.
If any of you could make an PR I will merge it directly. 

// EG

mstormi:
It's ok as a quick fix but generally speaking hardcoded URLs are not a good solution (as can be observed here).
As we have this in several locations I think it is worth a general review/refactoring of these.


On this here with Java, Azul seems to provide repos (again)
https://docs.azul.com/zulu/zuludocs/Content/ZuluReleaseNotes/Troubleshooting.htm#Azul_Repo_Tips

Maybe switch back to that install method ?

EliasGabrielsson:
[...]

PR #643, opened 2019-08-06
Hardcode JDK `downloadPath`

libo:
Azul changed their website and now it populates the list of JDKs
dynamically via Ajax call.

This PR hardcodes the 2 URLs in order to improve realiability of the
script and to point statically to a defined version so to avoid potential
compatibility issues. This also means that we will need to keep track
of new version and update the URLs.
We might consider extracting these URLs in two extranal constant defined
elsewhere.

cc: @teolairlasg @EliasGabrielsson

Closes https://github.com/openhab/openhabian/issues/642

### DIFF
--- a/functions/java-jre.bash
+++ b/functions/java-jre.bash
@@ -32,10 +32,10 @@ java_zulu(){
     file="/var/tmp/.zulu.$$"
 
     if is_aarch64; then
-      downloadPath=$(curl -s https://www.azul.com/downloads/zulu-embedded | grep -Eo "http://[a-zA-Z0-9./?=_-]*zulu8[a-zA-Z0-9./?=_-]*aarch64.tar.gz")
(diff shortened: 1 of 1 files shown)


EliasGabrielsson:
@libo Thanks so much for solving this and welcome as on contributor to the openHABian project. Feel free to give more opinions, ideas and RP for this project.

libo:
Thanks @EliasGabrielsson :-)

It looks like the latest master is back to green, I think we can release a new 1.5.1 whenever you have time...

EliasGabrielsson:
No need actually. We pull the latest version of the script directly from github during installation so you have probably already saved a lot of installations. 🥇

BClark09:
Unfortunately, openhab fails to launch at firstboot because the update-alternatives part of the script fails silently. See #645

PR #645, opened 2019-08-06
Fix download url typo in JDK setup

BClark09:
The script currently fails because the sed command fails to remove the different url path. Causing a silent error in `update-alternatives` and presumably in `ld`.


libo:
@EliasGabrielsson this is needed to fix the script actually... :-(

Issue #651, opened 2019-08-06
Rewrite KNXD to use Debian package during installation


holgerfriedrich:
Installing optional component knxd fails on RPI3B+ with Buster.
This is probably an issue with upstream repos.

<code>
/bin/bash ../../libtool  --tag=CXX   --mode=link g++  -g -O2 -std=c++0x -Wno-subobject-linkage   -o findknxusb findknxusb.o -lusb-1.0 -lev -lfmt 
libtool: link: g++ -g -O2 -std=c++0x -Wno-subobject-linkage -o findknxusb findknxusb.o  -lusb-1.0 -lev -lfmt
g++ -DHAVE_CONFIG_H -I. -I../..  -I../../src/libserver -I../../src/common -I/usr/include/libusb-1.0 -I/root/knxdbuild/fmt  -g -O2 -std=c++0x -Wno-subobject-linkage -MT usb.o -MD -MP -MF .deps/usb.Tpo -c -o usb.o usb.cpp
In file included from /usr/lib/gcc/arm-linux-gnueabihf/8/include-fixed/limits.h:194,
                 from /usr/lib/gcc/arm-linux-gnueabihf/8/include-fixed/syslimits.h:7,
                 from /usr/lib/gcc/arm-linux-gnueabihf/8/include-fixed/limits.h:34,
                 from /usr/include/libusb-1.0/libusb.h:62,
                 from usb.h:24,
                 from usb.cpp:23:
/root/knxdbuild/fmt/fmt/format.h: In member function ‘void fmt::internal::ArgFormatterBase<Impl, Char>::visit_char(int)’:
/root/knxdbuild/fmt/fmt/format.h:1825:20: error: expected unqualified-id before numeric constant
     const unsigned CHAR_WIDTH = 1;
                    ^~~~~~~~~~
[...few more warnings...]
make[3]: *** [Makefile:501: usb.o] Error 1
make[3]: Leaving directory '/root/knxdbuild/knxd/src/usb'
make[2]: *** [Makefile:394: all-recursive] Error 1
make[2]: Leaving directory '/root/knxdbuild/knxd/src'
make[1]: *** [Makefile:442: all-recursive] Error 1
make[1]: Leaving directory '/root/knxdbuild/knxd'
make: *** [Makefile:378: all] Error 2
Created symlink /etc/systemd/system/multi-user.target.wants/knxd.service → /lib/systemd/system/knxd.service.
[...]

Issue #321, opened 2018-01-22
knxd versions newer than 0.14.19-2 do not start at boot


guiott:
The knx daemon, installed with the openhabian-config script should start automatically at boot using systemd
The default parameters suggested at installation (KNXD_OPTIONS="–eibaddr=1.1.128 --client-addrs=1.1.129:8 -d -D -T -R -S -i --listen-local=/tmp/knx -b tpuarts:/dev/ttyAMA0") was good until version 0.14.19-2 of knxd-
As explained on knxd github the “-d” parameter is needed to demonize the program when launched manually from cli, so it’s redundant when launched by systemctl, but was “tollerated” until that version.
Using that parameter, starting from version 0.14.22, causes the program to quit immediately with exit code [0]Success, even if it is still good using the cli.
It must be removed manually editing /etc/default/knxd.
The correction must be repeated if you reinstall knxd (now at version 0.14.23) because openhabian-config installation repropose the same configuration file.

ThomDietrich:
Hello @guiott, thanks for the info! Would you like to contribute a PR?  
What do you think about replacing the used installation script as discussed in #85?

guiott:
Hi Thomas,

I’m not good on writing scripts but I’m a good tester, I install often OpenHAB and knxd on the board we developed (https://www.acmesystems.it/CM3-HOME <https://www.acmesystems.it/CM3-HOME>). The openhabian installation script is very useful because otherwise the installation is very complicated.
I’m going to test the suggested script ASAP and I’ll give my personal opinion.

ciao
Guido

ThomDietrich:
That would already be a big help! Thanks!

I know your CM3-Home and it's a great product! Happy to see that you are using openHABian with it! I'd be happy to discuss further integration of available hardware and setup steps in openHABian! At some point we could also mention the CM3-Home as an openHAB(ian)-ready product in our documentation

guiott:
GREAT! 
I’d be happy to. 
I’m working a lot on OpenHabian, trying to integrate all the devices I always desired to use in my home. Which way can be better than testing your own product in your own, real environment?

All the devices/configurations/services I’ve used are documented on this article series:
http://www.tanzolab.it/openhab_gettingstarted

I was still waiting to share the link because the articles are not yet be all translated in English, I’m working on that, but I’m quicker on adding functions than on translating :-) 
[...]

Issue #85, opened 2017-01-24
KNX installation not working (anymore)


StephanSchuster:
I just wrote the following comment to the owner of the underlying install script (currently awaiting moderation):

> Hey,
> 
> your script no longer seems to work since knxd moved to 0.12. Could you please check this? Would be great.
> 
> I came across the issue when using openhabian’s config option “Install KNX” which internally uses your script. This fails due to the above mentioned reason. At least that’s my first guess.
> 
> After adding ‘git checkout tags/v0.10.18 -b install’ after ‘git clone … cd knxd’ in your script, the installation succeeds. However, I guess it’s better to migrate your script to knxd 0.12 (no pthsem dependency).
> 
> Are there any chances that you “fix” this soon? I guess a lot of people are stumbling over this when they want to try out the new openhab 2.0 release via openhabian (+KNX)
> 
> By the way: Thank you for your great work!

ThomDietrich:
Interesting indeed. Please keep me posted. If you do not receive a response in the next few days, please tell me and I'll include a modified version directly in openHABian.

tischla:
Hi,
Script stops with:

checking for LIBUSB… yes
EMI because of usb
checking for ev_run in -lev… no
configure: error: libev not found
make: *** No rule to make target ‘clean’. Stop.


thx
Roland

cribskip:
There are 2 packages missing: libev-dev libev4
Afterwards, after triggering from within openhabian-config, it builds just fine.

KR,
Sascha

ThomDietrich:
@stephanschuster was the script updated or is the solution by @cribskip all that's needed? sure sounds like a clean solution.
@cribskip so adding these to packages to the installation routine is all that's needed?

stephanschuster:
Well, I'm not at home at the moment. So I cannot verify this. But I remember that I definitely installed libev-dev and libev4 and retriggered the install script. Not sure anymore if the build was really successful then. I've tried various stuff meanwhile. In the end I never managed to see KNX packages with knxtool. But that might have other reasons. My setup: RPI3, openHABianpi, knxd, TUL stick with "transparent" firmware.

Regarding the script: I think even if it compiles with libev, it needs to be adjusted a bit. According to the knxd homepage version 0.12 has no more deps to pthsem which is also installed by the script.

stephanschuster:
[...]

PR #652, opened 2019-08-06
hotfix for compilation errors in knxd on Buster #651

holgerfriedrich:
Decided to lift libfmt to from 3.0.0 to 5.0.0. Compiles and installs.
Not tested on my system, knxd not in use.

### DIFF
--- a/functions/packages.bash
+++ b/functions/packages.bash
@@ -341,6 +341,13 @@ Integration into openHAB 2 is described here: https://github.com/openhab/openhab
 
   #TODO serve file from the repository
   cond_redirect wget -O /tmp/install_knxd_systemd.sh http://michlstechblog.info/blog/download/electronic/install_knxd_systemd.sh || FAILED=1
+
+  #hotfix for #651: switch to newer libfmt
+  cond_redirect sed -i 's#git checkout tags/3.0.0#git checkout tags/5.0.0#' /tmp/install_knxd_systemd.sh || FAILED=1
+  cond_redirect sed -i 's#cmake -DCMAKE_INSTALL_PREFIX=$INSTALL_PREFIX fmt/#cmake -DCMAKE_INSTALL_PREFIX=$INSTALL_PREFIX .#' /tmp/install_knxd_systemd.sh || FAILED=1
+  #/hotfix
+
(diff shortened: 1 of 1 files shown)


mstormi:
why would you want to compile on your box ? Can't expect that to work IMHO.
knxd can be installed from the menu so there's not even a need for that.

holgerfriedrich:
@mstormi Compilation is actually triggered by installation via menu.
Not sure if the way of providing it is the best. There are two other open issues on knxd, #85 and #321. I think #321 discusses a different option of installing it.

My focus right now is just to make as many of the menu options work on the new 1.5 image / Buster.
I cannot test it anyways, not using knxd.

holgerfriedrich:
@EliasGabrielsson please merge the hotfix.

Good to have it available for Buster as a package. This will make it easier to keep it up to date and the system clean. Our overall approach should be to use as much packages as possible.

Though, unless we do not remove support for Ubuntu, Jessie, and Stretch it will complicate things.

EliasGabrielsson:
There is a planned clean up for at least Ubuntu and Jessie coming up this fall I guess.
Let's look into it when it happen. Just of curiosity do you use KNX system yourself?

holgerfriedrich:
> There is a planned clean up for at least Ubuntu and Jessie coming up this fall I guess.
> Let's look into it when it happen.
Agreed. I can support work for Debian/Raspbian based distros.

> Just of curiosity do you use KNX system yourself?
Yes, but not knxd. OH knx binding via IP interface.

mstormi:
does it work?
Please shellsheck all work from now on

 functions/packages.bash line 325:

  cond_redirect sed -i 's#cmake -DCMAKE_INSTALL_PREFIX=$INSTALL_PREFIX fmt/#cmake -DCMAKE_INSTALL_PREFIX=$INSTALL_PREFIX .#' /tmp/install_knxd_systemd.sh || FAILED=1

                       ^-- SC2016: Expressions don't expand in single quotes, use double quotes for that.

holgerfriedrich:
@mstormi this is intended and works.
Is it the correct approach to add `# shellcheck disable=SC2016` before this line?

PR #654, opened 2019-08-06
fix tellstick dependencies for Buster

holgerfriedrich:
libconfuse1 is required for package telldus-core.
For Buster it is not available, and the workaround seems to be to get it from Stretch.
Install works fine on Raspbian Buster. Now hardware to test functionality.

### DIFF
--- a/functions/packages.bash
+++ b/functions/packages.bash
@@ -621,7 +621,7 @@ nginx_setup() {
 
 tellstick_core_setup() {
   FAILED=0
-  introtext="[CURRENTLY BROKEN] This will install tellstick core services to enable support for USB connected Tellstick and Tellstick duo. For more details, have a look at http://developer.telldus.se/"
+  introtext="This will install tellstick core services to enable support for USB connected Tellstick and Tellstick duo. For more details, have a look at http://developer.telldus.se/"
   failtext="Sadly there was a problem setting up tellstick core. Please report this problem in the openHAB community forum or as a openHABian GitHub issue."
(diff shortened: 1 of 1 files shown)


holgerfriedrich:
@EliasGabrielsson can you give input why this is marked broken in the menu?

EliasGabrielsson:
`[CURRENTLY BROKEN]` was a "quick-fix" when releasing 1.5. Though that few people used this plugin so it could wait. Good that you fixed this. Can you remove the text?

I think this workaround will need to be in the codebase longterm. The main repository seams to be stale. In the best of world would an upstream patch be made using `libconfuse2`.

Thread about it: https://forum.telldus.com/viewtopic.php?t=15232

EliasGabrielsson:
`[CURRENTLY BROKEN]` was a "quick-fix" when releasing 1.5. Though that few people used this plugin so it could wait. Good that you fixed this. Can you remove the text?

I think this workaround will need to be in the codebase longterm. The main repository seams to be stale. In the best of world would an upstream patch be made using `libconfuse2`.

Thread about it: https://forum.telldus.com/viewtopic.php?t=15232

holgerfriedrich:
@EliasGabrielsson warning removed. Ready to merge.

I will not look into telldus upstream since I do not use this device.

EliasGabrielsson:
Thanks for the help! 
I actually use the device myself so will keep my self updated.
The Telldus-Core software is for legacy devices, still doing their job but not produced any longer.

Issue #655, opened 2019-08-07
Make openHABian less dependent of upstreams


mstormi:
We have a couple of places where - in addition to where we use debian repos - we wget some URL or git clone Github repos.
We're usually retrieving the latest version which means it is not known to work at the time of install. Some of these instances even require to compile code, others in turn use yet another upstream.

I suggest we increase reliability by either of these methods or combinations thereof:
* fetch _tagged_ git versions only that we know to work
  (instead of master or anything that's subject to change quickly and without that we notice).
* create a cache by creating github mirror repos (or mirror them into openhabian repo)
* fetch packages and store them in openhabian repo for download
* compile binaries at image creation time and store them in openhabian repo for download

Instances to work on (will track here, let me know more via comments)

Critical (any user needs it):
~~openhabian repo itself~~
~~openhabian-zram, overlayfstools~~
~~Zulu Java~~, ~~Adopt Java~~, (OpenJDK ?)

Optional:
knxd
Tellstick
miflora daemon

Fixed for zram in PR #656 
Created 2 nested(!) mirror repos and added/use "openhabian_v1.5" tag for working code.

EliasGabrielsson:
I actually discussed this topic with @ThomDietrich not long ago. First off all I don't see "one" correct way of solving this matter. I was exploring the thought of providing a verified configuration, per-installed image which is somewhat what you are suggesting. We ended up the discussion that is about where to put the maintenance resources and that when issues have occurred they have been solved rather quickly.

But maybe we need cherrypick some unstable upstream packages better. Which things have been-longterm broken so far because of upstream patches weren't merged?

mstormi:
zram, knxd and Tellstick I believe, all of which were not ready for Buster.
Agree there's no single "correct" way but I think what I did for zram is a good solution for Github repos.
Mirror repos effectively don't change unless we actively pull/merge the upstream so they cost no maintenance resources unless we're willing to spend them as part of component or openHABian upgrade activities. Meanwhile no untested upstream can break installs.

EliasGabrielsson:
Hmm, I think we should as long as possible stay with original repos to easliy track work. We should only use packages which are well maintained. It's always possible to checkout a single commit as well.

mstormi:
[...]

PR #656, opened 2019-08-07
fix zram creation

mstormi:
The upstream's 4yr old (!) upstream failed to compile on buster (it worked on stretch).
Created mirror repos so openHABian now will only ever clone tagged branches known to work.

### DIFF
--- a/docs/openhabian.md
+++ b/docs/openhabian.md
@@ -223,7 +223,6 @@ There are four measuers and methods included to cover this matter today but they
 3. (BETA) Reduce wear on SD-card by move write intensive actions temporary to RAM during operation (logs,persistant-data). Warning: powerfailure will result in lost data. [Menu option: 6A]
 4. (Advanced) Use [Amanda Network Backup](http://www.amanda.org/) for full-system backup, longer introduction [here](https://github.com/openhab/openhabian/blob/master/docs/openhabian-amanda.md). [Menu option: 51]
 
-
 ## Optional Components
 
 openHABian comes with a number of additional routines to quickly install and set up home automation related software.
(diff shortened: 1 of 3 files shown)


EliasGabrielsson:
Why didn't you provide a patch upstream instead?

mstormi:
I did but the upstream's upstream didn't respond so 2nd best was to mirror and fix it in my copy.

mstormi:
@EliasGabrielsson mind merging? actually the clone tag makes for a difference, I think it won't build/install on buster without.

mstormi:
I've changed the repo tag in use to point to the right code so it should work for now

EliasGabrielsson:
I don't get it, upstream was merging your code: https://github.com/kmxz/overlayfs-tools/commit/1cdf251eec41bf483c700c4dc794c58c88401fe3 ??

I also wonder why we not are using the original branch: https://github.com/StuartIanNaylor/zram-config . There have already been patches and will be many more, how do we make sure that we are uptodated? 
I also don't get why you haven't made a fork of the library so it will be easier to track how the repositories diverts from each other. At some point in future we will need to merge them together again.

mstormi:
> I don't get it, upstream was merging your code: [kmxz/overlayfs-tools@1cdf251](https://github.com/kmxz/overlayfs-tools/commit/1cdf251eec41bf483c700c4dc794c58c88401fe3) ??

Now he did, yes, but not by the time I made this PR.
Please merge this asap.
I'm not sure yet (very very difficult to debug) but there's people to have issues with zram this might fix.

mstormi:
> I also wonder why we not are using the original branch: https://github.com/StuartIanNaylor/zram-config
> There have already been patches and will be many more, how do we make sure that we are uptodated?

Let's continue discussion on #655 for that - we don't want updates to happen automatically.
But we can enforce them at any time by updating the repo mirror.

PR #671, opened 2019-08-23
fix zram uninstall

mstormi:


EliasGabrielsson:
Just for curiosity why didn't bash work to run the script?

mstormi:
bash would have worked too (the original author used this sh so I kept that)
but the file was not executable so the leading sh was missing

Issue #431, opened 2018-08-11
Frontail installation issues


ConfusedTA:
Hi all,

Have just tried to install Frontail on my headless Ubuntu 18.04 box, but the installation is giving the following messages, and Frontail service doesn't start successfully.

```
2018-08-11_13:51:45_BST [openHABian] Installing the openHAB Log Viewer (frontail)...
$ npm install -g frontail
npm WARN deprecated connect@2.11.0: connect 2.x series is deprecated
/usr/local/bin/frontail -> /usr/local/lib/node_modules/frontail/bin/frontail
/usr/local/lib
└── frontail@4.2.2


$ npm update -g frontail
cp: cannot create regular file '/usr/lib/node_modules/frontail/preset/openhab.json': No such file or directory
cp: cannot create regular file '/usr/lib/node_modules/frontail/web/assets/styles/openhab.css': No such file or directory

$ systemctl daemon-reload

$ systemctl enable frontail.service

$ systemctl restart frontail.service
OK
```

Service status:

```
confused@openhab:~$ sudo systemctl status frontail.service
● frontail.service - Frontail openHAB instance, reachable at http://openhab:9001
   Loaded: loaded (/etc/systemd/system/frontail.service; enabled; vendor preset: enabled)
   Active: failed (Result: exit-code) since Sat 2018-08-11 13:51:51 BST; 3min 10s ago
     Docs: https://github.com/mthenw/frontail
           https://github.com/openhab/openhabian/issues/28
  Process: 17446 ExecStart=/usr/bin/frontail --ui-highlight --ui-highlight-preset /usr/lib/node_modules/frontail/preset/
 Main PID: 17446 (code=exited, status=203/EXEC)

Aug 11 13:51:51 openhab systemd[1]: frontail.service: Main process exited, code=exited, status=203/EXEC
Aug 11 13:51:51 openhab systemd[1]: frontail.service: Failed with result 'exit-code'.
Aug 11 13:51:51 openhab systemd[1]: frontail.service: Service hold-off time over, scheduling restart.
Aug 11 13:51:51 openhab systemd[1]: frontail.service: Scheduled restart job, restart counter is at 5.
Aug 11 13:51:51 openhab systemd[1]: Stopped Frontail openHAB instance, reachable at http://openhab:9001.
Aug 11 13:51:51 openhab systemd[1]: frontail.service: Start request repeated too quickly.
[...]

PR #674, opened 2019-08-24
Frontail service fix

mstormi:
Not sure changing path in https://github.com/openhab/openhabian/pull/672 was a good idea, some people to complain on the forum.
Played it safe and now allow for both locations.
Also included https://github.com/openhab/openhabian/issues/431

### DIFF
--- a/functions/nodejs-apps.bash
+++ b/functions/nodejs-apps.bash
@@ -39,6 +39,7 @@ frontail_setup() {
   cond_redirect npm install -g frontail
   if [ $? -ne 0 ]; then echo "FAILED (frontail)"; exit 1; fi
   cond_redirect npm update -g frontail
+  ln -sfn /usr/bin/frontail /usr/local/bin/frontail
   #
   mkdir -p ${frontail_base}/preset ${frontail_base}/web/assets/styles
   
--- a/includes/frontail.service
+++ b/includes/frontail.service
@@ -5,13 +5,13 @@ After=openhab2.service
 
 [Service]
 Type=simple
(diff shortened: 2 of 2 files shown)


EliasGabrielsson:
These line should rather be:
```
Environment=PATH=/usr/local
```
```
WorkingDirectory=/usr/bin/
```
Sorry for not noticed that before. What kind of errors do people have?
Are there only new installation which have trouble. Regarding your git conflict resolve it by rebase ontop of upstream master as usual.

mstormi:
https://community.openhab.org/t/frontail-not-starting-after-fresh-install-openhabian/80597
only new installs apparently
Put in /usr/local/bin now 2 times.
What git conflict do you mean?

EliasGabrielsson:
I must been dizzy, I mixed up `Frontail` and `FireMotD`. 
The service file was before your "fix", just checked on a new installation.
There must been some other funky things going on for the one in the forum.

mstormi:
So the service file was correct before my yesterday's "fix" and is broken now so I just reverted that.
_I_ might have been dizzy yesterday too but I wonder why I had to change the path for the install test to complete. Do you mind if I keep the link /usr/local/bin/frontail->/usr/bin/frontail for safety reasons? I also left the WorkingDir change in as that might help some.

EliasGabrielsson:
Installation works at least on x86 (see pushed test.) If there are some issue they are ARM / Debian Buster specific. Let's first of all get grip of what is the cause of the issue.

EliasGabrielsson:
The old place for Frontail installation was: `/usr/local/bin/frontail`
The new one is: `/usr/bin/frontail`

mstormi:
ok will open another issue in case something comes up new

PR #676, opened 2019-08-24
Revert "fix frontail service startup (#672)"

EliasGabrielsson:
This reverts commit b52bd84a0be2f9abc48e1ce1533579c30e181635.

### DIFF
--- a/includes/frontail.service
+++ b/includes/frontail.service
@@ -6,7 +6,7 @@ After=openhab2.service
 
 [Service]
 Type=simple
-ExecStart=/usr/local/bin/frontail --ui-highlight --ui-highlight-preset /usr/lib/node_modules/frontail/preset/openhab.json -t openhab -l 2000 -n 200 /var/log/openhab2/openhab.log /var/log/openhab2/events.log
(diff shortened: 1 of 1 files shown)


mstormi:
[This](https://travis-ci.com/mstormi/openhabian/jobs/228052105) is from my current-most build. Line 1294 is why I changed frontail path to /usr/local/bin.
So can the path depend on npm config or package version ?
Shouldn't we play it safe and add at least a link ?

mstormi:
Even worse now: https://travis-ci.com/mstormi/openhabian/jobs/228391498
Where's frontail installed ??

EliasGabrielsson:
It works in master so it must be something in your branch or temporary install issues. See: https://travis-ci.org/openhab/openhabian/builds/576713562

mstormi:
It's because npm lives in /usr/local/lib/node_modules in the base image I'm using.
As it might be the same on other non-RPi-image installs I guess we should check for that and adapt $frontail_base

EliasGabrielsson:
Sounds fair. Lets make the installation aware of where node_modules are.

mstormi:
https://github.com/openhab/openhabian/pull/688

PR #682, opened 2019-08-26
adapt bats tests to upcoming build system

mstormi:
Signed-off-by: Markus Storm <markus.storm@gmx.net>

PR #683, opened 2019-08-26
parametrize for production, test and verbose debugging install runs

### LINKED ISSUE #408: Create new log file [first-boot-debug.log] containing all output while installing (instead of pipe to /dev/null)
The first-boot.sh script runs for a long time and seemingly fails for a fair amount of people for a wide variety of reasons.

I find it a very bad idea that it sends command output to /dev/null so the user cannot see the progress of the commands and neither the error messages when the script fails.

I recommend using section markers (like "------------------" before executing a command and leaving the command output intact (displayed on the screen and in the log file).

### LINKED ISSUE #735: Debugging instructions
Following up on a number of issues users brought up in response to 1.5 and comprehensive changes like the shellcheck stuff, I came to think that we should create a debug instructions README.
This (its Github link to always point to its current version) we can reference community users to so we don't have to individually help with explaining them how they would need to debug openHABian installs that fail.

Some unsorted, incomplete thoughts what should be contained, please add your own suggestions:
* prerequisites a user has to meet such as an (_totally?_) unfiltered Internet connect and a router to provide proper DNS and DHCP resolving (_is NTP mandatory, too?_)
* the location of the logfile(s) (_only first-boot.log ?_) to post
* a quick intro how to open a Github issue and - most noteworthy - what information to upload
* debugging options (`unattended`, `unattended_debug `and `debug_maximum`)
* the procedure to use in re-running an openhabian setup that failed with changed parameters
I'm not 100% sure of this myself. Can we change the `mode` parameter in `/etc/openhabian.conf`,
`rm -f /opt/openHABian-install-failed` and reboot ? If no please add the right instructions.

See also #408 and #683. In that latter I introduce another most-verbose debug level "debug_maximum" and the ability to define the debug level in openhabian.conf[.dist] which then is used by openhabian-set.sh and automated tests.
We might get to get around the need expressed in #408 to create a complementary log by only ever creating first-boot.log, have it remain to be the non-verbose default in the beginning but then instruct users when and how to switch to more verbose levels.

### LINKED ISSUE #808: Guide to open an issue
Users sometimes open rather useless issues like #806 because they're told to by the last line in `menu.bash`. I think we should remove that link and eventually put up one instead that is more verbose
such as https://community.openhab.org/t/how-to-file-an-issue/68464.
Also I have started an openHABian DEBUG guide we might want to refer to, see https://github.com/mstormi/openhabian/blob/buildsystem-init.bash/docs/openhabian-DEBUG.md
To make it effective we would need to merge some enhancements which are part of #683 first, though.

### LINKED ISSUE #839: Installation of Zulu 64 bit fails (at least on aarch64)
Installing the 64bit Zulu Java fails on aarch64.

To reproduce on aarch64:

1. Run openhabian-config
2. Select Fresh/Manual Install
3. Select only Zulu Java 64 bit
4. Start installation process

Zulu 32bit for armhf is installed instead of the 64 bit version.

The reason is that the java_arch variable is not written correctly into /etc/openhabian.config.

functions/config.bash saves an option called "option" into openhabian.config:
```
update_config_java() {
  cond_redirect grep -q '^java_arch' "$CONFIGFILE" && sed -i "s/^java_arch.*/java_arch=$1/" "$CONFIGFILE" || echo "option=$1" >> "$CONFIGFILE"
  # shellcheck disable=SC1090
  source "$CONFIGFILE"
}
```

The line
```
   cond_redirect grep -q '^java_arch' "$CONFIGFILE" && sed -i "s/^java_arch.*/java_arch=$1/" "$CONFIGFILE" || echo "option=$1" >> "$CONFIGFILE"
```

Should be
```
  cond_redirect grep -q '^java_arch' "$CONFIGFILE" && sed -i "s/^java_arch.*/java_arch=$1/" "$CONFIGFILE" || echo "java_arch=$1" >> "$CONFIGFILE"
```

This fix works for me. :)

mstormi:
Fixes #408
Fixes #808 
eventually also fixes #735 
eventually also fixes #839


mstormi:
@EliasGabrielsson @ThomDietrich 
I started a debugging guide but I'm certainly missing a number of items and TODOs that should be in there so I kindly request you to work together on 
https://github.com/mstormi/openhabian/blob/buildsystem-init.bash/docs/openhabian-DEBUG.md

holgerfriedrich:
can you please rebase after merginig #682?

mstormi:
Rebased against my repo from #682 instead of master. Hope that works out when we merge #682.

holgerfriedrich:
When I look at "Files changed", I see a lot of stuff which is already merged to master. Pls rebase against master and check if this vanishes...

mstormi:
Had to manually resolve a lot of stuff. Did so but cannot validate/test at the moment because Travis is down for maintenance.

build-image/first-boot.bash
```diff
@@ -2,11 +2,16 @@
 
 REPOSITORYURL=$(grep -i '^repositoryurl' /etc/openhabian.conf | cut -d '=' -f2)
```
holgerfriedrich, line 3: the strange thing about this 3 parameters is, that they are copied to openhab.conf... and the typical build-image/openhabian.conf looks quite different....
so a normal install first-boot.bash would not see any of them....

Dockerfile.openhabian
```diff
@@ -0,0 +1,56 @@
...
+RUN cp docker-systemctl-replacement/files/docker/systemctl.py /bin/systemctl
+
+ENTRYPOINT ["/usr/bin/entry.sh"]
+
+RUN adduser openhabian --gecos "Openhabian,,," --disabled-password
+RUN echo "openhabian:openhabian" | chpasswd   
+RUN /bin/echo -n "Running on " && /usr/bin/arch
+
+COPY . /opt/openhabian/
+COPY openhabian.conf.dist /etc/openhabian.conf
```
holgerfriedrich, line 47: hm, we have a build-image/openhabian.conf, which defines usernames, java_arch, etc which might be also relevant for the test.... 
though it might be fine to use this settings from the .dist file.
btw: if openhabian.conf.dist is only for testing-docker, move it to the dedicated directory

entry.sh
```diff
@@ -0,0 +1,71 @@
+#!/bin/bash
```
holgerfriedrich, line 1: see above, move to folder

openhabian-setup.sh
```diff
@@ -10,19 +10,6 @@
...
-  BASEDIR="$( cd -P "$( dirname "$SOURCE" )" && pwd )"
-  SOURCE="$(readlink "$SOURCE")"
-  [[ $SOURCE != /* ]] && SOURCE="$BASEDIR/$SOURCE"
-done
-BASEDIR="$( cd -P "$( dirname "$SOURCE" )" && pwd )"
-SCRIPTNAME="$(basename "$SOURCE")"
-
-REPOSITORYURL="https://github.com/openhab/openhabian"
-CONFIGFILE="/etc/openhabian.conf"
-
```
holgerfriedrich, line 25: **if you want to factor this out, remember sourcing init.sh here!**
otherwise this will break openhabian-config!

.travis.yml
```diff
@@ -1,7 +1,7 @@
-language: bash
-
+language: shell
+os: linux
 # using latest Ubuntu (not the travis-ci default) to get latest tool versions in Travis environment
-dist: disco
+dist: bionic
```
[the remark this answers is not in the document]
mstormi: Noticed a Travis log error msg saying it does not know "disco". May or may not have changed in recent Travis maintenance, "bionic" validates fine

.travis.yml
```diff
@@ -20,7 +20,11 @@ install:
 # some tests such as zram.bats may fail if we run in emulation mode (testing for ARM emulated on x86 Travis cloud servers)
   - sudo apt-get install virt-what && VIRT=$(sudo virt-what) && echo "Travis VM running on ${VIRT:-native HW}."
   - docker run --rm --privileged multiarch/qemu-user-static:register --reset
-  - docker build -t openhabian/rpi-openhabian -f Dockerfile .
+  - sed -i 's/mode=unattended$/mode=unattended_debug/' openhabian.conf.dist
```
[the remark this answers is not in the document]
mstormi: ok

.travis.yml
```diff
@@ -20,7 +20,11 @@ install:
...
+  - #sed -i 's/mode=unattended$/mode=debug_maximum/' openhabian.conf.dist
```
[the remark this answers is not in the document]
mstormi: ok

Dockerfile.openhabian
```diff
@@ -0,0 +1,56 @@
...
+#FROM balenalib/raspberry-pi2-debian-node:latest-buster
+
+
+ENV container docker
+ENV INITSYSTEM on
+ENV APT_KEY_DONT_WARN_ON_DANGEROUS_USAGE=DontWarn
+
+RUN apt-get update -qq \
+    && apt-get install -y -qq --no-install-recommends systemd systemd-sysv git wget python python-pip python3 python3-pip apt-utils jq virt-what \
+    && rm -rf /var/lib/apt/lists/*
```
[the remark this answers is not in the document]
mstormi: https://www.balena.io/docs/learn/deploy/build-optimization/#pro-tips-for-slimming-down-your-build

Dockerfile.openhabian
```diff
@@ -0,0 +1,56 @@
+FROM minimum2scp/systemd-buster
+#FROM balenalib/intel-nuc-debian-node:latest-buster
+#FROM balenalib/raspberry-pi2-debian-node:latest-buster
+
+
+ENV container docker
+ENV INITSYSTEM on
+ENV APT_KEY_DONT_WARN_ON_DANGEROUS_USAGE=DontWarn
+
+RUN apt-get update -qq \
```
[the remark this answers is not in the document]
mstormi: ok

Dockerfile.openhabian
```diff
@@ -0,0 +1,56 @@
...
+    && apt-get install -y -qq --no-install-recommends systemd systemd-sysv git wget python python-pip python3 python3-pip apt-utils jq virt-what \
+    && rm -rf /var/lib/apt/lists/*
+RUN git clone https://github.com/bats-core/bats-core.git && \
```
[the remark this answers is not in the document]
mstormi: ok

Dockerfile.openhabian
```diff
@@ -0,0 +1,56 @@
...
+    sys-kernel-config.mount \
+    display-manager.service \
+    getty@.service \
+    systemd-logind.service \
+    systemd-remount-fs.service \
+    getty.target \
+    graphical.target
+
+WORKDIR /
+COPY entry.sh /usr/bin/entry.sh
```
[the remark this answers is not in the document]
mstormi: Frankly I cannot explain all of the concept in detail. It's based on Balena works which in turn is for generic deployments (take this service as an example).
I don't want to deviate from their default proposal where not needed so we don't have to catch up with them whenever there's changes.
Doing so would be a potential error source and costs time & work, in particular if we discuss  rationales every time (and I'm not on Docker expert level to explain or challenge their setup).

For an entry point, I'll add this link as a comment
https://www.balena.io/docs/reference/base-images/base-images/#how-the-images-work-at-runtime

Dockerfile.openhabian
```diff
@@ -0,0 +1,56 @@
...
+RUN echo "openhabian:openhabian" | chpasswd   
+RUN /bin/echo -n "Running on " && /usr/bin/arch
+
+COPY . /opt/openhabian/
+COPY openhabian.conf.dist /etc/openhabian.conf
+
+WORKDIR /opt/openhabian/
+COPY start.sh ./
+
+VOLUME ["/sys/fs/cgroup"]
```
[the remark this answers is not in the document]
mstormi: see general comment

Dockerfile.openhabian
```diff
@@ -0,0 +1,56 @@
...
+STOPSIGNAL 37
```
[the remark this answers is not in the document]
mstormi: see general comment

start.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+while : ; do
```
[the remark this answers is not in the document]
mstormi: again, it's the Balena source copied so would not like to modify.
And you understood what it's doing without that comment, too :)

start.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
mstormi: rather not, see general comment

resin.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
```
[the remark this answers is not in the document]
mstormi: rather not, see general comment

Dockerfile.openhabian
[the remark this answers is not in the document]
holgerfriedrich: good, but please add `# delete temporary files to reduct build size`

Dockerfile.openhabian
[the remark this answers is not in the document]
holgerfriedrich: 1) Moving the files related to the docker testing to a separate directory has high priority for me. Getting path names wrong during an update will break the build anyways, so not a good chance to introduce errors. 
I do not like to mix up the build/install related files with testing environment.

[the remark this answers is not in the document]
holgerfriedrich: 2) Thanks for the link, I will have a look once I find some time.
Did you check their license? I quickly had a look and could not find anything. We need to be sure that we are allowed to include - even sample code from a website - into our code base. This is essintially what you state with the signed-off.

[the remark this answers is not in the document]
holgerfriedrich: btw: could you please leave the comments as unresolved if you do not agree with a proposal? this makes it easier for me to find the relevant comments which I need to check again. Thanks!

Dockerfile.openhabian
mstormi: it wasn't me to introduce openhabian.conf.dist so I presume it has another use and I did not want to break that. Don't fully know though what it has been there for.
On build/*conf, I'm still struggling with "wrong" java in #684 but have not seen any other impact  to date.

build-image/first-boot.bash
mstormi: This I was struggling with, too, and it was the reason to intro init.bash which is copying the relevant variables from there. Some of these I found to be missing and to be a problem on builds.

start.sh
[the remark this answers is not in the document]
mstormi: ok

resin.service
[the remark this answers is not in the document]
mstormi: ok

entry.sh
mstormi: ok

openhabian-setup.sh
mstormi: you mean sourcing the newly introduced init.bash ?
It is sourced from build-image/first-boot.bash

functions/openhab.bash
```diff
@@ -64,11 +64,7 @@ Check the \"openHAB Release Notes\" and the official announcements to learn abou
   cond_redirect adduser openhab gpio
   cond_redirect systemctl daemon-reload
   if cond_redirect systemctl enable openhab2.service; then echo "OK"; else echo "FAILED (usr)"; exit 1; fi
-  if [ -n "$UNATTENDED" ]; then
-    cond_redirect systemctl stop openhab2.service || true
-  else
-    cond_redirect systemctl restart openhab2.service || true
-  fi
+  cond_redirect systemctl restart openhab2.service || true
```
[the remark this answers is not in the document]
mstormi: ok

Dockerfile.openhabian
mstormi: now sourced/applied java_arch, too

Dockerfile.openhabian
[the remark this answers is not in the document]
mstormi: ok

Dockerfile.openhabian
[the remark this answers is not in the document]
mstormi: sure, didn't I? If so that was by accident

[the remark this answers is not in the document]
mstormi: I could not find anything on their website to either permit or deny anything with respect to their docs and code contained therein. On images, closest to match is https://github.com/balena-io-library/base-images/blob/master/LICENSE which I believe is fine for our use.

docs/openhabian-DEBUG.md
```diff
@@ -0,0 +1,93 @@
...
+:::
+
+The main program is in `openhabian-setup.sh`. If the initial unattended install fails again and again at the same step (say Java installation), you may comment that step out. But mind the code in build-image/first-boot.bash quite towards the end starting with`git clone`. This is where openHABian updates itself. If you don't comment that out as well, it'll overwrite your changes on next install run.
+
+::: warning Disclaimer
+For obvious reasons, this is not a supported procedure. We just want to give you a hint what you _could_ try doing if your install fails
+and you're sitting there, desperately looking for a fix.
+G*oo*gle and learn yourself what you need to edit a file, learn to understand shell programming basics, you're on your own here.
+If you change openHABian code on your box, temember for the time it takes to get openHABian officially fixed, you must not let openhabian-config update itself on start as that would overwrite your changes.
```
holgerfriedrich, line 92: temember -> remember

btw: are you sure that :::warning and :::tip works correctly? if I open as file, thhis is not processed.

functions/openhab.bash
```diff
@@ -63,12 +63,7 @@ Check the \"openHAB Release Notes\" and the official announcements to learn abou
   if ! cond_redirect apt-get ${APT_INST_OPTS} install "openhab2=${openhabVersion}"; then echo "FAILED (apt)"; exit 1; fi
   cond_redirect adduser openhab gpio
   cond_redirect systemctl daemon-reload
-  if cond_redirect systemctl enable openhab2.service; then echo "OK"; else echo "FAILED (usr)"; exit 1; fi
-  if [ -n "$UNATTENDED" ]; then
```
holgerfriedrich, line 67: what was this intended for? why are you sure you can remove?

build-image/first-boot.bash
```diff
@@ -1,12 +1,12 @@
...
-CLONEBRANCH=$(grep -i '^clonebranch' /etc/openhabian.conf | cut -d '=' -f2)
-export REPOSITORYURL CLONEBRANCH
-
 # apt/dpkg commands will not try interactive dialogs
 export DEBIAN_FRONTEND=noninteractive
 
+source "init.bash"
+# shellcheck disable=SC1090
+source "$CONFIGFILE"
+
```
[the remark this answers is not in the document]
mstormi: I'm confused ... I think it *is* working according to the Travis logs.
So what's your proposal here ?

init.bash
```diff
@@ -0,0 +1,19 @@
...
+  [[ $SOURCE != /* ]] && SOURCE="$BASEDIR/$SOURCE"
+done
+BASEDIR="$( cd -P "$( dirname "$SOURCE" )" && pwd )"
+SCRIPTNAME="$(basename "$SOURCE")"
+
+CONFIGFILE="/etc/openhabian.conf"
+REPOSITORYURL=$(grep -i '^repositoryurl' ${CONFIGFILE} | cut -d '=' -f2)
+CLONEBRANCH=$(grep -i '^clonebranch' ${CONFIGFILE} | cut -d '=' -f2)
+JAVA_ARCH=$(grep -i '^java_arch' ${CONFIGFILE} | cut -d '=' -f2)
+MODE=$(grep -i '^mode' /etc/openhabian.conf | cut -d '=' -f2)
```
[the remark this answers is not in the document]
mstormi: ok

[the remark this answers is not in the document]
mstormi: I used uppercase to ensure we only ever use what we obtain through init.bash.
Frankly I'm not sure we "source the config" as I have seen unset (lowercase) variables at times,
that's why this whole init.bash introduction

openhabian.conf.dist
```diff
@@ -26,5 +26,12 @@ wifi_psk=""
 # repo to clone from (if not master for build and test)
 repositoryurl=https://github.com/openhab/openhabian.git
 clonebranch=master
+# debug verbosity mode
+# unattended, unattended_debug (verbose output in log) or debug_maximum (show every command)
+mode=unattended
+
+# Java architecture mode
+# Use "32-bit" or "64-bit"
+java_arch="32-bit"
```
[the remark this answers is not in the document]
mstormi: We need to sort this .conf mess out but that's in the next step/PR.
For a start arm32hf is the target so this is the better default.

docs/openhabian-DEBUG.md
```diff
@@ -0,0 +1,93 @@
...
+
+<!-- Attention authors: Do not edit directly. Please add your changes to the appropriate source repository -->
+
+::: tip Purpose
+This document is a to give a guiding hand to users when their openHABian install fails in the first place.
+Read on to find out how to improve debug verbosity and how to proceed with that information.
+:::
+
+## Prerequisites
+First, please make sure you use proper host hardware that is supported as per README.
```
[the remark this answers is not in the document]
mstormi: Not really. (Updated) README is in PR #730 but that's not yet merged (and we have not agreed on HW to support, have we? PINE will be out but what about Pi0W, CM3 ?)

docs/openhabian-DEBUG.md
mstormi: no so I removed it

functions/openhab.bash
mstormi: See https://github.com/openhab/openhabian/blob/master/functions/openhab.bash#L67
... enable --now ... replaces all of this (and it works as can be seen obn Travis)

openhabian.conf.dist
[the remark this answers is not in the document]
holgerfriedrich: ...but here (test) and in build.bash we can easily change via sed, in a manual install you may end up using the conf.dist as it is.

build-image/first-boot.bash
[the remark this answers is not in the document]
holgerfriedrich: yes in travis it is working fine, as Dockerfile.openhabian copies the patched conf file to /etc/openhabian.conf before starting the tests.
Installing on a RPi, this is done only a few lines later (it is taken from /boot and copied over).
init.sh fails, as it is called earlier.

docs/openhabian-DEBUG.md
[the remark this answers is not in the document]
holgerfriedrich: ok, fine for me now, will come later.
HW discussion in the forum.

functions/openhab.bash
holgerfriedrich: my question was about the if "unattedded" is ture - I see a stop for the case in the old code. Not we changed to enable and run for both cases.

mstormi: the old ... stop IMHO was a bug, we want it to run in unattended mode, too (else final check will fail in Travis and on real HW, too)

openhabian.conf.dist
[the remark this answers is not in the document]
mstormi: I see your point, manual install is most likely to be 64bit so I'll change that to be the default.
_... bei dem Sch... kommt man ganz durcheinander, wenn man gleichzeitig noch an 32/64Bit Builds baut_

build-image/first-boot.bash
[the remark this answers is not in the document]
mstormi: _ok, ich sehe inzwischen den Wald vor lauter Code nicht mehr. Schlag mal eine Alternative vor._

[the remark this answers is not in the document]
holgerfriedrich: I have added a commit which re-orders the commands. Pls have a look.

init.bash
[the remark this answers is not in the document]
holgerfriedrich: actually I do not like having both versions (via init.bash in capital letters and via sourcing the config) - that is what I have seen in first-boot.bash.
I would actually prefer not to source any config file. sourcing could be removed from first-boot.bash (if something is missing, lets add it to init.bash).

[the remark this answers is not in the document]
mstormi: Granted it's doing things twice. I'll see to remove init.bash and go back to lowercase variables sourced from openhabian.conf.
Will also reorder sourcing in first-boot.bash (I saw you were working on it)

openhabian.conf.dist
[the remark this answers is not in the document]
mstormi: Moved to use build-image/openhabian.conf as source.

init.bash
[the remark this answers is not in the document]
mstormi: I rather decided myself to leave sourcing in because it was always built to work like that. 
Besides sourcing is a clever sort of parsing a config file, isn't it ? If there's anything that we would need to store in .conf that's incompatible with bash syntax , we can still change it and move over to init.bash, but for the time being I'll keep it. Don't want to upset Thomas :)

mstormi:
> for me it looks good now - besides the init.bash call in first-boot.bash and the change of java to 32bit openhabian.conf.dist ....

Please have another - hopefully final - look now.

init.bash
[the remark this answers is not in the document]
holgerfriedrich: Yes, parsing the content could be beneficial, error messages are always better than error hiding. Though, we never should give write permissions to openhab user (maybe also warn about deviations on startup). Otherwise one could inject code. Currently we are safe (755).

.travis.yml
```diff
@@ -20,7 +20,12 @@ install:
 # some tests such as zram.bats may fail if we run in emulation mode (testing for ARM emulated on x86 Travis cloud servers)
   - sudo apt-get install virt-what && VIRT=$(sudo virt-what) && echo "Travis VM running on ${VIRT:-native HW}."
   - docker run --rm --privileged multiarch/qemu-user-static:register --reset
-  - docker build -t openhabian/rpi-openhabian -f Dockerfile .
+# prepare configuration for tests, select debug level here:
+  - #sed -iE 's#mode=.*$#mode=unattended_debug#' openhabian.conf.dist
+  - sed -iE 's#mode=.*$#mode=debug_maximum#' openhabian.conf.dist
+  - sed -i 's#repositoryurl=https://github.com/openhab/openhabian.git#repositoryurl=https://github.com/'${TRAVIS_REPO_SLUG}'.git#' build-image/openhabian.conf
+  - sed -i 's#clonebranch=master#clonebranch='${TRAVIS_BRANCH}'#' build-image/openhabian.conf
+  - docker build -t openhabian/rpi-openhabian -f Dockerfile.openhabian .
```
holgerfriedrich, line 28: getting confused, which file do you want to modify? (see Dockerfile below, you are copying both?!)
Before I had the impression you wanted to base the Docker test on conf.dist.

build-image/first-boot.bash
```diff
@@ -22,14 +18,20 @@ fail_inprogress() {
...
 rm -f /opt/openHABian-install-failed
 touch /opt/openHABian-install-inprogress
-
 echo -n "$(timestamp) [openHABian] Storing configuration... "
 cp /boot/openhabian.conf /etc/openhabian.conf
 sed -i 's/\r$//' /etc/openhabian.conf
-# shellcheck disable=SC1091
-source /etc/openhabian.conf
+
+source "init.bash"
```
holgerfriedrich, line 25: Sorry, I have overlooked this. We have a problem here:
You cannot just source it, it will only work if current directory matches the one of the script. It might be the case for first-boot.bash, but not in genreal.
Looking at current init.bash - isn't it easier to drop it and use the hard coded filename instead of $CONFIGFILE?

build-image/first-boot.bash
```diff
@@ -149,29 +151,29 @@ if hash python3 2>/dev/null; then bash /boot/webif.bash reinsure_running; fi
 echo -n "$(timestamp) [openHABian] Installing git package... "
 if apt-get -y install git &>/dev/null; then echo "OK"; else echo "FAILED"; fail_inprogress; fi
 
-echo -n "$(timestamp) [openHABian] Cloning myself from ${REPOSITORYURL}, ${CLONEBRANCH} branch... "
 if [ -d /opt/openhabian/ ]; then cd /opt && rm -rf /opt/openhabian/; fi
-if git clone -q -b "$CLONEBRANCH" "$REPOSITORYURL" /opt/openhabian; then echo "OK"; else echo "FAILED"; fail_inprogress; fi
+# shellcheck disable=SC2154
+echo -n "$(timestamp) [openHABian] Cloning myself from ${repositoryurl}, ${clonebranch} branch... "
+if git clone -q -b "$clonebranch" "$repositoryurl" /opt/openhabian; then echo "OK"; else echo "FAILED"; fail_inprogress; fi
```
holgerfriedrich, line 157: actually it seems we have broken all 3 developer options in build.bash: local-test, inject_build_repo.
local-test has two sed commands which need to be adapted (target: remove L#154, L#157, and at best also L#156)
inject_build_repo: last sed command to be adapted, and file is no longer $1 but /etc/openhabian.conf

build-image/first-boot.bash
```diff
@@ -149,29 +151,29 @@ if hash python3 2>/dev/null; then bash /boot/webif.bash reinsure_running; fi
...
 ln -sfn /opt/openhabian/openhabian-setup.sh /usr/local/bin/openhabian-config
 
-echo "$(timestamp) [openHABian] Executing 'openhabian-setup.sh unattended'... "
-if (/bin/bash /opt/openhabian/openhabian-setup.sh unattended); then
-#if (/bin/bash /opt/openhabian/openhabian-setup.sh unattended_debug); then
-  systemctl start openhab2.service
+# shellcheck disable=SC2154
+echo "$(timestamp) [openHABian] Executing openhabian-setup.sh ${mode}... "
+if (/bin/bash "$BASEDIR"/openhabian-setup.sh "$mode"); then
```
holgerfriedrich, line 162: when we drop init.bash -> back to hard coded (anyway hard coded below)

functions/system.bash
```diff
@@ -69,6 +69,8 @@ timezone_setting() {
 
 locale_setting() {
   cond_redirect apt-get -q -y install locales
+  # shellcheck disable=SC1090
+  source "$CONFIGFILE"
```
holgerfriedrich, line 73: see problem with init.bash above, consider hard coding the filename

init.bash
```diff
@@ -0,0 +1,19 @@
...
+BASEDIR="$( cd -P "$( dirname "$SOURCE" )" && pwd )"
+SCRIPTNAME="$(basename "$SOURCE")"
+
+CONFIGFILE="/etc/openhabian.conf"
+#REPOSITORYURL=$(grep -i '^repositoryurl' ${CONFIGFILE} | cut -d '=' -f2)
+#CLONEBRANCH=$(grep -i '^clonebranch' ${CONFIGFILE} | cut -d '=' -f2)
+#JAVA_ARCH=$(grep -i '^java_arch' ${CONFIGFILE} | cut -d '=' -f2)
+#MODE=$(grep -i '^mode' ${CONFIGFILE} | cut -d '=' -f2)
+
+export SOURCE BASEDIR SCRIPTNAME CONFIGFILE # REPOSITORYURL CLONEBRANCH JAVA_ARCH MODE
```
holgerfriedrich, line 19: as stated already, it might be hard to locate init.bash for sourcing - actually this is the first part of this file what you actually need to execute in the script which wants to source init.bash.... 
given that trouble, consider removing it.... see comment above.

openhabian-setup.sh
```diff
@@ -10,19 +10,6 @@
...
 
-# Find the absolute script location dir (e.g. BASEDIR=/opt/openhabian)
-SOURCE="${BASH_SOURCE[0]}"
-while [ -h "$SOURCE" ]; do
-  BASEDIR="$( cd -P "$( dirname "$SOURCE" )" && pwd )"
-  SOURCE="$(readlink "$SOURCE")"
-  [[ $SOURCE != /* ]] && SOURCE="$BASEDIR/$SOURCE"
-done
-BASEDIR="$( cd -P "$( dirname "$SOURCE" )" && pwd )"
-SCRIPTNAME="$(basename "$SOURCE")"
```
holgerfriedrich, line 21: back to this implementation w/o init.bash?

Dockerfile.openhabian
```diff
@@ -0,0 +1,66 @@
...
+
+ENTRYPOINT ["/usr/bin/entry.sh"]
+
+RUN adduser openhabian --gecos "Openhabian,,," --disabled-password
+RUN echo "openhabian:openhabian" | chpasswd   
+RUN /bin/echo -n "Running on " && /usr/bin/arch
+
+COPY . /opt/openhabian/
+COPY openhabian.conf.dist /etc/openhabian.conf
+COPY build-image/openhabian.conf /etc/openhabian.conf
```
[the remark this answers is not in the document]
mstormi: forgot to remove old line

.travis.yml
mstormi: initially yes. Now changed, using build-image/openhabian.conf

openhabian-setup.sh
mstormi: yes

build-image/first-boot.bash
mstormi: ok

functions/system.bash
mstormi: ok, I added configfile= to /etc/openhabian.conf. Sort of self referential but it is not a config option and I dislike hardcoding strings in many places

init.bash
mstormi: ok, removed it

build-image/first-boot.bash
mstormi: I added configfile= to /etc/openhabian.conf because I dislike hardcoding strings

build-image/first-boot.bash
mstormi: not sure what you mean here. Marked file is first-boot.bash but you talk about build.bash,
and L154 is no sed on neither of these ?
Please recheck against latest code I've just uploaded

mstormi:
it's building fine so what's left now ?
Not sure what you mean by https://github.com/openhab/openhabian/pull/683#discussion_r410747943
Can you validate build.bash is working? Never used that, actually.

build-image/first-boot.bash
holgerfriedrich: yes, I am talking about build.bash.
build.bash is patching first-boot.bash with a few sed commands. Those relay on the exact wording in first-boot.bash. So we need to adapt those commands in build.bash L152/L153.
local-test removes the got clone command with sed, we need to adapt to our new text.
inject_build_repo patches the cit clone url, this needs to be applied to /etc/openhabian.conf, so build.bash L63 needs this change.
Not a big thing....

build-image/openhabian.conf
```diff
@@ -1,6 +1,7 @@
 # Modify your openHABian settings
 # Handle with care, only touch if you need to!!
 
+configfile=/etc/openhabian.conf
```
holgerfriedrich, line 4: `# do not edit configfile location`

holgerfriedrich: ... and pls add an empty line below this setting

openhabian.conf.dist
```diff
@@ -3,7 +3,9 @@
 # This file will only be used on initial install.
 # If you want to change anything after install completed, use menu in openhabian-config.
 
+configfile=/etc/openhabian.conf
```
holgerfriedrich, line 6: see above

build-image/openhabian.conf
holgerfriedrich, line 4: I think this could lead to trouble.... Why put that in all the configs, with the risk somone things he could modify the location of the config here (which does not make sense, since the pointer to this file is anyway hardcoded).
why not
CONFIGFILE=/etc/openhabian.conf in first-boot.bash and openhabian-setup.sh? (maybe I overlooked some place where it is needed....) - the files in functions are anyway not intended to be used separately, are they?

functions/system.bash
```diff
@@ -46,7 +46,7 @@ needed_packages() {
 
 timezone_setting() {
   # shellcheck source=/etc/openhabian.conf disable=SC1091
-  source "$CONFIGFILE"
+  source "$configfile"
```
holgerfriedrich, line 49: I think those calls can be removed. if configfile is defined, it was already sourced... 
Also with my proposed change, it should have been anyway from the calling bash script.

openhabian.conf.dist
holgerfriedrich, line 6: see above...

openhabian.conf.dist
```diff
@@ -3,7 +3,9 @@
...
 hostname=openHABianDevice
+username=openhabian
```
holgerfriedrich, line 8: may not be needed, the dist fle is only used for manual install. not sure if then a openhabian user is created. but we can sort out this later.

mstormi: ok will leave it (it was inconsistent)

openhabian.conf.dist
mstormi: ok

functions/system.bash
mstormi: ok

build-image/openhabian.conf
mstormi: ok I did that and changed all references to uppercase CONFIGFILE so we do not accidentially confuse the origin of the variable

mstormi:
ok it built https://travis-ci.com/github/mstormi/openhabian/builds/160886146
We should come to an end, the more changes forth and back the higher the risk to overlook things.

build-image/first-boot.bash
mstormi: ok I think I understand but as you're more familiar with that would you please patch build.bash?
After we merged #683 
On a quick look, build.bash L152 still clones master so that should be reworked, too

holgerfriedrich: fine for me, it is anyway broken right now

mstormi: I quickly replaced those occurrences in local-test but are you sure about replacing $1 by "/etc/openhabian.conf" in last sed line of inject_build_repo ? sed looks to replace "master http://github.com....." but this is not contained as a string in there.
Wouldn't it rather be to replace the repositoryurl= and clonebranch= parameters ?

build.bash
```diff
@@ -53,14 +53,16 @@ get_git_repo() {
...
+    echo_process "inject_build_repo() invoked without clone_string variable set, exiting...."
     exit 1
   fi
   sed -i '$a /usr/bin/apt-get -y install figlet &>/dev/null' "$1"
   sed -i '$a echo "#!/bin/sh\n\ntest -x /usr/bin/figlet || exit 0\n\nfiglet \"Test build, Do not use!\" -w 55" > /etc/update-motd.d/04-test-build-text' "$1"
   sed -i '$a chmod +rx /etc/update-motd.d/04-test-build-text' "$1"
   sed -i '$a echo "$(timestamp) [openHABian] Warning! This is a test build."' "$1"
-  sed -i "s@master https://github.com/openhab/openhabian.git@$clone_string@g" "$1"
+#  sed -i "s@master https://github.com/openhab/openhabian.git@$clone_string@g" "$1"
+  sed -i "s@master@$clonebranch@g" "/etc/openhabian.conf"
```
holgerfriedrich, line 64: maybe safer to use "s@^clonebranch=.*@clonebranch=$clonebranch@g"

build.bash
```diff
@@ -53,14 +53,16 @@ get_git_repo() {
...
+  sed -i "s@https://github.com/openhab/openhabian.git@$repositoryurl@g" "/etc/openhabian.conf"
```
holgerfriedrich, line 65: ...and same here

mstormi: ok

build.bash
mstormi: ok

mstormi:
Damn it the very last check failed https://travis-ci.com/github/mstormi/openhabian/builds/161116555
Cosmetic only, but good to know it's doing what we want it to.

mstormi:
Phew, now that was a piece of work. Thanks Holger for your help & patience.

Issue #679, opened 2019-08-25
getting error invalid distance too far back


idkpmiller:
I am looking to update using openhabian-config and when I do I get this output.
I am on OH 2.4 stable and I installed openhabian a year or more ago, not sure when I last tried to update it.
Please advise.

`[14:15:03] openhabian@openhab:~$ sudo openhabian-config
2019-08-25_14:15:18_NZST [openHABian] Checking for root privileges... OK
2019-08-25_14:15:19_NZST [openHABian] Loading configuration file '/etc/openhabian.conf'... OK
2019-08-25_14:15:22_NZST [openHABian] openHABian configuration tool version: [master]v1.4.1-413(7799e0d)
2019-08-25_14:15:22_NZST [openHABian] Checking for changes in origin... error: inflate: data stream error (nvalid distance too far back)
error: failed to read object 629870b3ebe5ca46fb22e78de1e6c2283d125338 at offset 377572 from .git/objects/pak/pack-94dc81e2b904f0c1ab298b1fa191427bef519c38.pack
fatal: packed object 629870b3ebe5ca46fb22e78de1e6c2283d125338 (stored in .git/objects/pack/pack-94dc81e2b90f0c1ab298b1fa191427bef519c38.pack) is corrupt
fatal: index-pack failed
OK
2019-08-25_14:15:29_NZST [openHABian] Updating myself... error: inflate: data stream error (invalid distanc too far back)
error: failed to read object 629870b3ebe5ca46fb22e78de1e6c2283d125338 at offset 377572 from .git/objects/pak/pack-94dc81e2b904f0c1ab298b1fa191427bef519c38.pack
fatal: packed object 629870b3ebe5ca46fb22e78de1e6c2283d125338 (stored in .git/objects/pack/pack-94dc81e2b90f0c1ab298b1fa191427bef519c38.pack) is corrupt
fatal: index-pack failed
FAILED - There was a problem fetching the latest changes for the openHABian configuration tool. Please chec your internet connection and try again later...
2019-08-25_14:15:39_NZST [openHABian] Checking for default openHABian username:password combination... OK
2019-08-25_14:15:39_NZST [openHABian] We hope you got what you came for! See you again soon ;)
`

Regards
Paul

idkpmiller:
I have moved on some what.
I renamed the /opt/openhabian directory to .bak
I deleted the /usr/local/bin/openhabian-config file and then followed the install process as described here under other Linux system.
[...]

Issue #666, opened 2019-08-16
Downgrade from Snapshot to Testing fails


Harpo3:
The issue can be duplicated by changing to snapshot in openhabian-config, then changing to testing. Output is:

```
$ apt-get -y install openhab2=2.5.0~M2-1 
Reading package lists... Done
Building dependency tree       
Reading state information... Done
Suggested packages:
  zulu-8 | zulu-embedded-8 | oracle-java8-installer | openjdk-8-jdk-headless
The following packages will be DOWNGRADED:
  openhab2
0 upgraded, 0 newly installed, 1 downgraded, 0 to remove and 0 not upgraded.
E: Packages were downgraded and -y was used without --allow-downgrades.
FAILED (apt)
```

PR #686, opened 2019-08-26
do not use --allow-downgrades on Jessie

holgerfriedrich:
do not use option --allow-downgrades for apt on Jessie, refs #679 #666


EliasGabrielsson:
Thanks Holger for lookin into this 👍

PR #688, opened 2019-08-27
set frontail install directory relative to npm (nodejs) base

mstormi:
to fix #674 etc


mstormi:
any reason why not to apply this?

Issue #621, opened 2019-08-04
Deprecate support for Ubuntu [RFC]


EliasGabrielsson:
I suggest that we drop support for Ubuntu distros and only supports Debian based ones to reduce maintenance cost of the project. 

- [ ] Remove all traces of Ubuntu in the codebase
- [ ] Update Documentation
- [ ] Inform the user in the update dialog that Ubuntu based distros have been deprecated.

mstormi:
Not sure I agree here.
Let's make sure first we talk about x86 only here (there's also Ubuntu on Raspi but that never has been supported so I think this talk is not about that).
Now there's way more users (out of those to run x86) that run Ubuntu than do run plain Debian x86.
A quick glance at the code shows Ubuntu specific code in InfluxDB and Homegear installation.
Both today seem to provide Ubuntu and Debian specific builds - not sure if they _require_ users to install this specific package (for InfluxDB the 'debian' seems to be a link to the ubuntu dirs?), but I assume the distinction is there for a reason.

Also remember Travis testing is based on Ubuntu not Debian.

mstormi:
As commented we *should* support Ubuntu. Taking this forward in #926.

PR #690, opened 2019-08-27
[homegear] added missing Ubuntu Bionic Beaver

holgerfriedrich:
Just noticed that we missed current Ubuntu release when installing homegear.
May no longer be needed if we implement #621.


EliasGabrielsson:
Observant. 
We have not fully committed on the deprecation yet so lets have your fixes included.
I will do a full "deprecation PR" later this autumn when the pine64 support have been exported to the armbian project.

PR #700, opened 2019-09-02
build.bash: image mounting in docker

holgerfriedrich:
Hey @EliasGabrielsson, what about this workaround:

Though I got guestmount running in one of my docker containers, I decided to switch back to kpartx method of mounting the disk image when running inside docker.
Drawback is that inside docker build.bash needs to be run as root.
Plus is that kpartx needs way less dependencies (which may be installed on standard Linux systems but not in the minimal Docker image).

Note that you can drop libguestfs-tools from docker runs.

### DIFF
--- a/build.bash
+++ b/build.bash
@@ -4,6 +4,7 @@ set -e
 ##########################
 #### Load help method ####
 ##########################
+source $(dirname "$0")/functions/helpers.bash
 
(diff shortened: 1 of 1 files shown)


EliasGabrielsson:
Thanks Holger!

PR #701, opened 2019-09-02
Update README.md for Pine64 deprecation

mcowger:
PIne64 is no longer supported for 1.5+.


EliasGabrielsson:
Hi @mcowger!
We actually support Pine64 but recommend users to use Armbian as an updated base OS and manually use this tool. We will officially remove all support when we have exported essential part of this script to the Armbian config utility.  https://github.com/armbian/config

mstormi:
We'll fundamentally rework the README to state supported HW and OS in #730.
Thanks for your contribution.

PR #613, opened 2019-08-02
Shellcheck!

mstormi:
Ha, I bet you [didn't expect this one]((https://www.youtube.com/watch?v=HqF_nPbX_Ow) in time for 1.5, did you ?

Hope I didn't break anything but it was just a small number of non-trivial fixes so I guess you should be able to review/test these.

Signed-off-by: Markus Storm markus.storm@gmx.net

PR #705, opened 2019-09-05
build.bash: image mounting on RPi

holgerfriedrich:
Switched back image mounting also on RPi.  Needs root on RPi to build image.
Patch to support #613. Relates #700.


holgerfriedrich:
I think this is quite an uncommon use case. Could not get libguestfs running on Raspbian Buster properly and it introduces too many package dependencies. Will retry on a clean install.

PR #712, opened 2019-09-07
Fix audio group membership

### LINKED ISSUE #673: adding user openhab late in the install process makes us miss some steps ?
just a quick issue to not forget about checking this out
I just noticed running the test suite:

[...]
2019-08-24_08:18:12_UTC [openHABian] Enabling Audio output...
$ adduser openhab audio
adduser: The user `openhab' does not exist.

mstormi:
to fix #673

### DIFF
--- a/openhabian-setup.sh
+++ b/openhabian-setup.sh
@@ -80,7 +80,7 @@ if [[ -n "$UNATTENDED" ]]; then
   permissions_corrections
   misc_system_settings
 # not per default for now
-# if is_pi; then init_zram_mounts install; fi
+# if is_pione || is_pitwo || is_pithree || is_pithreeplus || is_pifour || is_pine64; then init_zram_mounts install; fi
   samba_setup
   clean_config_userpw
   frontail_setup
--- a/openhabian-setup.sh
+++ b/openhabian-setup.sh
@@ -62,6 +62,7 @@ if [[ -n "$UNATTENDED" ]]; then
   timezone_setting
   locale_setting
   hostname_change
+  if is_pi; then memory_split; enable_rpi_audio; fi
   if is_pine64; then pine64_platform_scripts; fi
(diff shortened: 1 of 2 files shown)


openhabian-setup.sh
```diff
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
+# shellcheck disable=SC2034
```
ThomDietrich: Can this be avoided?

openhabian-setup.sh
```diff
@@ -61,7 +62,6 @@ if [[ -n "$UNATTENDED" ]]; then
   timezone_setting
   locale_setting
   hostname_change
-  if is_pi; then memory_split; enable_rpi_audio; fi
```
ThomDietrich: Would leave memory_split up here. No real reason, just for the sake of not changing more than meaningful

openhabian-setup.sh
mstormi: yes but then either
 - we need it multiple times 
 - we export these variables which is what I did before. But Elias didn't like it.

openhabian-setup.sh
mstormi: That would mean to clutter functionality.
As memory_split() only writes into /boot/config.txt it does not really matter when it does that.

openhabian-setup.sh
EliasGabrielsson: In this case I would say a global `shellcheck disable=SC2034` is preferred to not clutter the code.

EliasGabrielsson:
LGTM

mstormi:
Agree. I've changed it to match your proposal.

mstormi:
any reason left why not to apply this?

BClark09:
Thanks @mstormi

PR #684, opened 2019-08-26
refactored build system

mstormi:
Fixes #559

Signed-off-by: Markus Storm <markus.storm@gmx.net>

PR #681, opened 2019-08-26
Fixes in preparation of new build system

mstormi:
Fixes: #639 

Signed-off-by: Markus Storm markus.storm@gmx.net

PR #724, opened 2019-09-22
Run apt-get update on startup in background

### LINKED ISSUE #810: Output in non-verbose unattended installation
Check where these come from...

```
2020-03-25_21:09:05_UTC [openHABian] Executing 'openhabian-setup.sh unattended'...
2020-03-25_21:09:05_UTC [openHABian] Checking for root privileges... OK
Hit:1 http://raspbian.raspberrypi.org/raspbian stretch InRelease
Hit:2 http://archive.raspberrypi.org/debian stretch InRelease
Reading package lists...
2020-03-25_21:09:27_UTC [openHABian] Loading configuration file '/etc/openhabian.conf'... OK

```

mstormi:
... and wait for it to finish after menu selection.

Fixes #810 

### DIFF
--- a/functions/influxdb+grafana.bash
+++ b/functions/influxdb+grafana.bash
@@ -243,7 +243,6 @@ influxdb_install() {
 #  if [ ! -f /etc/influxdb/influxdb.conf ]; then
     cond_redirect wget -O - https://repos.influxdata.com/influxdb.key | apt-key add - || FAILED=1
     echo "deb https://repos.influxdata.com/$dist $codename stable" > /etc/apt/sources.list.d/influxdb.list || FAILED=1
-    cond_redirect apt-get update || FAILED=1
     cond_redirect apt-get -y install influxdb || FAILED=1
 
     # disable authentication, to allow changes in existing installations
@@ -289,7 +288,6 @@ grafana_install(){
   echo -n "Installing Grafana... "
   cond_redirect wget -O - https://packages.grafana.com/gpg.key | apt-key add - || FAILED=2
   echo "deb https://packages.grafana.com/oss/deb stable main" > /etc/apt/sources.list.d/grafana.list || FAILED=2
-  cond_redirect apt-get update || FAILED=2
   cond_redirect apt-get -y install grafana || FAILED=2
 
   cond_redirect systemctl daemon-reload
--- a/functions/menu.bash
+++ b/functions/menu.bash
@@ -47,9 +47,11 @@ show_main_menu() {
     openhabian_update
 
   elif [[ "$choice" == "02"* ]]; then
+    wait_for_apt_to_finish_update
     system_upgrade
 
   elif [[ "$choice" == "03"* ]]; then
+    wait_for_apt_to_finish_update
     openhab2_setup
 
   elif [[ "$choice" == "10"* ]]; then
@@ -61,6 +63,7 @@ show_main_menu() {
     "15 | FireMotD"               "Upgrade the program behind the system overview on SSH login" \
     3>&1 1>&2 2>&3)
     if [ $? -eq 1 ] || [ $? -eq 255 ]; then return 0; fi
+    wait_for_apt_to_finish_update
     case "$choice2" in
       11\ *) basic_packages && needed_packages ;;
       12\ *) bashrc_copy && vimrc_copy && vim_openhab_syntax && nano_openhab_syntax && multitail_openhab_scheme ;;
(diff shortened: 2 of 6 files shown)


EliasGabrielsson:
Like the Idea of this patch Markus.
Will hopefully install a new system this weekend, will try to test it out then.

mstormi:
> Will hopefully install a new system this weekend, will try to test it out then.

Consider patching #681 - #684 in for that test only, too.

functions/influxdb+grafana.bash
```diff
@@ -218,7 +218,6 @@ influxdb_install() {
   influxdb_address="http://localhost:8086"
   influxdb_admin_username="admin"
   if [ ! -f /etc/influxdb/influxdb.conf ]; then
-    cond_redirect apt-get -y install apt-transport-https
```
ThomDietrich: Please move to independent PR, with rationale

openhabian-setup.sh
```diff
@@ -87,7 +87,8 @@ if [[ -n "$UNATTENDED" ]]; then
   clean_config_userpw
   frontail_setup
 else
-  apt-get update
+  apt-get -q update >/dev/null 2>&1 &
+  PID_APT=$!
```
ThomDietrich: Both in one line please

openhabian-setup.sh
```diff
@@ -59,7 +59,7 @@ OLDWD=$(pwd) && cd /opt
 if [[ -n "$UNATTENDED" ]]; then
   # apt/dpkg commands will not try interactive dialogs
   export DEBIAN_FRONTEND=noninteractive
-  apt-get update
+  cond_redirect apt-get update
```
ThomDietrich: Would also put in it's own PR. Unrelated

functions/openhab.bash
```diff
@@ -21,8 +21,6 @@ Check the \"openHAB Release Notes\" and the official announcements to learn abou
     TESTING=1
   fi
 
-  cond_redirect apt-get -y install apt-transport-https
```
ThomDietrich: Other PR

functions/menu.bash
```diff
@@ -47,10 +47,10 @@ show_main_menu() {
     openhabian_update
 
   elif [[ "$choice" == "02"* ]]; then
-    system_upgrade
+    wait ${PID_APT} 2>/dev/null; system_upgrade
```
ThomDietrich: Can we move this into a function with a speaking name, please? It's impossible to maintain this in the long run.

ThomDietrich:
Like this PR overall

functions/openhab.bash
mstormi: ok

functions/menu.bash
mstormi: ok now?

openhabian-setup.sh
```diff
@@ -59,7 +59,7 @@ OLDWD=$(pwd) && cd /opt
 if [[ -n "$UNATTENDED" ]]; then
   # apt/dpkg commands will not try interactive dialogs
   export DEBIAN_FRONTEND=noninteractive
-  apt-get update
+  apt_update
```
ThomDietrich: Do we need to wait for it to finish here?

mstormi: not here but before calling apt-get next time which is in timezone routine

functions/openhabian.bash
```diff
@@ -9,6 +9,16 @@ get_git_revision() {
...
 }
 
+apt_update() {
+  echo -n "$(timestamp) [openHABian] Updating Linux package information ... "
+  apt-get -q update >/dev/null 2>&1 &
+  PID_APT=$!
+}
+
+wait_for_apt_to_finish_update() {
+    wait ${PID_APT} 2>/dev/null
```
holgerfriedrich, line 19: Pls make sure that PID_APT is set, otherwise call apt_update. Then we can still source the function files directly from shell without changing the experience (wait w/o a pid will just wait for child tasks, if there are none, it will continue).

holgerfriedrich: did you think of using `wait -f``?

holgerfriedrich: Might be a bit strange for users: a) it might wait for long without showing a reason b) on connection errors we are hiding the reason....

mstormi: ok now check for PID_APT and used wait -f
On your last point I would leave it like that. IIRC it does not work without that redirection so that's the lesser pain point (and only when Inet is down).

functions/openhabian.bash
```diff
@@ -9,6 +9,20 @@ get_git_revision() {
...
+  echo -n "$(timestamp) [openHABian] Updating Linux package information ... "
+  apt-get -q update >/dev/null 2>&1 &
+  PID_APT=$!
+}
+
+wait_for_apt_to_finish_update() {
+  if [ -v PID_APT ]; then
+    wait -f ${PID_APT} 2>/dev/null
+  else
+    apt_update
```
holgerfriedrich, line 22: ...and wait for completion
so the better approach would be
`if not ... then apt_update; fi
wait -f ...`

mstormi: right

Issue #732, opened 2019-10-07
Openhabian-config v1.5-522 will not let you change hostname


mljbr4:
openhabian-config v1.5-520 works: there is a line in file /opt/openhabian/functions/system.bash which reads:

`new_hostname=$(whiptail --title "Change Hostname" --inputbox "Please enter the new system hostname (no special characters, no spaces):" 10 60 3>&1 1>&2 2>&3)
if [ $? -ne 0 ]; then return 1; fi
if ( echo "$new_hostname" | grep -q ' ' ) || [ -z "$new_hostname" ]; then
  whiptail --title "Change Hostname" --msgbox "The hostname you've entered is not a valid hostname. Please try again." 10 60`

but in openhabian-config v1.5-522 it reads:

`if ! whiptail --title "Change Hostname" --inputbox "Please enter the new system hostname (no special characters, no spaces):" 10 60 3>&1 1>&2 2>&3; then return 1; fi
if ( echo "$new_hostname" | grep -q ' ' ) || [ -z "$new_hostname" ]; then
  whiptail --title "Change Hostname" --msgbox "The hostname you've entered is not a valid hostname. Please try again." 10 60`

So in v1.5-522 the script refers to the variable $new_hostname which has not been set to anything and hence is treated as an invalid hostname in the validation.

PR #733, opened 2019-10-07
Fix Openhabian-config unable to change hostname

mljbr4:
#732 
Revert code to that found in openhabian-config v1.5-520


mstormi:
Reverting to old code is not a proper fix as the change - albeit bad - was intentional  Can you please change your PR to work with shellcheck?
Use it on the old and new script and you'll see what I mean. I'd try to re-add the new_hostname=... before whiptail but make sure it functions properly when you do.

EliasGabrielsson:
Let's rework this once more 🎡

mstormi:
I had already mentioned what to change...ok so here's #741 (unit tested on my dev box)

PR #761, opened 2019-11-06
use $HOSTNAME in webif for install log

mstormi:
instead of hostname from openhabian.conf as it is the client that needs to resolve it

### DIFF
--- a/build-image/webif.bash
+++ b/build-image/webif.bash
@@ -25,7 +25,7 @@ if [ "$1" = "start" ]; then
   echo '<body>
         <h1>openHABian Installation Status</h1>
         the log will be refreshed automatically every 10 seconds
-        <iframe src='"http://${hostname:-openhab}:$port/first-boot.txt"' scrolling="yes" width="100%" height="90%"></iframe>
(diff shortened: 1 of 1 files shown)


ThomDietrich:
Hey Markus!

> as it is the client that needs to resolve it

What does that mean?

Also, could you please explain why for the webinterface we should use the system hostname rather than the configuration one? I don't think it's wrong or anything but we should try to keep consistency and the rest of the script uses $hostname as far as I'm aware (might not be correct anymore).

Best!

mstormi:
> > as it is the client that needs to resolve it
> What does that mean?

The openHABian box-under-installation is the server and the user's desktop PC is the client.
While it's openHABian to output that URL to contain the hostname, that name will not be resolved on openhabian but instead on the client. 
$hostname, if not changed via menu, is a fixed string ultimately taken from openhabian.conf while $HOSTNAME is what the box thinks it's named (dynamically set on boot/DHCP) so for *this* use I think it's the better choice. 
In most installations, server and client use a common DNS resolver and the DHCP server allows for the client to define the hostname so "openhab" (or whatever the user changed it to in openhabian.conf) is used on both sides and it works to access the URL.
But in more restricted setups this may be different and both systems can end up to use different names.

PR #762, opened 2019-11-06
cd on cloning openhabian so script does not break when working dir is removed while we're inside

mstormi:


ThomDietrich:
cd should not be used inside scripts. Why is this step suddenly necessary? Which other change introduce the bug?

mstormi:
> cd should not be used inside scripts
Why? That's not necessarily true.

> Why is this step suddenly necessary?
read the PR title :)

> Which other change introduce the bug?
Dunno. I believe it has been in ever since but didn't show up as a problem until we started work on automating testing with Travis.

mstormi:
reworked so it's to work with all current+future instances of deleting dirs

Issue #763, opened 2019-11-14
don't put backups into zram


mstormi:
On ZRAM activation, set ${OPENHAB_BACKUPS} in /etc/defaults/openhab2 to something other than the default so backups don't consume valuable zram space
Inform users about changed backup location (any good idea where to place them instead?)

ethyr2000:
Is this a backup or an archive?  If it is a backup of settings, then the ideal would be to have it write to the base image that loads into zram.

mstormi:
I don't understand your point. It's a archive.
And it's what's live on disk, so whenever you shutdown zram, it's synced to disk and loaded on next start. It's no "image" BTW but an OverlayFS.

shutterfreak:
@mstormi How about a folder in /var/local/openhab or /opt/openhab ?

mstormi:
@BClark09 would you mind changing the default backup location to /var/local/openhab-backup as
suggested ?
I would rather not like to start using different locations depending on whether ZRAM is on or off.

BClark09:
`/var/lib/openhab/backup` was chosen because openHAB and the script that creates the backup it is installed through a package.

> I would rather not like to start using different locations depending on whether ZRAM is on or off.

But using a different location depending on if ZRAM is _installed_ should be simple, no?

mstormi:
> But using a different location depending on if ZRAM is _installed_ should be simple, no?

Sure (that's what I, too, meant by "on"/"off" btw).
But I dislike the idea of having different locations for ZRAM users than others, will lead to confusion. Hence my proposal to move. ...?

mstormi:
>  We could turn `/var/lib/openhab2/backup` into a link to the new location, wdyt?

@BClark09 ?

BClark09:
I'm not sure about how I feel about changing the default location for all apt/yum users to solve an issue with ZRAM for openHABian users. Won't that have the same issue as giving different locations for different users depending on what version they're using, but with even more people?

I'm not completely against changing it, but if we do then it'd probably have to come with an announcement.

mstormi:
I'd suggest to, on openhab2 package install, turn the old location /var/lib/openhab2/backup into a link to the new location so noone has to change (people can keep using the location they're used to) and everybody new to install oh will get the new location.
Sure you can have an announcement come along with that change, too.

holgerfriedrich:
..,.or the other way around: leave it as it is for package install, existing and new installations.
[...]

PR #764, opened 2019-11-15
Enable ZRAM per default for logs and persistence; enable delayed rules

mstormi:
ZRAM has proven to work reliably for several months.
I think it's time to make it the default now so new openHABian users will also benefit from it.
Today, most new users are not aware of it or do not dare to 'manually' change any setting.
Note this PR will only enable this on new, unattended installs but will not change existing installations so the potential impact is rather acceptably low.
Plus the CI to contain unit and integration tests for ZRAM is in place now (and green ;)). I also had to adapt Travis testing for ZRAM.

Second, I also enabled the 'delayed rules load' per default now. Delayed rules were long in as an option and have proven to work, too.

On the documentation part, initially this patch was only to rework Availability/Backup _documentation_ section of openhabian.md to recommend enabling ZRAM and delayed rules.
Now also recommending to install Amanda (as that cannot be automated to work w/o user input).

Fixes: #763 


ecdye:
Please Rebase.

docs/openhabian.md
```diff
@@ -77,9 +77,9 @@ Learn more about the Raspberry Pi as your platform for openHAB and about the req
 
 **Setup:**
 
-- [Download the latest "openHABian" SD card image file](https://github.com/openhab/openhabian/releases) (Note: the file is *xz* compressed)
-- Write the image to your SD card (e.g. with [Etcher](https://www.balena.io/etcher/), able to directly work with *xz* files)
-- Insert the SD card into the Raspberry Pi, connect Ethernet ([Wi-Fi supported](#wi-fi-based-setup-notes)) and power
+- [Download the latest "openHABianPi" SD card image file](https://github.com/openhab/openhabian/releases) (Note: the file is *xz* compressed)
```
ecdye, line 80: Re-base artifacts?

docs/openhabian.md
```diff
@@ -77,9 +77,9 @@ Learn more about the Raspberry Pi as your platform for openHAB and about the req
...
+- Write the image to your SD card (e.g. with [Etcher](https://etcher.io), able to directly work with *xz* files)
```
ecdye, line 81: more, re-base artifacts.

docs/openhabian.md
```diff
@@ -163,7 +163,7 @@ The configuration tool is the heart of openHABian.
 It is not only a menu with a set of options, it's also used in a special unattended mode inside the ready to use images.
 
 ⌨ - A quick note on menu navigation.
-Use the cursor keys to navigate, <kbd>Enter</kbd> to execute, <kbd>Space</kbd> to select and <kbd>Tab</kbd> to jump to the actions on the bottom of the screen. Press <kbd>Esc</kbd> twice to exit the configuration tool.
+Use the cursor keys to navigate, &lt;Enter&gt; to execute, &lt;Space&gt; to select and &lt;Tab&gt; to jump to the actions on the bottom of the screen. Press &lt;Esc&gt; twice to exit the configuration tool.
```
ecdye, line 166: Again?

docs/openhabian.md
```diff
@@ -275,7 +283,7 @@ During and after the first boot of your Raspberry Pi, the green on-board LED wil
 -->
 
 **RPi note:**
-The progress indication via the **green Raspberry Pi LED** is currently not possible and hence not part of the openHABian v1.3+ image.
+The progress indication via the **green Raspberry Pi LED** is currently not possible and hence not part of the openHABian v1.3 image.
```
ecdye, line 286: Same?

BClark09:
The ZRAM implementation is really good work, but I have a reservation about enabling ZRAM as a default option currently, I'm happy to be outvoted by @openhab/openhabian-maintainers though.

I think I've mentioned before that I dont think including the entirety of `/var/lib/openhab2` makes sense for many users. What happens if you save a complicated binding configuration through the UI and then you lose power, what happens to your persistance enabled items or historical data?

mstormi:
Well. now that I know a little more about the inner workings of ZRAM, I think I could replace /var/lib/openhab2 by a number of its subdirs (jsondb,persistence, eventually cache+tmp or leave those out). That would also solve the issue of .../backups.

Then again I guess that wouldn't convince you either because even with that change, you could raise the same questions. 
In the end, you cannot have both, SD mitigation AND protection from power loss at the same time.
You need to complement that with a backup, that's why there's also a reenforced recommendation in this PR to install Amanda.

BClark09:
Hi @mstormi. Sorry, I didn't mean to sound as opposed to it as my message probably sounds! I trust your judgement on ZRAMs use much more than mine! I just wanted to mention my reservations for the point of discussion, as I think it's important to discuss changes that effect any new user of openHABian, especially if it has the potential to behave slightly differently than to what you'd expect on other systems. My reasons for bringing a counter-argument were indeed to be objective, to weigh the pros and cons.

If no one else in @openhab/openhabian-maintainers has anying to add, I think you should keep this PR how you originally proposed if you think that's best, because that's how it's been working for ZRAM testers with success.

mstormi:
>  I think it's important to discuss changes that effect any new user of openHABian,

You're off course right there. Unfortunately Elias and Thomas are pretty unresponsive, and I've been waiting very long already for them to review all my PR throughout the last 3/4 of a year - that didn't happen until Holger joined. So I fear I'll be waiting till Christmas to get that discussion done ....

> I think you should keep this PR how you originally proposed (instead of just including persistance and jsondb)

Well you (unintentionally) nudged me into that direction ... but I wouldn't do it if I didn't think it's quite an improvement. Just like you I wanted to do it like that from the very beginning so now when this reminded me of the fact that I'm _now_ sure we're _not_ wasting memory (which I was not in the beginning) I finally did.
ZRAM'ing a subset / a lesser part of the filesystem should be as reliable as to do it for all of `/var/lib/openhab2` so I wouldn't think we need a special retest phase here.

functions/openhab.bash
```diff
@@ -63,11 +78,13 @@ Check the \"openHAB Release Notes\" and the official announcements to learn abou
   if ! cond_redirect apt-get ${APT_INST_OPTS} install "openhab2=${openhabVersion}"; then echo "FAILED (apt)"; exit 1; fi
   cond_redirect adduser openhab gpio
   cond_redirect systemctl daemon-reload
+  cp "${BASEDIR}"/includes/start-level.script "${OPENHAB_USERDATA}"/etc/
```
holgerfriedrich, line 81: this is not active, right?
It could be implemented as an alternative approach compared to the renaming you added via systemd.
But I think it is not intended to be touched, and the different bundle names seem to change slightly between different OH releases.
So do we need this here?

functions/openhab.bash
```diff
@@ -63,11 +78,13 @@ Check the \"openHAB Release Notes\" and the official announcements to learn abou
...
   if cond_redirect systemctl enable --now openhab2; then echo "OK"; else echo "FAILED (usr)"; exit 1; fi
 
   if is_pi || is_pine64; then
-    cond_echo "Optimizing Java to run on low memory single board computers... "
-    if is_pizerow || is_pione ; then
+    echo -n "$(timestamp) [openHABian] Optimizing Java to run on low memory single board computers... "
+    totalmemory=$(grep MemTotal /proc/meminfo |awk '{print $2}')
```
holgerfriedrich, line 86: seems to be time for a helper function.... ;-)
if we reuse it again, we should do so...

includes/start-level.script
```diff
@@ -0,0 +1,13 @@
+bundle:start-level org.openhab.core.model.thing 83
```
holgerfriedrich, line 1: see comment above

docs/openhabian.md
```diff
@@ -213,18 +207,24 @@ All password can be changed from openHABian menu.
...
 
-1. Use openHAB integrated [backup tool](https://community.openhab.org/t/recommended-way-to-backup-restore-oh2-configurations-and-things/7193/82?u=elias_gabrielsson).
-2. Move the root filesystem to an external USB-memory. Warning: USB stick are as-susceptible to flash wearout as SD cards. [Menu option:37]
-3. (BETA) Reduce wear on SD card by moving write intensive actions temporary to RAM during operation (logs,persistant-data). Warning: power failure will result in lost data. [Menu option: 6A]
-4. (Advanced) Use [Amanda Network Backup](http://www.amanda.org/) for full system backup, longer introduction [here](https://github.com/openhab/openhabian/blob/master/docs/openhabian-amanda.md). [Menu option: 51]
+Standard openHABian install enables ZRAM (#1) by default. Once installed, you can disable ZRAM and choose to install to a safe external medium such as an SSD (#2) instead via menu options in menu 30.
+Delayed rules load (#3) will also be enabled by default. You can toggle to use this feature in menu option 44.
```
holgerfriedrich, line 221: this seems no longer the case as you removed it yesterday

docs/openhabian.md
```diff
@@ -191,17 +183,19 @@ These are just some first hints.
...
 
 {: #further-config}
-### Further Configuration Steps
+### Recommended Configuration Steps
 
-openHABian is supposed to provide a ready-to-use openHAB base system. There are however a few things we can not decide for you.
+openHABian is supposed to provide a ready-to-use openHAB base system.
+There are a few things, however, we need you to decide and act on right now at the beginning:
 
+- **Delayed Rules Loading** openHAB startup times can be annoyingly long. There's an optimization available that *delays* loading the rules, *effectively speeding up* openHAB startup. We recommend to enable this now. See [menu option: 44].
```
holgerfriedrich, line 191: Nothing specific against delayed rules, but users should be aware of this.
What happens if you keep you rules open via am Samba share or fuse mount?
Maybe a small hint in the docs that this feature actually _moves_ files and rules will only be active 2 minutes after start....
Maybe it is already in the docs, I look only at the patch.

Dockerfile.openhabian-aarch64-64bit
```diff
@@ -9,7 +9,7 @@ ENV INITSYSTEM on
 ENV APT_KEY_DONT_WARN_ON_DANGEROUS_USAGE=DontWarn
 
 RUN apt-get update -qq \
-    && apt-get install -y -qq --no-install-recommends systemd systemd-sysv git wget python python-pip python3 python3-pip apt-utils jq virt-what \
+    && apt-get install -y -qq --no-install-recommends systemd systemd-sysv git wget python python-pip python3 python3-pip apt-utils jq virt-what gcc \
```
holgerfriedrich, line 12: 1) Any hint why we need gcc here? Difference between raspbian and balena images?
2) Where do we have a feature with dependency on a compiler -> maybe better add the install there.
In case someone is running openhab on a Debian minimal install there might be no gcc installed.

openhabian-setup.sh
```diff
@@ -92,11 +92,10 @@ if [[ -n "$UNATTENDED" ]]; then
   srv_bind_mounts
   permissions_corrections
   misc_system_settings
-# not per default for now
-# if is_pione || is_pitwo || is_pithree || is_pithreeplus || is_pifour || is_pine64; then init_zram_mounts install; fi
   samba_setup
   clean_config_userpw
   frontail_setup
+  zram_setup
```
holgerfriedrich, line 98: I do not think it is good to do this for default installation.

holgerfriedrich:
@mstormi regarding the directories to put into ZRAM: did you collect statistics on the data writes? Looking at my installation `/var/log/openhab2/` seems to be quite actively used.

mstormi:
> @mstormi regarding the directories to put into ZRAM: did you collect statistics on the data writes? Looking at my installation `/var/log/openhab2/` seems to be quite actively used and I would also suspect `/var/lib/influxdb` (which is also outside if the openhab folder).

Yes, long ago. It's logs and persistence to account for the majority of writes, hence my choice of directories in /etc/ztab. This is already an optimized list to not zram more than useful..

docs/openhabian.md
mstormi: there's no extra docs so I added a note here

docs/openhabian.md
mstormi: no I re-added delayed_rules to openhab2_setup()

functions/openhab.bash
mstormi: done

functions/openhab.bash
mstormi: (this is a lengthy discussion you weren't involved in, please don't act on fractional knowledge)
Yes it *is* active. Yes it's an alternative approach but not comprehensive that's why I keep the delayed rules function active, too. They complement each other to some extent.
It is not intended to be touched _by users_ but for us it's the right place/way to enforce that.
No names don't change between releases (... any more)
No it does not do harm even if names should change again some day.

docs/openhabian.md
```diff
@@ -275,7 +274,7 @@ During and after the first boot of your Raspberry Pi, the green on-board LED wil
 -->
 
 **RPi note:**
-The progress indication via the **green Raspberry Pi LED** is currently not possible and hence not part of the openHABian v1.3+ image.
+The progress indication via the **green Raspberry Pi LED** is currently not possible.
```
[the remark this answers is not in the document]
mstormi: no there's noticket, I have never seen this function either (it's one of those legacy things Thomas added to the docs long ago and noone dares throwing out things when you have not seen them before)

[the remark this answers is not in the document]
ecdye: Same here, never seen it, on the other hand I doubt it is going to change any time soon. I don't think it would hurt to remove it. On the other hand it is not hurting us to leave it there.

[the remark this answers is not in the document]
holgerfriedrich: @mstormi you have my support for removing it. I have not seen the code which tries to control the LEDs...

functions/openhab.bash
holgerfriedrich: I'd like to understand it when I to a review. 
1) Using the script for startup levels: I tried to do this on my installation, copied over the file to `/var/lib/openhab2/etc`, restarted OH. But `bundle:list` did not show any difference after restart. It should, right? This is why I asked if this is active.
2) Ok, if we do both, then please add a comment. For me it looked like two independent approaches for the same problem.

mstormi: It worked on my test box. You must place it in `/var/lib/openhab2/etc`**scripts/**

docs/openhabian.md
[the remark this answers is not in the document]
mstormi: ok, I removed the docs on indicator LED

mstormi:
Ben, Holger:
I've updated the code to only zram `/var/log` and `/var/lib/openhab2/persistence`.
So no essential data to be affected (see above).
WDYT _now_ ?

docs/openhabian.md
holgerfriedrich: comment in the code is good, do you think we should also state the second part here in the docs?
"quickly rename rules forth and back after 2 minutes"

functions/openhab.bash
```diff
@@ -63,11 +79,12 @@ Check the \"openHAB Release Notes\" and the official announcements to learn abou
   if ! cond_redirect apt-get ${APT_INST_OPTS} install "openhab2=${openhabVersion}"; then echo "FAILED (apt)"; exit 1; fi
   cond_redirect adduser openhab gpio
   cond_redirect systemctl daemon-reload
+  cp "${BASEDIR}"/includes/start-level.script "${OPENHAB_USERDATA}"/etc/scripts/
```
holgerfriedrich, line 82: should this be standard for all installs or part of the delayed rules option?

mstormi: I thought about that beforehand and think it should be part of any install hence I've put it here.

functions/openhab.bash
```diff
@@ -1,5 +1,21 @@
 #!/usr/bin/env bash
 
+# two overlapping methods to enforce proper load order:
+# /var/lib/openhab2/etc/scripts/start-level.script to provide default start levels
```
holgerfriedrich, line 4: It seems a bit strange that this is _not_ done when you enable delayed rules, but only when you install a new openhabian version.
So we don't have is in existing installations, even if user selects  and also not in first_boot.
Maybe add the copy of start-level script also here if it does not exist...

holgerfriedrich:
Hey @openhab/openhabian-maintainers !
Finally we are back to the question how we want to introduce ZRAM feature.
@mstormi has put a lot of work there and wants to see this in use, of course.

I think most of us see a risk that this increases the complexity of openHABian, and this is why we have some concerns about introducing this as a default feature. Just remember the goal of openHABian to provide easy installs of openHAB!
Markus tried to reduce the risk of loosing data by cutting down the number of cached directories to `/var/log` and  `var/lib/openhab2/persistence/`. He proposes to enable ZRAM by default for new installs now.

WDYT?

functions/openhab.bash
mstormi: yep. I moved it to delayed_rules().
Now if we wanted to also enable it on existing installs that would require another call but my intention was to have it for new installs only.

includes/start-level.script
BClark09, line 1: This shoudn't be specific to openHABian, nor should we be changing the behaviour of openHAB compared to other instalation types. 

If this improves the startup of openHAB then it should be suggested as a change in openHAB-distro or openHAB-core instead.

mstormi: Yep. That's what I already did. But Kai and Wouter opted against it so I've removed it here again.
Kai left me with some hope there's going to be a solution mid-term.
https://github.com/openhab/openhab-core/issues/1467#issuecomment-626358060

ecdye:
FWIW, I think that we should enable zram by default. I have had zero issues with it and think it is an excellent addition.

mstormi:
@openhab/openhabian-maintainers I'm really getting tired of this - please respond latest by this weekend else I will assume you have no interest in this. I will merge this PR by then if you don't veto.


> Finally we are back to the question how we want to introduce ZRAM feature.

PR #770, opened 2019-11-24
fix for multiple nginx users .htpasswd

### LINKED ISSUE #769: Fix change passwords menu
Authentication is broken (menu optics, too) if there's multiple nginx users in .htpasswd:
```+ '[' -f /etc/nginx/.htpasswd ']'
++ cut -d: -f1 /etc/nginx/.htpasswd
+ nginxuser='openhab
markus'
```

mstormi:
nginx .htpasswd screws up functionality if it contains more than 1 entry
Fixes #769 

8 chars are way more common, so don't require users to enter 10 char passwords
(due to this many people need to think of a new one)

### DIFF
--- a/functions/passwords.bash
+++ b/functions/passwords.bash
@@ -51,7 +51,7 @@ change_password() {
         passwordChange="$(whiptail --title "Authentication Setup" --passwordbox "Enter a new password: " 15 80 3>&1 1>&2 2>&3)"
         if [[ "$?" == 1 ]]; then return 0; fi
         if [ ! ${#passwordChange} -ge 10 ]; then
-          whiptail --title "Authentication Setup" --msgbox "Password must at least be 8 characters long... Please try again!" 15 80 3>&1 1>&2 2>&3
+          whiptail --title "Authentication Setup" --msgbox "Password must at least be 10 characters long... Please try again!" 15 80 3>&1 1>&2 2>&3
         else
           secondpasswordChange="$(whiptail --title "Authentication Setup" --passwordbox "Please confirm the new password:" 15 80 3>&1 1>&2 2>&3)"
           if [[ "$?" == 1 ]]; then return 0; fi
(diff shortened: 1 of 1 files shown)


EliasGabrielsson:
If I remember correctly one of the services needed 10 characters and that limit have been applied on all services in order to have maintainable and clean code.

ThomDietrich:
I can't remember the reason for 10 but I'd like to refer to Randall Munroe: https://xkcd.com/936/
I wouldn't really mind but changing the rule just because "more common" wouldn't cut it for me. This is a public interface through which an attacker can cause real world harm. We should educate our users about that and raise the minimum password length to 16.

mstormi:
Well.
First I've removed the pw length change so you can apply the fix without that we need to finish the password discussion.
Second, on passwords. 
I'm totally reluctant of any app that forces my password to be of some sort or another
(minimum length 8 being the exception as that's widely accepted as a requirement). 
Main reason being I cannot reuse passwords. Mine, albeit only 8 in length, are well more than 28 bits of entropy as I don't start with a dictionary word, and they certainly are safer than a 10 char d1ct10n4ry word is.
Not being able to reuse passwords (as well as all those dumb company rules requiring you to change yours every xx days) result in people forgetting about it, resulting in them trying all sorts of passwords they use on all sorts of apps. If I was to collect passwords, as a server operator I'd just log failed inputs - would get me fair amount of password to work on other sites.
Security is everyone's own, personal duty. It is incredibly tiresome (and certainly not our job) to educate users on that.

ThomDietrich:
Hey Markus,
your thoughts on passwords are definitely worth discussing. Please open a new issue on it and please read the link I've provided. Also feel free to do some further reading about short passwords with high entropy and reusing passwords :)

Staying with 10 or reducing to 8 is basically fine with me. @EliasGabrielsson could you check the case you mentioned?

EliasGabrielsson:
I will need to test each service regarding the password length.
Will look into it if we find password length an issue.

PR #774, opened 2019-12-04
fix influx setup problems

RoadRunnr:


EliasGabrielsson:
Thanks for your contributions!
It looked good so I merged it with out further investigation.
Will make a test installation some upcoming day. 👍

Issue #790, opened 2020-01-22
openhabian-config suggesting to install embedded mqtt broker?


JueBag:
When selecting the option 23 in openhabian-config the user gets displayed following note:
```
Do you really want to install a separate MQTT broker? │
│ Note that openHAB version 2.5M1 or higher comes with a new OHv2 MQTT │
│ binding that if you choose to enable it by default will use thus occupy │
│ the same ports as mosquitto here would try to use.
```
IMHO the MQTT binding version 2.5 does by itself not come with a broker, instead the broker Moquette can be installed. Several post in the community suggest that this one isn't maintained anymore, hence most of the times the use of the Mosquitto MQTT broker is suggested.

IMHO the note posted is not inline with the current suggestions, should the note be changed or the suggestions on the forum?

mstormi:
I think you're right, but I don't use it so ultimately I'm not sure. Can you confirm OH mqtt2 binding to work with mosquitto on the same box (no port clashing?) If yes, please create a PR here.

JueBag:
Yes. My setup is a Raspi3 running OH2.5 and Mosquitto (I did set it up manually before OH2.4, i.e. for the MQTT binding version 1). 
Doing a PR I have a problem. I would have suggested to remove that Note and found it incooperated as "Question1",  never worked with that (is this a standard feature of those scripts?). So I would not know what I'm doing. Sorry for that.

mstormi:
>  Doing a PR I have a problem. I would have suggested to remove that Note and found it incooperated as "Question1", never worked with that (is this a standard feature of those scripts?). So I would not know what I'm doing. Sorry for that.

That's simple bash programm. Text is assigned to a variable, command further down the line
(' whiptail ... ') makes use of it.
Take your time. Try editing your local install (where the "framework" is already working).
You can also interactively copy lines to shell to execute them.  It's not hard, you'll see once you started.

JueBag:
👍I'll do my very best!

JueBag:
Done it. [PR](https://github.com/openhab/openhabian/pull/791)
Although I have problems again with the signature, which I do not understand. I edited the code on the browser, having the signature inserted both under the code change and on the PR.
I can't do (at least I think so) the suggested `git commit --amend --signoff` since I do not have a local copy (because I just edited in the browser).

mstormi:
You successfully created your first PR. I just merged it.
Thank you for your contribution.

PR #791, opened 2020-01-24
[MQTT Setup] Removed Note concerning embedded MQTT broker

JueBag:
When selecting to install the Mosquitto MQTT broker via openhabian-config actually the user is asked if (s)he does really want to install a seperate MQTT broker. IMHO users that reachthis point are sure about that.
A (thinkable) problem would arise only if two MQTT brokers would be installed, however such wil not be prevented by this note.

Signed-off-by: Juergen Baginski opus42@gmx.de


mstormi:
Fixes #790

PR #798, opened 2020-02-23
use dumpcycle = runspercycle = 14 to avoid frequent level 0 dumps

mstormi:


mstormi:
fixes #801

PR #807, opened 2020-03-24
replace fstab by explicit systemd mount units to ensure proper order

### LINKED ISSUE #795: /srv needs to be mounted AFTER ZRAM starts
else it's pointing to the lower FS (original directories on SD that do not change as all changes are stored in the upper FS in ZRAM only)

mstormi:
... to make ZRAM work with Samba shares of /srv/openhab2-*

Note this effectively removes most entries from /etc/fstab but that's intentional else we would not get the intended order of mounts to work.

Use option 13 to validate.

Fixes #795 

### DIFF
--- a/functions/system.bash
+++ b/functions/system.bash
@@ -134,23 +134,33 @@ vimrc_copy() {
   echo "OK"
 }
 
+create_mount() {
+  F=/etc/systemd/system/$(systemd-escape --path /srv/openhab2-"${2}").mount
+  sed -e "s|%SRC|$1|g" -e "s|%DEST|$2|g" "$BASEDIR"/includes/mount_template > "$F"
+  systemctl -q enable "srv-openhab2\\x2d$2.mount"
+  systemctl -q start "srv-openhab2\\x2d$2.mount"
+}
+
 srv_bind_mounts() {
   echo -n "$(timestamp) [openHABian] Preparing openHAB folder mounts under /srv/... "
+  systemctl is-active --quiet smbd && systemctl stop smbd
+  systemctl is-active --quiet zram-config && systemctl stop zram-config
+  cond_redirect umount -q /srv/openhab2-{sys,conf,userdata,logs,addons}
   sed -i "\\#[ \\t]/srv/openhab2-#d" /etc/fstab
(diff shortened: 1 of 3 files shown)


mstormi:
we should have this before default ZRAM (#764)

mstormi:
@holgerfriedrich @BClark09 anyone care to quickly give this a review ? (or @ecdye?)
I'd like to have this in before activating ZRAM.
(just apply and reboot your box to see it work)

mstormi:
On a sidenote, does anyone know where openHABian/Raspbian mounts /tmp as tmpfs ?
Not is fstab....

mstormi:
> but the openHAB-share share won't load any of its subfolders

by "the openHAB-share share" you mean the first one from smb.conf, i.e. /srv as quoted below ?
Ok now that possibly is because it is exported before ZRAM,  that's what happened with all the /srv/* before my patch).
So you'll export/access the lower FS (which is the never changing part in the OverlayFS that ZRAM sets up, all changes go to the upper FS).

I didn't create the Samba stuff and don't know what the person to share that had in mind so wondering how to cope with that... easiest is to leave that as-is or switch that share off, I don't know what it is good for anyway. If we wanted to share more than the confdir (/etc/openhab2) we could add more share (they're already there just commented out for the time being).

#=================== Custom Share Definitions ====================
```

[openHAB-share]
  comment=openHAB2 combined folders
  path=/srv
  writeable=yes
  public=no
  create mask=0664
  directory mask=0775
  veto files = /Thumbs.db/.DS_Store/._.DS_Store/.apdisk/._*/
  delete veto files = yes

```

ecdye:
Yes that is what I was talking about. It is good for accessing the other openHAB related items that may need to be edited. I don't think we should get rid of it but is up to you.

mstormi:
what other ones? srv/openhab2-conf is /etc/openhab, everything an ordinary user should access is below isn't it ? Logs are accessible via Frontail.

And they can always logon to see more.

mstormi:
see changed includes/smb.conf

Is that now a reasonable set to share ? @holgerfriedrich ?

ecdye:
True, I'm ok with it.

PR #812, opened 2020-03-25
build.bash libguesfs changes required for running on Debian10

holgerfriedrich:
Debian10 libguestfs is a bit picky about two active mounts in one image due to qemu bug.
Now separating the mounts.


mstormi:
Cannot really comment on the contents but I'm sure you know that part of the code better than anyone else and did the unit testing.
Other than that, LGTM.

PR #819, opened 2020-03-31
fix Java install/upgrade

### LINKED ISSUE #723: current Java gets installed again and openHAB restarted on every system upgrade from openhabian-config
See https://community.openhab.org/t/openhabian-config-zulu-openjdk-forces-openhab-to-restart/82173/2

### LINKED ISSUE #729: Not installing 64bit Zulu although selected
see https://community.openhab.org/t/openhabian-and-zulu-64-bit/82636
This is on x86.
But independent of the arch, when a user instructs OH to install 64bit, we should do. Or omit that choice.

I suggest we change install to only make it 32 bit if on ARM and 64 if on x86 HW. We should NOT offer 64bit for ARM as that has no advantages but many drawbacks.
I think that at the same time to restrict x86 install to 64 only (as it simplifies things and AFAIK has no disadvantages on x86 boxes which have enough memory).

### LINKED ISSUE #805: [raspberry] openhabian-config symlink aarch32 w/ arm
Hi,
I have a raspberry with unifi and openhab running. Unifi requires Java 1.8 too, but original oracle or the open-jdk. With some testing I got unifi also running with the zulu java installed by openhabian-config. It runs since 2 month without any issues.
Unifi uses jsvc to start the java processes. Jsvc checks on unusual places for java libraries. With symlinking the 2 "aarch32" directories to "arm" within the zulu jdk jsvc is now happy with the zulu jdk.
Is it possible to add the 2 symlinks in the install process of the jdk?

### LINKED ISSUE #839: Installation of Zulu 64 bit fails (at least on aarch64)

mstormi:
remove calling java (re)install on Option 03 (Upgrade) to avoid Java reinstall of same version
fix installing 64bit where applicable (x86) or explicitly requested

Fixes #723
Fixes #729
Fixes #805  
Fixes #839

I have been waiting so long before publishing this PR because I would have preferred Elias to check for installed java version in his install routine as I don't want to touch that code, but now that he's offline I think it's reasonable to implement it.

### DIFF
--- a/functions/system.bash
+++ b/functions/system.bash
@@ -9,17 +9,7 @@ whiptail_check() {
 
 system_upgrade() {
   echo -n "$(timestamp) [openHABian] Updating repositories and upgrading installed packages... "
-  # TODO: Remove updating the key on each update after some time (in 2020)
-  cond_redirect wget -O openhab-key.asc 'https://bintray.com/user/downloadSubjectPublicKey?username=openhab'
-  if ! cond_redirect apt-key add openhab-key.asc; then
-    echo "FAILED (updating OH key)"
-    echo -n "attempting to continue..."
-  else
-    rm -f openhab-key.asc
-  fi
   cond_redirect apt-get --yes upgrade
-  # shellcheck disable=SC2154
-  if cond_redirect java_install_or_update "$java_arch"; then echo "OK"; else echo "FAILED"; exit 1; fi
 }
 
 basic_packages() {


mstormi:
@holgerfriedrich mind to finally review this so I can merge?
There's still people to be affected on Java being replaced/OH restarted on selecting option 02

We could go straight for #848 but that's not neither reviewed nor tested yet so I'd like to have these two steps.

holgerfriedrich:
@mstormi  I do not like changing the scope of the "Upgrade System - Upgrade all installed software packages to their newest version". This is silently skipping the Java update and users still think everything is upgraded to latest. Current behaviour is annoying but the result is ok.

As a hotfix this would have been nice in conjunction with a popup notifying the users when upgrade is selected. Maybe better to go for #848.

mstormi:
Well it's not really changing scope because today it's a **fixed** Azul version that gets installed (and not the latest from repo as it was the case before Elias had to change that because Azul stopped providing repo updates an started changing their download website).
So "upgrade Java" hasn't been in scope of option 02 for quite some time now.

I don't mind skipping this as it was meant to be a hotfix only and to go for #848 instead.
But is that ready for prime time ?

PR #820, opened 2020-03-31
start debug guide and change default output on failures to point there

### LINKED ISSUE #808: Guide to open an issue

mstormi:
Fixes #808
 
### DIFF
--- a/functions/menu.bash
+++ b/functions/menu.bash
@@ -205,5 +205,5 @@ show_main_menu() {
   fi
 
   # shellcheck disable=SC2154,SC2181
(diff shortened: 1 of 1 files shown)


mstormi:
Superseded by #683

PR #829, opened 2020-04-04
Add WiFi for RPI4

### LINKED ISSUE #725: WiFi config fails on RPiA3+
Setup:
Platform: RPi3A+
Storage: Kingston 64GB µSD
Build: openhabian-pi-raspbian-201908050414-gitca0976f-crc6a66b5a1
WiFi Accesspoint: AVM Fritz!Box 6490 (FW: 07.10)/Samsung Galaxy S7 (LOS 16)

The config fails to properly connect to WiFi and subsequently fails the connectivity check. The SSIDs are correctly correctly displayed afterwards and WiFi is working when manually connected. Tested several times, with multiple APs and varying SSID/PW configs.

Workaround: Boot without WiFi config in openhabian.conf, login with default raspbian user (user/pw: pi/raspberry), connect to WiFi using `sudo raspi-config` and then restart the first-boot script.

### LINKED ISSUE #767: Wifi Password with special Characters needs to be escaped in openhabian.con
When following instructions
https://www.openhab.org/docs/installation/openhabian.html#wi-fi-based-setup-notes

I need to take care that password with special characters needs to be escaped in `wifi_psk`, e.g. `$` in password needs to be written as `\$`

Maybe this needs to be considered or a hint in the installation description would have saved me a half day

mstormi:
Adds RPi4 to supported systems checked for in wifi_setup()

Changed that to also allow for continuing if none of the boxes to have WiFi builtin are detected (as someone might be using a WiFi hat or USB stick).
Added comment to openhabian.conf to escape special characters on manual WiFi setup

NOTE:
My fixes should not do harm (Pi4 WiFi setup is broken anyway), but I have no RPi4 to test against so cannot validate.

Fixes #767 

Might fix #725

### DIFF
--- a/build-image/openhabian.conf
+++ b/build-image/openhabian.conf
@@ -29,4 +29,4 @@ wifi_country=""
 
 # Java architecture mode
 # Use "32-bit" or "64-bit"
-java_arch="32-bit"
+java_arch="64-bit"
--- a/build-image/openhabian.conf
+++ b/build-image/openhabian.conf
@@ -12,8 +12,10 @@ system_default_locale="en_US.UTF-8"
 #timezone=Europe/Berlin
 
 # Wi-Fi settings. An ethernet connection is recommended
-# Requirement: RPi3, RPi0W, PineA64 or a supported external Wi-Fi dongle
+# Requirement: RPi4, RPi3, RPi0W or a supported external Wi-Fi dongle
 # Fill in your SSID and password below, leave empty to use ethernet
+# ATTENTION: you need to escape these special characters: $, `, ", \, (newline)
+# 'Escaping' means to put an additional \ in front of that character
 wifi_ssid=""
 wifi_psk=""
 
@@ -27,4 +29,4 @@ wifi_country=""
 
 # Java architecture mode
 # Use "32-bit" or "64-bit"
-java_arch="64-bit"
+java_arch="32-bit"
--- a/build-image/openhabian.pi-raspbian.conf
+++ b/build-image/openhabian.pi-raspbian.conf
@@ -12,8 +12,10 @@ system_default_locale="en_US.UTF-8"
 #timezone=Europe/Berlin
 
 # Wi-Fi settings. An ethernet connection is recommended
-# Requirement: RPi3, RPi0W, PineA64 or a supported external Wi-Fi dongle
+# Requirement: RPi4, RPi3, RPi0W or a supported external Wi-Fi dongle
 # Fill in your SSID and password below, leave empty to use ethernet
+# ATTENTION: you need to escape these special characters: $, `, ", \, (newline)
+# 'Escaping' means to put an additional \ in front of that character
 wifi_ssid=""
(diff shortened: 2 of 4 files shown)


build-image/openhabian.conf
```diff
@@ -12,8 +12,10 @@ system_default_locale="en_US.UTF-8"
 #timezone=Europe/Berlin
 
 # Wi-Fi settings. An ethernet connection is recommended
-# Requirement: RPi3, RPi0W, PineA64 or a supported external Wi-Fi dongle
+# Requirement: RPi4, RPi3, RPi0W or a supported external Wi-Fi dongle
```
holgerfriedrich, line 15: ~~Ok to state RPi4 here, as it would be rolled out not before we publish a new image....
In other documents I would like to see it running before we put it as supported.~~
Maybe PineA64 should be kept (and removed in the corresponding PR which removed Pine support)

build-image/openhabian.conf
```diff
@@ -12,8 +12,10 @@ system_default_locale="en_US.UTF-8"
...
 # Fill in your SSID and password below, leave empty to use ethernet
+# ATTENTION: you need to escape these special characters: $, `, ", \, (newline)
+# 'Escaping' means to put an additional \ in front of that character
```
holgerfriedrich, line 18: I think this needs to stay here since we need to write down a properly escaped bash string even though wpa_passphrase will handle the actual writing to wpa_supplicant.conf....
It will anyway not handle special characters in SSID, so entering " in SSID will break the config file, \" should be fine.

build-image/openhabian.conf
```diff
@@ -27,4 +29,4 @@ wifi_country=""
 
 # Java architecture mode
 # Use "32-bit" or "64-bit"
-java_arch="64-bit"
\ No newline at end of file
+java_arch="32-bit"
```
holgerfriedrich, line 32: change not useful here. 
did you notice that we have 3 files in build-image? 
openhabian.conf: -> local_test
openhabian.pi-raspbian.conf: -> build image for pi
openhabian.pine64-xenial.conf: -> build image for pine
...plus openhabian.conf.dist....

I think we need to cleanup this mess later....

build-image/openhabian.conf
mstormi: No as a PINE already is not guaranteed to work

build-image/openhabian.conf
mstormi: ok so we'll keep this new comment

build-image/openhabian.conf
mstormi: agree, I removed the change here it slipped in when working on the java stuff in parallel

mstormi:
are you ok with this ?

holgerfriedrich:
I'd suggest do copy the change "Requirement: RPi4, RPi3, RPi0W..." also to openhabian.pi-....conf.
Then you can go ahead and merge!

PR #836, opened 2020-04-06
Change java-jre.bash to use AdoptOpenJDK

### LINKED ISSUE #766: Change Java provider from Azul
Why not change our Java provider from Azul to something like AdoptOpenJDK? This is more of a brainstorming post than a true issue mainly stemming from the fact that whit the change of Azul's API they no longer appear to incorporate Linux in the mix of available Java builds in their API.

ecdye:
After the Zulu API left alpha the Java versions that openhabian required
were no longer being distributed. As a result this commit changes the
default Java provider on openhabian to be AdoptOpenJDK.

I have personally tested this change and in my use it had performed
without any issues however, that many not be the case for everywhere.

Closes #766

### DIFF
--- a/functions/java-jre.bash
+++ b/functions/java-jre.bash
@@ -34,14 +34,13 @@ java_adoptopenjdk_8_tar(){
   local jdkBin
   local jdkLib
   local jdkArch
-  local jdkSecurity
   if [ "$1" == "32-bit" ]; then
     echo -n "$(timestamp) [openHABian] Installing Java AdoptOpenJDK 32-Bit OpenJDK... "
     if is_arm; then
-      link="https://api.adoptopenjdk.net/v2/binary/releases/openjdk8?openjdk_impl=hotspot&os=linux&arch=arm&release=latest&type=jdk"
+      link="https://api.adoptopenjdk.net/v3/binary/latest/8/ga/linux/arm/jdk/hotspot/normal/adoptopenjdk?project=jdk"
       jdkArch="aarch32"
     else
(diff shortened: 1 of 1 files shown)


ecdye:
Also this is not a completely perfect solution as my programming skills are meager and this script could use the ability to properly check if an update is needed but that is beyond my skill level and since it never worked in the previous script I see no harm in leaving it out for now.

mstormi:
Hello Ethan, thank you for your submission. 
This is not full a code review yet but I think I should give a little bit of background to explain what we would expect from a contribution, to avoid disappointment on everyone's part.
For a start, please have a look at CONTRIBUTING.md.
On Adopt: we have not decided which Java to go for and when, but I'm pretty sure that we'll not simply replace Azul by Adopt or any other JDK in a single step right tomorrow as your patch would do if we merged it, and to do so just because there's word in the world that Adopt is working fine.
We would need to validate that first, and - even more important - we need users to participate in testing, too, using different hardware of theirs, including x86, so we get more information on stability, performance, mem usage etc and thus a much broader basis to decide upon.
You could help by turning your Adopt addition into an _option_ in the first place, i.e. add your code, add an uninstall routine, too, and add a 4X option in menu.bash to install Adopt and use update-alternatives to make "java" and "javac" point there. Switching defaults will then become a much quicker task.
AFAIK Adopt allows for a repo based install which is the preferred way to install things in openHABian, Azul Java being an exception to the rule because they don't offer package install (any more). Please change your code to make installation work based on apt packages.
Don't remove Azul install right away, keep it to be the standard for the time being.
Please also contribute a bats unit test for Adopt install, see java-jre.bats.
I hope this response does not demotivate you - again, thanks for your contribution!

PR #837, opened 2020-04-07
re-enable shellcheck in builds

mstormi:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,6 +45,7 @@ jobs:
         docker exec -it "bats-dest-tests" bash -c 'bats -r -f "destructive-." .'
 # for now spellcheck tests are not enforced, return true to consider result as ok
     - stage: Shellcheck
-      script: shellcheck -x -s bash openhabian-setup.sh
-    - script: shellcheck -x -s bash functions/*.bash
-    - script: shellcheck -x -s bash build-image/*.bash
+      script: shellcheck -x -s bash openhabian-setup.sh && echo "shellcheck openhabian-setup.sh - OK"
+    - script: shellcheck -x -s bash functions/*.bash && echo "shellcheck functions/*.bash - OK"
+    - script: shellcheck -x -s bash build-image/*.bash && echo "shellcheck build-image/*.bash - OK"
+    - script: shellcheck -x -s bash build.bash && echo "shellcheck build.bash - OK"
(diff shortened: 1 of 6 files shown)


.travis.yml
```diff
@@ -45,6 +45,6 @@ jobs:
         docker exec -it "bats-dest-tests" bash -c 'bats -r -f "destructive-." .'
 # for now spellcheck tests are not enforced, return true to consider result as ok
     - stage: Shellcheck
-      script: shellcheck -x -s bash openhabian-setup.sh || true
-    - script: shellcheck -x -s bash functions/*.bash || true
-    - script: shellcheck -x -s bash build-image/*.bash || true
+      script: shellcheck -x -s bash openhabian-setup.sh
```
holgerfriedrich, line 48: since shellcheck is quiet if noting is found, you do not see any sign of it in the logs... I would rather write 
`script: shellcheck -x -s bash openhabian-setup.sh && echo "shellcheck openhabian-setup.sh - OK"`

holgerfriedrich, line 48: what about build.bash? (sorry it is not shellcheck-clean right now, but nothing major)

.travis.yml
```diff
@@ -45,6 +45,6 @@ jobs:
...
+    - script: shellcheck -x -s bash functions/*.bash
+    - script: shellcheck -x -s bash build-image/*.bash
```
holgerfriedrich, line 50: maybe the more general approach would be better, i.e. do it per default for any .bash / ,sh file in the repo.
This could by done by running a openhabian-shellcheck.bash script which does a find over the repo and calls shellcheck accordingly for each file except the dashboard-include sh. preferrably with an output in the log for each checked file.
(find . -iname "*.sh" && find . -iname "*.bash")|grep -v dashboard)... 
But we can change this later. Please do not block this merge by introducing this now.

.travis.yml
mstormi: ok

mstormi: shellchecked && added build.bash

.travis.yml
mstormi: Ok I'll merge like this. I wouldn't agree anyway as only a build with NO shellsheck errors whatsoever is a good build, and any author should know how to manually shellcheck his changes anyway

mstormi:
didn't know about "draft" status until now.
Ready to merge ? (I think there should be a button you would need to hit to confirm review is finished

holgerfriedrich:
@mstormi cannot see the echo of shellcheck ok in the travis log, can you please verify

mstormi:
This is because of an error in .travis.yml which gets fixed in #682. I moved the fix here, too, I assume it'll work out now, see https://travis-ci.com/github/mstormi/openhabian/builds/159817936

PR #841, opened 2020-04-15
WiFi credentials handled by wpa_passphrase

holgerfriedrich:
WiFi credentials handled by wpa_passphrase to ensure correct escaping of special characters


holgerfriedrich:
fixes #767

Issue #721, opened 2019-09-18
Beautify install output


mstormi:
Using the RPi image, openhabian install starts before the boot process ended, resulting in an ugly mix of log lines printed to the console.
The whole look isn't _trustworthy_ .
Not sure what's the best fix, a sleep or dependency in systemd to start later.

ThomDietrich:
I didn't get that. Could you explain in another way? There are steps (like the loop around apt-get update) to ensure that parts of the system are ready when needed

mstormi:
![IMG_20190923_165738](https://user-images.githubusercontent.com/12220522/65438731-4b99b980-de26-11e9-9694-7026c0b653c5.jpg)

This looks messy because installation is started before the boot sequence finishes.

mstormi:
@ThomDietrich did this happen on your recent install ? (or did you not use a console? If no please try with one)

holgerfriedrich:
Looked the same when I did an install recently.

Maybe we should make sure normal system start is done before starting installation. This may also have positive side effects like system time being set before....

I would also propose to use another console for the detailed output which can instantly be seen when strg-alt-f2 is hit.

mstormi:
Holger would you say we're done here ?

mstormi:
My most recent install with the beta image on a RPi4 showed little but still some remaining overlapping.
The first OHian install related message after which there were no more boot messages has a timestamp of 60.xxx.

So Holger can insert an additional delay of 1 minute before executing first-boot?
That should do I guess.

mstormi:
@holgerfriedrich could you add that 1 min delay then close this ?

holgerfriedrich:
@mstormi I don't think that adding delays is the right way to go here, though it makes the install look nice. People will wonder why they cannot connect via webif for such a long time, etc. WiFi based install will wait twice due to reboot. 
For me it would be much cleaner solution to look which services are producing the output and play with the systemd dependency/after/... options to resolve.
I won't be able to do this the next days.

ecdye:
> For me it would be much cleaner solution to look which services are producing the output and play with the systemd dependency/after/... options to resolve.

In this case I agree with Holger, we should look into other solutions. This does not need to be in before a new image so we can take our time.

mstormi:
Did that already: https://github.com/openhab/openhabian/pull/985#issuecomment-654664698

ecdye:
Yeah, I noticed that after I posted that comment.

mstormi:
[...]

Issue #828, opened 2020-04-04
first-boot.sh start should not rely on rc.local


holgerfriedrich:
Running of first-boot.sh is currently done via the rc.local file on Rapsbian based image.
This is long deprecated and only works via a compat layer, i.e. a service which runs this script.
Typically it is expected that the commands from rc.local complete in short time; though we run the full installation there in the foreground, which might not be a good idea.

We have seen some trouble caused by this, e.g. #648, #721, and maybe more since the installations starts already during system start.

My proposal would be to replace the rc.local hook and properly register a system service.

obmgodiya:
this is what I did for the mean while as a workaround
1. let the first boot fail
2. comment the following line under /boot/first-boot.bash

`#[` -d /opt/openhabian/ ] && rm -rf /opt/openhabian/ # check if we have remnants of a previous installation attempt.
#git clone -b master https://github.com/openhab/openhabian.git /opt/openhabian &>/dev/null`

3. comment the following lines under /opt/openhab/functions/openbah.bash
`#  cond_redirect wget --no-check-certificate -qO - 'https://bintray.com/user/downloadSubjectPublicKey?username=openhab' | apt-key add -`
4. comment under /opt/openhabian/build-image/webif.bash
`#  rm /boot/webif.bash > /dev/null`
5. remove the following files
`rm /boot/first-boot.log
rm /opt/openHABian-*
rm /opt/afterfirstboot.lock
touch /opt/openHABian-install-inprogress`
6. reboot

PS: sorry the the poor formatting. I will check this thread for updates.

gitMiguel:
> My proposal would be to replace the rc.local hook and properly register a system service

+1 to this.

And when the installation depends on internet connectivity we should wait for proper systemd hooks before starting the install script. Please see https://wiki.freedesktop.org/www/Software/systemd/NetworkTarget/#runningservicesafterthenetworkisup.

holgerfriedrich:
@gitMiguel thanks for the input. 

A few other things are to be considered, e.g.
- wifi setup stored in openhabian.conf needs to be patched into the system config (as early as possible....)
- for sake of clean outputs of the installer at best the system startup is finished
- Raspbian itself has delayed tasks which start with a small delay after the graphical target is reached
- apt daily needs to be suppressed since it interferes with apt called by our installer (that causes some of the failed installations)
[...]

PR #844, opened 2020-04-16
delayed execution of installer until boot process completed

holgerfriedrich:
delayed start of installer, wait until multiuser target is reached

### DIFF
--- a/build-image/first-boot.bash
+++ b/build-image/first-boot.bash
@@ -17,6 +17,11 @@ fail_inprogress() {
   exit 1
 }
 
+restart_apt-daily() {
+  echo -n "$(timestamp) [openHABian] Restarting apt-daily service... "
+  service apt-daily start
+}
+
 echo "$(timestamp) [openHABian] Starting the openHABian initial setup."
 rm -f /opt/openHABian-install-failed
 touch /opt/openHABian-install-inprogress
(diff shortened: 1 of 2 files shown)


mstormi:
So that'll apply only if we build a new image, right?
Good all by itself thus approved it, but is there any chance to get this out without replacing the image ? Maybe as a routine called by menu option 65  ?

Also, can you check if that fixes #721 ?

holgerfriedrich:
Partly. I delayed until multiuser target is reached. So mixing installer output with system messages is reduced.

However, Raspbian delayed some apt operations even longer, and those interfere with the installer.

I don't think I will merge that. Let me see if I can install a service, then we get rid of the rc.local messages on failed installs.

And yes, rollout is once we release a new image... I would not add this to openhabian-config. Let's red rid of those workarounds.

mstormi:
> And yes, rollout is once we release a new image... I would not add this to openhabian-config. Let's red rid of those workarounds.



>  I would not add this to openhabian-config. Let's red rid of those workarounds.

Don't understand what you mean by 'workaround'.
I didn't mean to keep rc.local but meant to add _editing_ of systemd config to openhabian-config so anyone can benefit not just those to download the new image (once built).

holgerfriedrich:
Just delaying does not work, will need some more efforts to correctly WiFi credentials from openhab.conf.

gitMiguel:
Please see my https://github.com/openhab/openhabian/issues/828#issuecomment-628601940. I think that delaying until internet connectivity would solve this too.

holgerfriedrich:
superseded by #890

PR #845, opened 2020-04-16
added repository url to clone from to openhabian.conf templates

holgerfriedrich:
new images won't work as first-boot.bash won't find a repo to clone from.
Alternative approach would be to code a default value into the scripts.


build-image/first-boot.bash
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 
-REPOSITORYURL=$(grep -i '^repositoryurl' /etc/openhabian.conf | cut -d '=' -f2)
-CLONEBRANCH=$(grep -i '^clonebranch' /etc/openhabian.conf | cut -d '=' -f2)
+REPOSITORYURL=$(grep -i '^repositoryurl' /boot/openhabian.conf | cut -d '=' -f2)
+CLONEBRANCH=$(grep -i '^clonebranch' /boot/openhabian.conf | cut -d '=' -f2)
```
mstormi, line 4: Consider merging #683 first, then source init.bash

build-image/openhabian.conf
```diff
@@ -23,8 +23,12 @@ wifi_psk=""
...
-
 # Java architecture mode
 # Use "32-bit" or "64-bit"
-java_arch="64-bit"
\ No newline at end of file
+java_arch="64-bit"
+
+# repo to clone from (if not master for build and test)
+repositoryurl=https://github.com/openhab/openhabian.git
+clonebranch=master
```
mstormi, line 32: I know it's messy but for a start, let not build-image/openhabian*conf get out of sync with openhabian.conf.dist
See #683 is adding a "mode" parameter
(check https://github.com/mstormi/openhabian/blob/buildsystem-init.bash/build-image/first-boot.bash)

In general, it's probably a better idea to refactor this and _generate_ build-image/openhabian*conf _from_ openhabian.conf.dist, or turn them into 'diff' files and copy openhabian.conf.dist, then apply the diff everywhere in the code where that so far accessed static build-image/openhabian*conf.

build-image/openhabian.pi-raspbian.conf
```diff
@@ -23,8 +23,12 @@ wifi_psk=""
...
 
-# vim: filetype=sh
-
 # Java architecture mode
 # Use "32-bit" or "64-bit"
 java_arch="32-bit"
+
+# repo to clone from (if not master for build and test)
+repositoryurl=https://github.com/openhab/openhabian.git
+clonebranch=master
```
mstormi, line 32: see above

build-image/openhabian.pine64-xenial.conf
```diff
@@ -23,8 +23,12 @@ wifi_psk=""
...
 wifi_country=""
 
-# vim: filetype=sh
-
 # Java architecture mode
 # Use "32-bit" or "64-bit"
 java_arch="32-bit"
+
+# repo to clone from (if not master for build and test)
+repositoryurl=https://github.com/openhab/openhabian.git
```
mstormi, line 31: see above

build-image/openhabian.conf
holgerfriedrich: I would propose to keep them in sync manually for now. Let us first focus on the important problems, then remove pinea64 support from build.bash and then see what to refactor....
Ok?

build-image/first-boot.bash
holgerfriedrich: 683 has a few open points I don't know how to handle, especially introducing foreign code in out repo - would like to discuss this a bit more before merging.

/boot/openhabian.conf is copied over to /etc, but not yet when the code is executed in first-boot.bash.... if we want to use init.bash, we need to put it after this commands.

holgerfriedrich:
@mstormi I still like to merge this one, as it blocks my tests of #844 - ok for you?

mstormi:
> @mstormi I still like to merge this one, as it blocks my tests of #844 - ok for you?

I'd rather like to sort things out in the proper order rather then do it half-baked and twice.
I.e. 683 first then this one sourcing init.bash - and preferrably right in the first place, i.e. deduplicating the multiple .confs - then finally PINE removal (which is way bigger than this part here).
Sorry I don't see the point/don't understand what's blocking you ? Just copy over to your #844 dev repo whatever you need from this one here.

Let me know what holds you off from merging 683 so we can find a quick solution to that.

build-image/first-boot.bash
mstormi: > 683 has a few open points I don't know how to handle, especially introducing foreign code in out repo - would like to discuss this a bit more before merging.

ok. I don't understand what you refer to here (foreign code?) but let's discuss over at #683

mstormi:
_Ich verstehe deinen Verweis auf 683 und dort offene Punkte nicht.
Hilft es wenn wir mal darüber telefonieren? Ansonsten beschreib deine Fragen mal ausführlich auf der Teamseite, aber manchmal ist Telefonieren besser._

mstormi:
With latest #683, repositoryurl and clonebranch are now source in from (build-image/)openhabian.conf

Issue #629, opened 2019-08-04
New JAVA for openHAB3


EliasGabrielsson:
There are discussion on updating JAVA for openHAB 3.0 to at least JAVA 11.
https://github.com/openhab/openhab-distro/issues/768

Nothing new really, just update the packages when times come.
Java Zulu are providing 32-bit packages so we don't need to drop support for older RPis.

mstormi:
There have been a number of proposals, issues and pieces of information related to what Java to use in the future such as  #805, #766, #723, #714, a number of forum posts such as https://community.openhab.org/t/running-openhab-2-on-openjdk/21443/8?u=benjy and more.

I'd like to be this the central place for discussing the way forward w.r.t. Java provider.

mstormi:
Ok, I'll throw my hat to the round as a starter proposal:
* Let's not act overhasty and change the standard from Azul to whatever before OH 3 is ready.
  We should fix current issues to affect our current (Azul) Java install first such as #723 and #729 .
* Let's identify a JDK to be:
   Java v11, to have a JIT compiler, to be Open Source of course, to offer repo based install
   AdoptOpenJDK was proposed, other suggestions welcome
* let's add that _asap_ as an install _option_ in openHABian (alternative to Azul) to get this out to as many voluntary testers as possible to gain experiences
  (not sure if we can go straight to v11 with 2.5.3/3.0beta or still need to provide a v8 package)

mstormi:
Going for Adopt 11 _and_ Zulu 11 now, with Zulu 8 to remain the default for now.

mstormi:
@ecdye would you say we're done with this ?

ecdye:
Yeah, It has been implemented and so far seems to be solid so I think we are done.

mstormi:
Great. Thanks again for your contribution.
(btw I'm running Zulu11 for some days now).

ecdye:
Glad to hear that it is working for you.

PR #848, opened 2020-04-16
Use Zulu API v1 when upgrading Java

### LINKED ISSUE #714: Filter Java Zulu JSON on "os-type" again
Azul changed the JSON structure so OS is now an array parameter instead of a key-value.
Refactor the `jq` filter to make use of this OS parameter.

### LINKED ISSUE #723: current Java gets installed again and openHAB restarted on every system upgrade from openhabian-config

ecdye:
Additionally, to clarify what Java version this will install, when
available, the API will prefer downloading a JRE. However, when a JRE is
not available it will use a JDK instead. However, when JDK's are
installed only the essential binaries are linked to `update-alternatives`
so extra JDK related software is unlinked (i.e. `javac`).

At the moment, Azul appears to only provide JDK builds for ARM platforms 
and so, as such, ARM platforms are most often the ones to not get a JRE
installed and the resulting installation tends to be a bit larger (~100
MB).

One minor inconvenience is that with the current Azul API it does not
allow explicitly choosing builds without the JavaFX additional binaries.
So at times, the Java version installed will have extra JavaFX binaries.
This is usually seen when the architecture being installed is x86 based.

Fixes #714
Fixes #723
See Also #629 

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -62,8 +62,7 @@ jobs:
         - docker build -t openhabian/${HW}-openhabian -f Dockerfile.openhabian-x86 .
         - echo -e "\n\e[36mopenHABian test installation 64bit Azul Enterprise Java on virtual $HW ...\n"
         - docker run --name "openhabian-${HW}" -d openhabian/${HW}-openhabian;
-          docker exec -i "openhabian-${HW}" bash -c \
-            'VIRT=$(sudo virt-what); echo -e "\n\e[37mTesting openhabian install on ${VIRT} ...\n"; ./build.bash local-test && mv ~/.profile ~/.bash_profile && /etc/rc.local'
+          docker exec -i "openhabian-${HW}" bash -c './build.bash local-test && mv ~/.profile ~/.bash_profile && /etc/rc.local'
     - env: arm32hf
       arch: arm64
       script:
@@ -73,8 +72,7 @@ jobs:
         - docker build -t openhabian/${HW}-openhabian -f Dockerfile.openhabian-ARMv6hf-native .
         - echo -e "\n\e[36mopenHABian test installation 32bit Azul Embedded Java on virtual $HW ...\n"
         - docker run --name "openhabian-${HW}" -d openhabian/${HW}-openhabian;
-          docker exec -i "openhabian-${HW}" bash -c \
-            'VIRT=$(sudo virt-what); echo -e "\n\e[37mTesting openhabian install on ${VIRT} ...\n"; ./build.bash local-test && mv ~/.profile ~/.bash_profile && /etc/rc.local'
+          docker exec -i "openhabian-${HW}" bash -c './build.bash local-test && mv ~/.profile ~/.bash_profile && /etc/rc.local'
     - env: arm32hf
       arch: arm64
       script:
@@ -84,20 +82,7 @@ jobs:
         - docker build -t openhabian/${HW}-openhabian -f Dockerfile.openhabian-arm32hf-native .
         - echo -e "\n\e[36mopenHABian test installation 32bit Azul Embedded Java on virtual $HW ...\n"
         - docker run --memory="1g" --memory-reservation="1g" --name "openhabian-${HW}" -d openhabian/${HW}-openhabian
-        - docker exec -i "openhabian-${HW}" bash -c \
-            'VIRT=$(sudo virt-what); echo -e "\n\e[37mTesting openhabian install on ${VIRT} ...\n"; ./build.bash local-test && mv ~/.profile ~/.bash_profile && /etc/rc.local'
-    - env: aarch64
-      arch: arm64
-      script:
-        # 64bit Azul Java on 64bit ARMv8 (RPi3, CM3) on 32bit OS on Travis Arm64 HW -- supposed to fail !
-        - export HW=rpi3
-        - sed -i 's/java_arch=32-bit$/java_arch=64-bit/' build-image/openhabian.conf
-        - docker build -t openhabian/${HW}-openhabian -f Dockerfile.openhabian-aarch64-native .
-        - echo -e "\n\e[36mopenHABian test installation 64bit Azul Embedded Java on virtual $HW ..."
-        - echo -e "\e[31m ** This is known to not work. openHABian is supposed to force install down to 32bit **\n"
-        - docker run --memory="1g" --memory-reservation="1g" --name "openhabian-${HW}" -d openhabian/${HW}-openhabian;
-        - docker exec -i "openhabian-${HW}" bash -c \
-            'VIRT=$(sudo virt-what); echo -e "\n\e[37mTesting openhabian install on ${VIRT} ...\n"; ./build.bash local-test && mv ~/.profile ~/.bash_profile && /etc/rc.local'
+        - docker exec -i "openhabian-${HW}" bash -c './build.bash local-test && mv ~/.profile ~/.bash_profile && /etc/rc.local'
     - env: arm32hf
       arch: arm64
       script:
@@ -107,8 +92,31 @@ jobs:
         - docker build -t openhabian/${HW}-openhabian -f Dockerfile.openhabian-arm32hf-CM3 .
         - echo -e "\n\e[36mopenHABian test installation 32bit Azul Embedded Java on virtual $HW ...\n"
         - docker run --memory="1g" --memory-reservation="1g" --name "openhabian-${HW}" -d openhabian/${HW}-openhabian;
-        - docker exec -i "openhabian-${HW}" bash -c \
-            'VIRT=$(sudo virt-what); echo -e "\n\e[37mTesting openhabian install on ${VIRT} ...\n"; ./build.bash local-test && mv ~/.profile ~/.bash_profile && /etc/rc.local'
+        - docker exec -i "openhabian-${HW}" bash -c './build.bash local-test && mv ~/.profile ~/.bash_profile && /etc/rc.local'
+    - env: arm32hf
+      arch: arm64
+      script:
+        # 64bit Azul Java on 32bit ARMv8 (RPi3, CM3) on 32bit OS on Travis Arm64 HW
+        # This is known to not work - openHABian is supposed to downgrade to 32bit on install.
+        - export HW=rpi3
+        - sed -i 's/java_arch=32-bit$/java_arch=64-bit/' build-image/openhabian.conf
+        - docker build -t openhabian/${HW}-openhabian -f Dockerfile.openhabian-arm32hf-native .
+        - echo -e "\n\e[36mopenHABian test installation 64bit Azul Embedded Java on virtual $HW ..."
+        - echo -e "\e[31m ** This is known to not work. openHABian is supposed to force install down to 32bit **\n"
+        - docker run --memory="1g" --memory-reservation="1g" --name "openhabian-${HW}" -d openhabian/${HW}-openhabian;
+        - docker exec -i "openhabian-${HW}" bash -c './build.bash local-test && mv ~/.profile ~/.bash_profile && /etc/rc.local'
+    - env: aarch64
+      arch: arm64
+      script:
+        # 64bit Azul Java on 64bit ARMv8 (RPi3,4) on 32bit OS on Travis Arm64 HW
+        # This is known to not work - openHABian is supposed to downgrade to 32bit on install.
+        - export HW=rpi3
+        - sed -i 's/java_arch=32-bit$/java_arch=64-bit/' build-image/openhabian.conf
+        - docker build -t openhabian/${HW}-openhabian -f Dockerfile.openhabian-aarch64-native .
+        - echo -e "\n\e[36mopenHABian test installation 64bit Azul Embedded Java on virtual $HW ..."
+        - echo -e "\e[31m ** This is known to not work. openHABian is supposed to force install down to 32bit **\n"
+        - docker run --memory="1g" --memory-reservation="1g" --name "openhabian-${HW}" -d openhabian/${HW}-openhabian;
+        - docker exec -i "openhabian-${HW}" bash -c './build.bash local-test && mv ~/.profile ~/.bash_profile && /etc/rc.local'
     - env: aarch64
       arch: arm64
       script:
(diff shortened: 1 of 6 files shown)


ecdye:
@mstormi, comments on implementation and any changes needed? I could also include the changes made by in #819 to simplify the changes made to the Java implementation.

ecdye:
Now please give suggestions on ways to simplify the implementation I have created. Especially on checking for new updates.

mstormi:
Some comments:
Yes include #819 else you would have to rebase when that one is merged.
Don't load the JDK but stay with JRE. No need to bloat the system.
Don't add a new menu option but add your upgrade routine to option 02 (system upgrade).
Add a BATS test to execute install & uninstall. Java is a crucial component, those we need unit tests for.

You can request a review from any maintainer when you think you're done.

ecdye:
It appears that there is already a BATS test for install. Is uninstall necessary? Especially since there is no reason to have a valid uninstall routine because the openHAB runtime can't function without Java.

ecdye:
@mstormi since GitHub limits their formal review request UI to collaborators, this is an old fashioned request for a Formal Review of Code.

mstormi:
>  Is uninstall necessary?

Agree, no.  But make sure the existing BATS test work with your code. 32 and 64 bit.
You removed the fetch test. Can you re-add it to match your new code?

> this is an old fashioned request for a Formal Review of Code.

ok but have a little patience please. I'm very busy and we need to merge complex stuff first to automate testing. I wouldn't want to touch java install code without that in place.

PS: you can switch to declare a PR as Draft or ready for review.

ecdye:
Thanks

mstormi:
is this supposed to fix #714? If so state in your first comment please (that'll auto-close the referenced issue on merge)

ecdye:
Oh, yes I forgot about that, thanks for the reminder. Should be fixed now

mstormi:
> Should be fixed now

err, no?

ecdye:
> > Should be fixed now
> 
> err, no?

On my end it says it is. It appears under the linked issues. Try refreshing on your end. GitHub might just be slow.

mstormi:
yeah really slow still doesn't show. Maybe because of draft status ?

ecdye:
Possibly

mstormi:
> @mstormi since GitHub limits their formal review request UI to collaborators,

To collaborators or contributors ? The latter you are now, see if you can request now.

ecdye:
I don't see any option to do so yet. I think that the team settings limit requests for reviews to those with write access only.

ecdye:
@mstromi if you really want to, you may be able to change what levels of organization membership are allowed to request reviews possibly allowing me to be able to.

However, if it has the potential to cause workflow problems I wouldn't worry about it too much.

ecdye:
One question I do have, what are the opinions on whether the Zulu version string should be used to check for updates, or should I leave it as is and use the JDK version string to check for updates? Eg. Zulu version: `8.44.0.213` vs. JDK version: `1.8.0_242`

ecdye:
If you don't prefer the JDK then it will error on non arm builds because the JRE doesn't include javac and the Azul appears to only build JDKs for arm at the moment. Additionally, the JDK builds for arm appear to be updated less often, however they are updated (possibly one release behind the x86 versions but I have no proof).

mstormi:
> If you don't prefer the JDK then it will error on non arm builds because the JRE doesn't include javac and the Azul appears to only build JDKs for arm at the moment.

Hmm. Now IMO we don't need javac (@openhab/openhabian-maintainers @holgerfriedrich any objection ?) so you could remove that two javac lines, too.
We also don't need frequent builds as there's no pending Java issues.

Please rebase (eventually patch #819 in, too).

ecdye:
Well, then ok, if the user wants to have javac then they can link it themselves. Unless any objections I'll remove it. I just figured since @EliasGabrielsson added it in the first place it was probably there for a reason.

EliasGabrielsson:
@ecdye First of all thanks for your contributions in this matter!
As you see have I not really caught up with updating this code when Azul have updated their distrubition mechanics. Good to get some help!

I don't think there are any specific need for`javac` but I have not gone through all addons.
I would say let's remove it and re-add it if something breaks.

holgerfriedrich:
@ecdye I cannot give a good advise here. I stayed away from the java topics until now. I would follow the recommendation by @EliasGabrielsson.

ecdye:
After general consensus, JRE is preferred over JDK and javac is no longer registered with update-alternatives. I left the code to remove javac from update-alternatives in place for now though and it can be removed sometime late this year or early next once we are fairly certain that it is no longer applicable.

ecdye:
@EliasGabrielsson if you can, could you review this PR?

mstormi:
Could you extend your code with a routine to auto-select 32 or 64 bit Java based on detected HW (is_pi2() etc) and memory of the box, then use that one on unattended install ? 
Background: Travis testing shows that 64bit Zulu does not work on aarch64 (RPi3+4).
So for those, enforce 32 bit for the time being.

ecdye:
It already selects based on platform however the arch is up to the user, I believe that for all arm platforms it already defaults to 32-bit. I can enforce 32 bit on pi4 \3 though becuse I just tested on my personal pi4 that no the aarch64 builds dont work. Azul says somewhere that aarch64 builds for armv7> are only by special request.

mstormi:
Strange as the binary is aarch64 and that by definition is ARMv8 or higher.
Yes please enforce 32bit on pi3/4 ... make it a routine with a decision matrix, then call that from unattended install. But don't change anything about the 6X options.

ecdye:
> Strange as the binary is aarch64 and that by definition is ARMv8 or higher.
Yes but Rasbian uses an armv7l kernel which is backwards compatible and Azul builds are only for armv8

Does that look like what you want?

functions/java-jre.bash
```diff
@@ -100,95 +99,117 @@ java_zulu_8_tar(){
...
+
+  elif [ "$1" == "64-bit" ]; then
+    if is_arm; then
+      availableVersion=$(curl -s -H "Accept: application/json" "${link}&arch=arm&hw_bitness=64" | jq -r "$filter")
+    else
+      availableVersion=$(curl -s -H "Accept: application/json" "${link}&arch=x86&hw_bitness=64" | jq -r "$filter")
+    fi
+
+  else
+    echo -n "[DEBUG] Invalid argument to function java_zulu_tar_update_available()"
```
mstormi, line 201: do not output [DEBUG] but "[$timestamp] Installing XXX ...." at the beginning and "OK" or "FAILED" at the end to be consistent with rest of code.

mstormi:
> Yes but Rasbian uses an armv7l kernel which is backwards compatible and Azul builds are only for armv8

Wonder what that is good for then... anyway. I had another test cas with aarch64 Azul on 64bit BalenaOS (no Raspbian 64 available so far) and was wondering why that actually passed, now that's an explanation.

functions/java-jre.bash
```diff
@@ -100,95 +99,117 @@ java_zulu_8_tar(){
   jdkBin=$(find "${jdkInstallLocation}"/*/bin ... -print -quit)
   jdkLib=$(find "${jdkInstallLocation}"/*/lib ... -print -quit)
   cond_redirect update-alternatives --remove-all java
+  ## TODO: remove sometime in late 2020
   cond_redirect update-alternatives --remove-all javac
```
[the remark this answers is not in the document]
ecdye: It is removed, I am leaving that there to make sure it is removed from `update-alternatives` so that there are no remnants of it left behind.

ecdye:
@mstormi I have little experience with docker and Travis, so I hope that this works, if not I am going to need help.

mstormi:
> 
> 
> @mstormi I have little experience with docker and Travis, so I hope that this works, if not I am going to need help.

Neither do I, it's learning on the go :)
But it's building https://travis-ci.com/github/openhab/openhabian/builds/163285067 so you didn't mess up .travis.yml. Let's wait to see if it works out.

mstormi:
See shellcheck errors https://travis-ci.com/github/openhab/openhabian/jobs/326504518

ecdye:
It may not work as intended because of the fact that I am explicitly checking to see if it is a RPi3+/RPi4 and Travis may not report as an RPi. But we shall see.

functions/config.bash
```diff
@@ -33,7 +33,14 @@ clean_config_userpw() {
 ## Update java architecture in config file
 ## Valid options: "32-bit" & "64-bit"
 update_config_java() {
-  cond_redirect grep -q '^java_arch' "$CONFIGFILE" && sed -i "s/^java_arch.*/java_arch=$1/" "$CONFIGFILE" || echo "java_arch=$1" >> "$CONFIGFILE"
+  if [ $1 == "64-bit" ] && is_pithreeplus || is_pifour || is_armv7l || is_armv6l; then
```
mstormi, line 36: This is giving a shellcheck error:
https://travis-ci.com/github/openhab/openhabian/jobs/326601913#L482

.travis.yml
```diff
@@ -85,9 +85,19 @@ jobs:
...
+        # Docker doesn't work on Travis ARM yet (it's still alpha), cannot access privileged FS for now ?
+        # - docker run --name "openhabian-${HW}" --privileged -t -d openhabian/${HW}-openhabian;
         - docker run --name "openhabian-${HW}" -d openhabian/${HW}-openhabian;
           docker exec -i "openhabian-${HW}" bash -c \
             'VIRT=$(sudo virt-what); echo -e "\n\e[37mTesting openhabian install on ${VIRT} ...\n"; ./build.bash local-test && mv ~/.profile ~/.bash_profile && /etc/rc.local'
-
+    - env: arm32hf
+      arch: arm64
+      script:
+        - export HW=rpi3
```
mstormi, line 96: cosmetic only but it's rpi2

functions/config.bash
```diff
@@ -33,7 +33,14 @@ clean_config_userpw() {
 ## Update java architecture in config file
 ## Valid options: "32-bit" & "64-bit"
 update_config_java() {
-  cond_redirect grep -q '^java_arch' "$CONFIGFILE" && sed -i "s/^java_arch.*/java_arch=$1/" "$CONFIGFILE" || echo "java_arch=$1" >> "$CONFIGFILE"
+  if [ "$1" == "64-bit" ] && is_pithreeplus || is_pifour || is_armv7l || is_armv6l; then
```
mstormi, line 36: Is the logic correct?
The test build failed because your code installed 64bit as requested:
https://travis-ci.com/github/openhab/openhabian/jobs/326715364#L1229
But it should have recognized that und downgraded to 32bit.

FWIW use mode=debug_maximum in openhabian.conf if you want debug output of every command in the test build to find the mistake.

mstormi:
Ethan when you change something about a code location where a reviewer requested changes, please comment there, but do not resolve the conversation (leave that up to the requestor to let him confirm)

functions/config.bash
ecdye: Ok, so I'm debugging the implementation right now and will let you know when it works or if I need some help.

mstormi:
It's still installing 64 bit when it should not:
https://travis-ci.com/github/openhab/openhabian/jobs/326774953#L1235

mstormi:
I just merged #859, please rebase on master.
You can omit your changes to .travis.yml, I've put that in

functions/config.bash
mstormi: Latest version still fails to correct 64->32bit on arm32hf
See this (from auto build upon your last upload)
https://travis-ci.com/github/openhab/openhabian/jobs/326831221#L1522

ecdye:
So, at this point I think that is is has something to do with the Virtual environment not being correct as it shown by adding the debug statement `uname -a` right before installing Java. When this is run it still thinks it is an aarch64 platform but the command does not recognize that.

mstormi:
> I think that is is has something to do with the Virtual environment not being correct as it shown by adding the debug statement `uname -a` right before installing Java

I have not seen that. Give a link into the build log please and what you think it should output.

Have you overread my suggestion to set mode=debug_maximum  ?

Also note .... .6 build *are* supposed to fail (because it's all-aarch64 but the kernel is not) while the last (.9) build fails because of 64 installed instead of 32.

mstormi:
More merges hence another need to rebase

ecdye:
https://travis-ci.com/github/openhab/openhabian/jobs/326831221#L1229
Then shouldn't the .6 build be allowed to fail?

ecdye:
I can't figure out how to make the test work on Travis but it works on my own RPi.

mstormi:
> Then shouldn't the .6 build be allowed to fail?

Yes, I changed that in another patch.

> Also I removed .9 because it was not necessary, as .6 should cover the test base that we a
> targeting

No, .6 is 64bit _on aarch64_ so java install cannot know it must throttle back to 32.
Then again .9 I believe was 64bit on arm32hf which was right our target here.
So we still need that, better put it back end else we don't know (by looking at the tests) if it properly downgrades now.

mstormi:
> I can't figure out how to make the test work on Travis but it works on my own RPi.

which one? .9 ?

use the 2nd line in the install: section of .travis.yml to enable maximum debugging

ecdye:
> which one? .9 ?
 Yes, I cant get it to downgrade to 32-bit in Travis testing but in my RPi4 it does what it is supposed to. I'm going to push one more time using you suggestions. If that doesn't work then i think we need to edit the docker container.

ecdye:
@mstormi Thoughts on the weird travis test results?

mstormi:
Your .7 test fails early because it has env:arm32hf but arch: amd64 instead of arm64
We don't know if it would downgrade Java as it does not reach that point.

Change your .travis.yml to https://github.com/mstormi/openhabian/blob/patch-1/.travis.yml
and retry

mstormi:
You fixed the test run now but it's still doing 64bit:
https://travis-ci.com/github/openhab/openhabian/jobs/327435517#L1234
The check in
https://github.com/ecdye/openhabian/blob/patch-3/functions/java-jre.bash#L69
is bad. Get debug output what is `uname -m` in that test case #.7.
Maybe we need to change the helper functions that are based on uname ...

Better to change the check (untested!) to only go 64 when
"is_aarch64() && `getconf LONG_BIT` = "64"
(the latter is a check if the kernel is running in 32 or 64)
Possibly also replace is_aarch64() with "is_pithree || is_pithreeplus || is_cmthree || is_pifour" which are based on reading IDs from the HW not uname -m

ecdye:
That worked.

mstormi:
Ah, finally: https://travis-ci.com/github/openhab/openhabian/jobs/327724052#L1232
It works better than expected, it applies on .8 test too so that one now succeeds, too
(well for now it fails because I had inverted the result in .travis.yml, so please undo that inverting).

Please also output on unattended runs ($INTERACTIVE not set) that it throttles back to 32. 

Yes please reset the clonebranch and repositoryurl.

functions/config.bash
```diff
@@ -33,7 +33,18 @@ clean_config_userpw() {
 ## Update java architecture in config file
 ## Valid options: "32-bit" & "64-bit"
 update_config_java() {
-  cond_redirect grep -q '^java_arch' "$CONFIGFILE" && sed -i "s/^java_arch.*/java_arch=$1/" "$CONFIGFILE" || echo "java_arch=$1" >> "$CONFIGFILE"
+  if [ "$1" == "64-bit" ]; then
+    if ! is_aarch64 && ! [ "$(getconf LONG_BIT)" == "64" ]; then
```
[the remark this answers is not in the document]
ecdye: Good Catch.

mstormi:
Thanks for your contribution.
Just an idea, maybe you want to look into adding AdoptOpenJDK as an option ?

ecdye:
Yep, I've been experimenting in the background with it while working on this.

Issue #832, opened 2020-04-06
Use of GitHub Actions in tests


ecdye:
Should we consider extending our testing procedures to include a GitHub Actions workflow?

holgerfriedrich:
Worth a look, as we could also automate the image creation....

holgerfriedrich:
I played a bit with this feature, and even managed to build our release image for the PI using actions...
https://github.com/holgerfriedrich/openhabian/tree/pr-actions
https://github.com/holgerfriedrich/openhabian/actions/runs/83309994

Not sure if I like this, though. Feels we are duplicating travis CI builds.

ecdye:
Yes, but also the more tests the better. I think it really just depends on how much cost it adds to the workflow. 

If we end up using it for automating releases I think it could be worth it. I'm also interested to hear others thoughts on this.

ecdye:
Also in the last statement of goals made by the Maintainers team one of the points was 
> "Focus of rapid deployment to ensure robust and tested software in opposite to create ready made images with cherrypicked artifacts."

Using GitHub action to create artifacts would certainly help accomplish that.

ecdye:
This would also resolve #697

holgerfriedrich:
Open questions for me:
* I don't want to do things twice, for me it is not clear what we should to with travis and what we should do here.
* Using the build action for the image works on my repository and will create an artifact of ~300MB. For me it is not clear how we deal with storage space and get rid of older artifacts, AFAIK github policy is to expire all artifacts after 90 day.

PR #852, opened 2020-04-22
implemented shellcheck as action; preparations for build

holgerfriedrich:
refers to #832 

### DIFF
--- a/.github/workflows/shellcheck-action.yml
+++ b/.github/workflows/shellcheck-action.yml
@@ -1,0 +1,31 @@
+# This is a basic workflow to help you get started with Actions
+
+name: shellcheck
+
+# Controls when the action will run. Triggers the workflow on push or pull request
+# events but only for the master branch
+on:
+  push
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  # This workflow contains a single job called "build"
+  build:
+    # The type of runner that the job will run on
+    runs-on: ubuntu-latest
+
+    steps:
+       - name: Checkout repository
+         uses: actions/checkout@v2
+       - name: Environment variables
+         run: sudo -E bash -c set
+       - name: Install required packages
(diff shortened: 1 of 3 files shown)


mstormi:
you want this to close #832 ? Or was it Github to misunderstand ?

holgerfriedrich:
@mstormi I closed 697 manually because workaround and error message is in place; this one should not actually close 832, it is just a first step implementing shellcheck as an action and prepare for doing the image preparation via actions. This is already in place at my forked repo.

mstormi:
understood. But note Github interpreted your "refers to ..." as it would interpret "fixes ..." so it'll close that when merging this

mstormi:
Then again while that probably was not your intention you IMHO could leave it like that.
#832 more of an idea than a real proposal where to apply.
I like this idea of yours here to move Shellsheck from Travis to Github as contributors get feedback earlier.
But for the more complex CI stages I'd like to stay (well, in the first place, start ;-)) with Travis.

ecdye:
True enough, #832 really was just a suggestion. It doesn't need to be implemented, and if it is better to do it in stages to adjust our already complicated build and test system.

holgerfriedrich:
What we do not have in place right now is automated creation of the SD card image, and this is what I did with actions on my repo. Not sure if I should add this to the PR.

@mstormi If you want, you could merge this here as as starting point and we will open an new issue to explore the options for creating the image in a new issue....

mstormi:
ok but please proceed and review #684

mstormi:
I can see the action being excuted, but no output whatsoever such as echo OK ?
https://github.com/mstormi/openhabian/actions/runs/85147087/workflow

ecdye:
@mstormi You have to click through to the build menu, then you can see the actions executed and look at the output of each one.

mstormi:
I see

Issue #856, opened 2020-04-26
Update test.bash and documentation


holgerfriedrich:
Test.bash is outdated.

* In #683 Dockerfile has been renamed. We need to update testing related scripts
and CONTRIBUTING.md, README.md `docker build -f Dockerfile.openhabian-x86 ...` *OR* decide  to rename Dockerfile.openhabian-x86 to Dockerfile (affects .travis.yml)
* Shellcheck test is incomplete (build.bash is not shellchecked, see .travis.yml for reference)
* Feature request: add calling bats test w/o Docker
* Add something about test.bash to documentation

mstormi:
@holgerfriedrich mind to finally review ?
It's simple and still affecting users.

mstormi:
Did you try build.bash after #684 meanwhile ?
Latest *is* shellchecked BTW.

holgerfriedrich:
@mstormi I think you got me wrong, test.bash does not do it, so pre-commit testing might overlook it

mstormi:
from .travis.yml: build.bash *is* shellchecked
      - shellcheck -x -s bash openhabian-setup.sh functions/*.bash build-image/*.bash build.bash && echo "shellcheck - OK"

mstormi:
btw does building an image work ? Would like to provide a new one to contain the 02/2020 update from raspberry.org plus our pre-fetched java.

holgerfriedrich:
@mstormi it should. I started using the image 1.6 milestone to collect what is essential from my point of view.

mstormi:
haeh?
Won't wait for 1.6, that would be in time for bullseye upgrade.

holgerfriedrich:
For me 1.6 is just next image milestone. I did remove "bullseye" from title.... bullseye is years away from now

mstormi:
Well usually we only ever released images together with debian upgrades so I was thinking to call the next 1.5.1 then 1.6 would be for bullseye.
Then again, the Java upgrade also is pretty substantial. We can discuss for sure. Or roll dices ;-)

mstormi:
@ecdye did your recent update on test.bash affect or fix this ?

ecdye:
It should have, yes (once #938 is merged).

PR #857, opened 2020-04-26
Fix build.bash parameter check

holgerfriedrich:


mstormi:
is that all what needs to be fixed in build.bash ?
Don't want to do many micro PRs.

holgerfriedrich:
Yes, don't know why we did not notice that normal image creation is broken since last update....

mstormi:
> Yes, don't know why we did not notice that normal image creation is broken since last update....

because we don't test for it ? Can you come up with a BATS or docker test case of building the image ?

mstormi:
what about #856 ?

holgerfriedrich:
@mstormi opened https://github.com/orgs/openhab/teams/openhabian-maintainers/discussions/15

PR #865, opened 2020-05-03
Allow failures in Travis aarch64 builds

ecdye:


mstormi:
not the right fix as we have an aarch64 on 64bit OS, too, that may NOT fail

Use https://github.com/mstormi/openhabian/blob/patch-1/.travis.yml

PR #867, opened 2020-05-04
fix nginx install

### LINKED ISSUE #866: nginx install broken
There's at least 2 problems on nginx install:
Seems a number of directories fail to get created on nginx package install. Need to take care of that in our install routine.
My problem with nginx failing to start started with a missing logdir after works on ZRAM (where /var/log/nginx was deleted).
This was on buster w/ ZRAM to "own" `/var/log` (incl. "`nginx`" below), but there is nowhere in neither the nginx package nor nginx_setup() that creates the `/var/log/nginx` directory for logfiles as they're used in the config that is generated - causing nginx to fail starting up.
Things went south when I advised some user with a broken nginx config to delete /etc/nginx and to reinstall. Seems even /etc/nginx is not created on install, let alone the stuff below.

Second, the config generated with default inputs (default user openhab, HTTPS yes, blank domain) is bad, the first section is partly still commented, see below. @BClark09 could you take a look please ?
```
#################################
# openHABian NGINX Confiuration #
#################################

## Redirection
#server {
#   listen                          80;
   server_name                     localhost;
   return 301                      https://$server_name$request_uri;
#}


```

mstormi:
We may not assume that nginx is installed/running as part of Raspbian already.

So we need to
* create nginx log dir (it may be missing)
* (re)start instead of reload

Fixes #866

Siggned-off-by: Markus Storm <markus.storm@gmx.net>

### DIFF
--- a/functions/packages.bash
+++ b/functions/packages.bash
@@ -518,7 +518,6 @@ nginx_setup() {
 
   if (whiptail --title "Confirmation" --yesno "$confirmtext" 22 80) then
     echo "Installing NGINX..."
-    mkdir -p /etc/nginx && chmod 755 /etc/nginx
     apt-get -y -q install nginx || FAILED=true
 
     rm -rf /etc/nginx/sites-enabled/default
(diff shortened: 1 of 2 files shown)


BClark09:
> EDIT: seems /etc/nginx does not get created on install via apt-get or elsewhere in option 43.
Seems we need to do that as well ?

I'm not sure if I understand this, the `nginx-common` package should be installed that contains this?

mstormi:
> I'm not sure if I understand this, the `nginx-common` package should be installed that contains this?

openHABian does not install it. You may not assume it's present as that's OS dependent.
nginx packages apparently has no dependency on nginx-common.

I now added the latter to the apt-get install line in this patch.

PR #876, opened 2020-05-08
Remove python2 from openHABian

### LINKED ISSUE #875: needed_packages() - python-pip fails on Ubuntu 20.04
Ubuntu 20.04 does not support python 2 by default. So installing need packages through Configuration Tool fails because python-pip package does not exist.

python-pip can still be installed but is not as easy as "apt install ..."
It's described for example here: https://linuxize.com/post/how-to-install-pip-on-ubuntu-20.04/

BTW: Is python 2 really needed by OpenHabian?

holgerfriedrich:
* replace all occurences of python and python-pip by python3 equivalent
* fixes #875 (compatibility with Ubuntu 20.04)

### DIFF
--- a/functions/system.bash
+++ b/functions/system.bash
@@ -26,13 +26,13 @@ needed_packages() {
   # Install apt-transport-https - update packages through https repository
   # Install bc + sysstat - needed for FireMotD
   # Install avahi-daemon - hostname based discovery on local networks
-  # Install python/python-pip - for python packages
+  # Install python/python3-pip - for python packages
   echo -n "$(timestamp) [openHABian] Installing additional needed packages... "
-  if cond_redirect apt-get -y install apt-transport-https bc sysstat avahi-daemon python python-pip avahi-autoipd fontconfig; then echo "OK"; else echo "FAILED"; exit 1; fi
+  if cond_redirect apt-get -y install apt-transport-https bc sysstat avahi-daemon python3 python3-pip avahi-autoipd fontconfig; then echo "OK"; else echo "FAILED"; exit 1; fi
 
   if is_pithree || is_pithreeplus || is_pizerow; then
     echo -n "$(timestamp) [openHABian] Installing additional bluetooth packages... "
-    if cond_redirect apt-get -y install bluez python-bluez python-dev libbluetooth-dev raspberrypi-sys-mods pi-bluetooth; then echo "OK"; else echo "FAILED"; exit 1; fi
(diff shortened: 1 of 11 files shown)


functions/system.bash
```diff
@@ -26,13 +26,13 @@ needed_packages() {
...
-  # Install python/python-pip - for python packages
+  # Install python/python3-pip - for python packages
   echo -n "$(timestamp) [openHABian] Installing additional needed packages... "
-  if cond_redirect apt-get -y install apt-transport-https bc sysstat avahi-daemon python python-pip avahi-autoipd fontconfig; then echo "OK"; else echo "FAILED"; exit 1; fi
+  if cond_redirect apt-get -y install apt-transport-https bc sysstat avahi-daemon python3 python3-pip avahi-autoipd fontconfig; then echo "OK"; else echo "FAILED"; exit 1; fi
 
   if is_pithree || is_pithreeplus || is_pizerow; then
     echo -n "$(timestamp) [openHABian] Installing additional bluetooth packages... "
-    if cond_redirect apt-get -y install bluez python-bluez python-dev libbluetooth-dev raspberrypi-sys-mods pi-bluetooth; then echo "OK"; else echo "FAILED"; exit 1; fi
+    if cond_redirect apt-get -y install bluez python3-bluez python3-dev libbluetooth-dev raspberrypi-sys-mods pi-bluetooth; then echo "OK"; else echo "FAILED"; exit 1; fi
```
mstormi, line 35: So effectively this was the only place (the rest is just used in Travis testing).

However, IIRC Raspbian still has python2 as default what "python" links to, so you actually have to replace all occurrences of "python" and "pip" by "python3"/"pip3", don't you ?

holgerfriedrich: I think I got everything replaced. Let's see what the tests do and then also try a fresh install....

Are you aware of any dependencies of OH2.5 towards Python version 2?

mstormi: no

holgerfriedrich: @mstormi any idea why starting openhab fails in Travis builds for this PR? On my rpi it works.
https://travis-ci.org/github/holgerfriedrich/openhabian/jobs/685007366
Do you know more about the systemd to replacement in those docker images?

mstormi:
Not sure. See https://travis-ci.com/github/openhab/openhabian/jobs/330950127#L1834

Docker builds install python/python-pip (i.e. Python 2) before Python 3 packages so /usr/bin/python will probably point to v2. Try to omit the former.

holgerfriedrich:
@mstormi finally found the call to python2 in the systemd replacement we use in the tests.
Switched to py3 implementation which is also available on the same repo.

mstormi:
thanks

PR #882, opened 2020-05-11
intro stable branch

### LINKED ISSUE #692: Staging/Versioning tags
Currently any user starting openhabian-config gets an immediate update of latest code because HEAD/master is cloned/pulled.
I suggest introducing versioning and tags "current_install" and "current_active" (or similar) that openhabian-config clones instead on image boot and interactive startup, respectively.
OH itself uses snapshot/milestone/release _in addition to versioning_ and we also already have these tags in the issue/PR tracking system for good reason. But our code is still "flat" unversioned, all-or-nothing.
This change will reduce the risk (number of affected installations, actually) of providing bad or insufficiently tested code to all of our users.
Note if it's a fix everybody needs or would benefit from we can always put the "current_active" tag on commits that we want to be distributed asap. But not every change is necessarily a step forward see e.g. those frontail issues.

### LINKED ISSUE #881: openhabian-config to enforce use of openhabian.conf
Use these parameters:
clonebranch, repositoryurl to clone latest openhabian from on startup ?
mode to turn openhabian-config into verbose debug mode(s)

mstormi:
Introduce `stable` as the default branch to clone openHABian from on self-update.

See #692 for the background.
The idea is to have 'patch days' to merge the diffs `master` has over `stable` by (manually) creating PRs to promote those commits that have proven to work from master to stable. This can be all diffs or a selected subset only. Promoting commits via PRs implicitly creates a changelog, too.

The branch to fetch can be selected in menu option 01 (choices: `stable` and `master`).
It can also be overridden by setting `clonebranch=mybranch` in `/etc/openhabian.conf` for the purpose of providing a special version to some openHABian user to have them test it).
$repositoryurl cannot be made point outside to someone's personal repo for safety reasons to avoid accidents and manipulation

TODO:
* ~~make mode=debug_maximum take effect for interactive openhabian-config, too~~
* ~~make clonebranch=(branch) take effect for openHABian repo if set~~
* ~~inform/ask/allow user to permanently switch branches stable<->master~~
* ~~tidy up: do we use master elsewhere?~~
* more testing. You can select stable or master or enter any branch in /etc/openhabian.conf
  "testbuild" is available as a branch for testing. It contains this PR already (plus Adopt JDK to try).

Fixes #692
Fixes #881


mstormi:
@holgerfriedrich could you please review this next (after RPi4 one) ?
I'd like to get that out by Pentecost (Pfingsten).
I intend to ask for a public testdrive of the testbuild branch (to contain new Java providers) on the community forum. This PR (to allow for switching between branches) unfortunately is a prerequisite.
cc: @ecdye

mstormi:
Ready for review now.

functions/openhabian.bash
```diff
@@ -32,26 +31,52 @@ openhabian_console_check() {
...
-    openhabian_update
+    openhabian_update "$branch"
   fi
+  echo -n "$(timestamp) [openHABian] Switching to branch $clonebranch ... "
+  # shellcheck disable=SC2015
+  git -C "$BASEDIR" checkout --quiet "$clonebranch" && echo "OK" || (FAILED=1; echo "FAILED"; return 0)
 }
 
 openhabian_update() {
+  local branch shorthash_before
```
ecdye, line 59: Why not create introtext here too?

functions/openhabian.bash
```diff
@@ -32,26 +31,52 @@ openhabian_console_check() {
...
+
   FAILED=0
+  if [[ -n "$INTERACTIVE" ]]; then
+    if [[ "$current" == "stable" || "$current" == "master" ]]; then
+      if ! sel=$(whiptail --title "openHABian version" --radiolist "$introtext" 14 75 2 stable "recommended standard version of openHABian" on master "very latest version of openHABian" off 3>&1 1>&2 2>&3); then return 0; fi
+    else
+      if ! sel=$(whiptail --title "openHABian version" --radiolist "$introtext" 14 75 3 stable "recommended standard version of openHABian" off master "very latest version of openHABian" off "$current" "some other version you fetched yourself" on 3>&1 1>&2 2>&3); then return 0; fi
+    fi
+    sed -i "s@^clonebranch=.*@clonebranch=$sel@g" "/etc/openhabian.conf"
+  fi
```
ecdye, line 79: So on patch day, are we dumping everyone straight to stable? If so, we are making them aware of the change, correct?

functions/openhabian.bash
mstormi: done

functions/openhabian.bash
mstormi: Yes, we put them on stable.
Haven't decided yet how to make users aware of it.
I'm not even sure we should ?

ecdye: I think we should at least make a statement in the output, perhaps one that contains a link to a community post regarding the change and the new behavior of the utility. 

I don't think it is right to introduce a major change in the functionality / inner workings of the tool without making the user aware.

mstormi: Well it's not really a functionality change, is it ? Less than those we do every day ...
But you're right let's not annoy anybody by hiding that. I'll insert a popup.

mstormi: Like so ?

ecdye: Good enough, better than dealing with the unavoidable posts that say something is wrong.

mstormi: @ecdye @holgerfriedrich 
Just wondering, according to the install docs
https://www.openhab.org/docs/installation/openhabian.html#other-linux-systems-add-openhabian-just-like-any-other-software
if someone git cloned openhabian, he would be getting whatever branch is marked as 'default', wouldn't he ? That's currently 'master'.
Now as we want new users to get stable from now on, would we need to change the default branch ? I _guess_ (but do not know for sure) that that would have unwanted side effects such as whenever you create PRs they will (unless manually changed) also be against stable which also is not what we want.
So question: change default ? or change download instructions to include the stable branch ?

ecdye: Hmm, good question. Perhaps better to open a separate issue discuss it in.

mstormi: I added clone -b stable to the docs that's the safe route

ecdye: I noticed that just after commenting.

ecdye: You are correct in your assumption, GitHub would default to the stable branch if we marked that as the default branch. For now I think that the documentation change should be enough to avoid most of the issues with manual cloning.

PR #883, opened 2020-05-12
Remove setting gnuio options on serial ports

### LINKED ISSUE #532: Don't set Dgnu.io.rxtx.SerialPorts
Currently we are making the following serial port configuration:

```
Dgnu.io.rxtx.SerialPorts=/dev/ttyUSB0:/dev/ttyS0:/dev/ttyS2:/dev/ttyACM0:/dev/ttyAMA0
```

This shouldn't be necessary, as all ports mentioned above are covered in the serial library's port enumeration:

https://github.com/NeuronRobotics/nrjavaserial/blob/4ce383ae8dc8f702438323dd52585d993f58693b/src/main/java/gnu/io/RXTXCommDriver.java#L653-L664

Setting them explicitly prevents the search, so devices not in the list such as `/dev/ttyUSB1` can be missed.

mstormi:
The (current) serial library (now) will take care of scanning ports anyway, and a static list is just _restricting_ use further down to these ports which is not what want.
So I remove menu option 35 selection 3 with this patch.

Fixes #532

### DIFF
--- a/functions/system.bash
+++ b/functions/system.bash
@@ -130,7 +130,6 @@ vimrc_copy() {
 srv_bind_mounts() {
   echo -n "$(timestamp) [openHABian] Preparing openHAB folder mounts under /srv/... "
   sed -i "\\#[ \\t]/srv/openhab2-#d" /etc/fstab
-  # dummy "
   sed -i "/^$/d" /etc/fstab
   ( echo ""
     echo "/usr/share/openhab2          /srv/openhab2-sys           none bind 0 0"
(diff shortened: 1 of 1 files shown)


functions/system.bash
```diff
@@ -130,6 +130,7 @@ vimrc_copy() {
 srv_bind_mounts() {
   echo -n "$(timestamp) [openHABian] Preparing openHAB folder mounts under /srv/... "
   sed -i "\\#[ \\t]/srv/openhab2-#d" /etc/fstab
+  # dummy "
```
BClark09, line 133: ?

mstormi: 😆 vim coloring fails to work properly without that because it doesn't detect the opening " in the line before.
just forgot to remove. There's another PR coming up to remove that part anyway

Issue #648, opened 2019-08-06
First-boot failed on Raspi 3B+ : Failed to start /etc/rc.local Compatibility


lsteelandt:
First boot after installation an SD card with openhabian-pi-raspbian-201908050414-gitca0976f-crc6a66b5a1.img, failed. 

The problem seems to come while executing " Starting /etc/rc.local Compatibility... "
Failed to start /etc/rc.local Compatibility.

Is it suitable for a Raspberry pi 3 model B+   ?

holgerfriedrich:
It should work on a 3B+, this is the same setup I am currently using....

mstormi:
Saw someone else to mention this https://community.openhab.org/t/raspberry-pi-4-openhabian-v1-5-0-error-w-installing-zulu-embedded-openjdk/79370
See there for a quick fix.

mstormi:
@EliasGabrielsson will you create an image to contain java install fix ?

lsteelandt:
@mstormi for the issue describe on this feed, I don't think it's a matter of java installation failure. 
But for the feed you give the link yes. 
Maybe the next image will hold other fixes that resolve my issue (hope to :wink:).

EliasGabrielsson:
I don't get why we should build an new image. To part which were broken, i.e. the failed downloaded Java package have been patched and it not part of the `first-boot.sh` script.

The error here should be resolved by easily flash a new image from scratch.

mstormi:
> The part which were broken [...] is not part of the `first-boot.sh` script.

Wasn't sure of that myself so ok no need for another image.

@lsteelandt did the recent updates fix your problem ?

lsteelandt:
@mstormi oh, i didn't notice that there is a new update.  I will try when I will get a new raspberry.  The one I send for this installation is not yet on my hand

EliasGabrielsson:
Hi @lsteelandt!
Please make another try with the installation and we take it from there. 👍

mstormi:
Got this very same error when testing latest image against a RPi 2

syslog:
```
Sep  6 11:10:11 openhab systemd[1]: rc-local.service: Control process exited, code=exited, status=1/FAILURE
Sep  6 11:10:11 openhab systemd[1]: rc-local.service: Failed with result 'exit-code'.
Sep  6 11:10:11 openhab systemd[1]: Failed to start /etc/rc.local Compatibility.

```
root@openhab:/home/openhabian# systemctl status rc.local
[...]

PR #890, opened 2020-05-17
Introduce systemd service openhabian-installer.service

### LINKED ISSUE #828: first-boot.sh start should not rely on rc.local

holgerfriedrich:
* replace rc.local by openhabian-installer;
  installer is executed at the end of boot process which slightly increases
  the install time for wifi-based installations.
  apt-daily is scheduled after installation to avoid conflicts with apt install
  called by the installer.
* increased duration of network availability check to 60s
* improve output of apt-get update (background)

Fixes #828, #721, #648.


holgerfriedrich:
WIP. 
- Not happy with the output of the installer (should not by tty only, journalctrl should display it as well...)
- Webif.bash should not be removed to allow starting it again
- no special handling for restart, stop yet

holgerfriedrich:
@mstormi yes ready for review. Some room for improvement, but worked fine for me during the tests with the new images.
Maybe the dpkg configure call you suggested should be added to first-boot.bash.
We can start a review... We will see more testing as we proceed building the test images for RasPiOS.....

mstormi:
ok, unmark it from draft status and request a review from either me, Ethan or both
we should stay clean here in terms of procedures

PR #898, opened 2020-05-20
installation of additional bluetooth packages on RPi

holgerfriedrich:
* python3-bluez not installed on stretch based installations, fixes #893
* additional bluetooth packages are now installed also on RPi4

Signed-off-by: Holger Friedrich <mail@holger-friedrich.de>

PR #902, opened 2020-05-20
Fix bluetooth package installation

holgerfriedrich:

### DIFF
--- a/functions/system.bash
+++ b/functions/system.bash
@@ -38,7 +38,8 @@ needed_packages() {
     if ! is_stretch; then
       BTPKGS="$BTPKGS python3-bluez"
     fi
-    if cond_redirect apt-get -y install "$BTPKGS"; then echo "OK"; else echo "FAILED"; exit 1; fi
+    # shellcheck disable=SC2086
+    if cond_redirect apt-get -y install $BTPKGS; then echo "OK"; else echo "FAILED"; exit 1; fi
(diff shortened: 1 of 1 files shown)


holgerfriedrich:
As @mstormi pointed out, #898 fails installing the packages due to the latest shellcheck change. The solution with smallest impact was to remove the quotes again and add an exception.
Other possible way would be to use an array.

Merging (small patch exception)

mstormi:
See https://travis-ci.com/github/openhab/openhabian/jobs/338380361 and
https://travis-ci.com/github/openhab/openhabian/jobs/338380367
That's for x86 and 64-bit OS ARM.

Seems there's more packages that do not exist ?

mstormi:
See also https://community.openhab.org/t/openhabian-install-fails-installing-additional-bluetooth-packages-failed-install-broken-and-unusable/99193

holgerfriedrich:
@mstormi I do not understand what is going on here, I have those packages installed on my system! I had a closer look, the missing ones seem to be from `archive.raspberrypi.org/debian` and not from `raspbian.raspberrypi.org/raspbian`. Let's wait another few days and see the logs.
Thanks for the hotfix, this is good anyway.

mstormi:
Note test builds only fail (due to packages missing) for specific architectures, it's never all of them.

PR #904, opened 2020-05-21
[testbuild] Add Option for AdoptOpenJDK

### LINKED ISSUE #868: AdoptOpenJDK Option
So while researching adding AdoptOpenJDK as a Java provider, I became clear that there are two options. I could create a similar routine to what we use for Zulu or I could create a much more complicated script for `apt-get`. While for managing the installs for just AdoptOpenJDK would be fairly easy. Switching between Zulu and AdoptOpenJDK becomes complicated.

Thoughts?

ecdye:
Fixes #868
See Also #629

### DIFF
--- a/build-image/openhabian.conf
+++ b/build-image/openhabian.conf
@@ -38,7 +38,7 @@ clonebranch=stable
 debugmode=on
 
 # Java architecture mode
-# Use Zulu64, Zulu32, AdoptOpenJDK
-java_opt=Zulu64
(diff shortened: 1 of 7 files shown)


ecdye:
@mstormi Just noticed a re-base artifact in both branches that is causing it to use the old `java_arch` instead of the new `java_opt` in the openhabian*.conf files. I'm going to open a new PR to fix it on the testbuild branch and update the other PR as well.

PR #909, opened 2020-05-23
check for size of external medium before move-root-to-USB

### LINKED ISSUE #424: Move Root on Openhabian-config may need a check for memory stick size
I installed a 128mb memory stick by accident and used the move root.  The function worked fine, but the pi would not boot.  I suspect something happened at copy time or boot time as 128mb is too small.

I would request if possible maybe add a check to make sure memory stick is a minimum size.  Probably no less than the size of the installed SD card.  

Thanks!

### LINKED ISSUE #743: openHABian interactive install/menu option descriptions
Following up on a number of community posts, there's a number of options in openHABian that are unclear what they actually do and when you should or can use them.
Examples:
* 02 Upgrade System
It'll upgrade ALL packages on the system .... including openHAB2 !!
Which may be intentional or not - least thing is to make users aware of it.

* move root to USB
Does it work with any supported HW? Seems not out of the box, see e.g. #731 
How does it relate to ZRAM?
Situations to apply this in:
~~1) to move to an USB stick to avoid SD corruption=> bad idea as the USB is as-bad, better go ZRAM~~
~~2) to move to a "reliable" external medium such as SSD or HDD => ok, but don't use ZRAM on top~~

* ~~Delay rules load~~ --> fixed in #764

* ~~Mosquitto ? How does it relate to MQTTv2 ? Can only have one broker bind to the ports~~
--> there's discussions to deprecate moquette (OH-embedded broker) rather than to extend use
--> more people to switch back to use Mosquitto --> no real need to change openHABian docs

Need to think of the best way of documenting this to users per-issue.
Change menu text ? whiptail infobox ?

mstormi:
enhance in-program menu descriptions and infotexts on the move-root-to-USB option

Fixes: #424 
Fixes: #743 

### DIFF
--- a/functions/ext-storage.bash
+++ b/functions/ext-storage.bash
@@ -15,17 +15,17 @@ move_root2usb() {
 
   if [ -f /etc/ztab ]; then
     if [ -n "$INTERACTIVE" ]; then
-      whiptail --title "Incompatible selection detected" --msgbox "Move root to USB must not be used together with ZRAM.\\nIf you want to mitigate SD card corruption, don't move root but stay with ZRAM, it is the proper choice. If you want to move for other reasons, uninstall ZRAM first then return here." 10 60
+      whiptail --title "Incompatible selection detected" --msgbox "Move root to USB must not be used together with ZRAM.\\nIf you want to mitigate SD card corruption, don't move root but stay with ZRAM, it is the proper choice.\\nIf you want to move for other reasons, uninstall ZRAM first then return here." 10 65
     fi
     echo "FAILED"; return 1
   fi
   if ! (whiptail --title "Move system root to '$NEWROOTPART'" --yes-button "Continue" --no-button "Back" --yesno "$infotext" 22 116); then echo "CANCELED"; return 0; fi
 
-  # check if system root is on partition 2 of the SD card
-  # since 2017-06, rasbian uses PARTUUID=.... in cmdline.txt and fstab instead of /dev/mmcblk0p2...
+  #check if system root is on partition 2 of the SD card
+  #since 2017-06, rasbian uses PARTUUID=.... in cmdline.txt and fstab instead of /dev/mmcblk0p2...
   rootonsdcard=false
 
-  # extract rootpart
+  #extract rootpart
   rootpart=$(sed "s/.*root=\\([a-zA-Z0-9\\/=-]*\\)\\(.*\\)/\\1/" < /boot/cmdline.txt)
 
   if [[ $rootpart == *"PARTUUID="* ]]; then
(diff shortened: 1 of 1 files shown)


functions/ext-storage.bash
```diff
@@ -15,7 +15,7 @@ move_root2usb() {
 
   if [ -f /etc/ztab ]; then
     if [ -n "$INTERACTIVE" ]; then
-      whiptail --title "Incompatible selection detected" --msgbox "Move root to USB must not be used together with ZRAM.\\nIf you want to mitigate SD carsd corruption, don't move root but stay with ZRAM, it is the proper choice. If you want to move for other reasons, uninstall ZRAM first then return here." 10 60
+      whiptail --title "Incompatible selection detected" --msgbox "Move root to USB must not be used together with ZRAM.\\nIf you want to mitigate SD carsd corruption, don't move root but stay with ZRAM, it is the proper choice.\\nIf you want to move for other reasons, uninstall ZRAM first then return here." 10 65
```
radokristof, line 18: Typo: "carsd"

mstormi:
or @ecdye?

functions/ext-storage.bash
```diff
@@ -40,10 +40,25 @@ move_root2usb() {
   if ! [ $rootonsdcard = true ]; then
     infotext="It seems as if your system root is not on the SD card.
        ***Aborting, process cant be started***"
-    whiptail --title "System root not on SD card?" --msgbox "$infotext" 8 78
+    whiptail --title "System root not on SD card ?" --msgbox "$infotext" 8 78
```
ecdye, line 43: Why the extra space before the "?"?

functions/ext-storage.bash
```diff
@@ -40,10 +40,25 @@ move_root2usb() {
...
     return 0
   fi
 
+  #exit if destination is not available
+  if ! [[ -b "$NEWROOTPART" ]]; then
+    infotext="Seems there is no external SD card inserted that we could move openHABian root to.\\n***Aborting, process cant be started***"
```
ecdye, line 49: "external SD card" would be better said as "external storage medium"

functions/ext-storage.bash
```diff
@@ -40,10 +40,25 @@ move_root2usb() {
...
+    whiptail --title "No destination SD card ?" --msgbox "$infotext" 8 78
```
ecdye, line 50: extra space again

functions/ext-storage.bash
```diff
@@ -40,10 +40,25 @@ move_root2usb() {
...
+    return 0
+  fi
+
+  srcsize=$(blockdev --getsize64 /dev/mmcblk0)
+  destsize=$(blockdev --getsize64 "$NEWROOTDEV")
+  if [[ "$srcsize" -gt "$destsize" ]]; then
+    infotext="Your internal SD card is larger than the external device you want to move your root to. This will very likely break your system.\\n\\nDo you still REALLY want to continue ?"
```
ecdye, line 57: here to.

functions/ext-storage.bash
mstormi: my personal preference

functions/ext-storage.bash
mstormi: see up

functions/ext-storage.bash
mstormi: done

functions/ext-storage.bash
mstormi: done

PR #913, opened 2020-05-25
don't exit installation/openhabian-config on nonfatal errors

### LINKED ISSUE #702: Error handling in "firemotd_setup()" is not working
Rewrite the error handling in firemotd_setup() to not use the "FAILED" variable.
Instead check for error along when the script executes.

### LINKED ISSUE #899: Refactor code to not stop on package installation failures
Today this makes openhab-config or an unattended installation fail.
Code should continue rather than exit (routines to use `return` instead of `exit`).

mstormi:
replace exit by return so installation/openhabian-config keeps going

Fixes #702 
Fixes #899 

### DIFF
--- a/functions/ext-storage.bash
+++ b/functions/ext-storage.bash
@@ -52,6 +52,7 @@ move_root2usb() {
 
      echo
      echo "********************************************************************************"
+     # shellcheck disable=SC2086
      echo "REBOOT, run $(basename $0) again and recall menu item 'Move root to USB'"
      echo "********************************************************************************"
      whiptail --title "Reboot needed!" --msgbox "USB had to be set to high power (1A) first. Please REBOOT and RECALL this menu item." 15 78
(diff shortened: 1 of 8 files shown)


holgerfriedrich:
Hi Markus, did you check that we do not hide errors when someone is using openhabian-config to manually install or configure? It should be unrelated if you target only the install.
Just asking, did not look at the changes....

mstormi:
It's not perfect because the install routines are not transactionally safe ... but I believe it's not any different from today. If today on errors the script exits, it does not undo what it changed before the exit so in a worst case there's a change on disk that is persistent when you start openhabian-config again. We now just return to the menu instead.

A second opinion on this is highly appreciated :)

mstormi:
another one that should be in before tomorrow (@ecdye ? can's select you as reviewer you can only volunteer to so it seems)

ecdye:
> another one that should be in before tomorrow (@ecdye ? can't select you as reviewer you can only volunteer to so it seems)

I'm not sure why it does that, especially as I am a member of the openHAB organization. Maybe look at the repo permissions and see if you can figure that out?

mstormi:
pinging @ecdye can you review?

Issue #917, opened 2020-05-26
several BATS tests to fail


mstormi:
Merging #918 made BATS output appear in 2nd Travis test stage
That revealed several BATS tests to fail once you're willing to look at them :-|
See https://travis-ci.com/github/mstormi/openhabian/jobs/339942816

We need to fix:
- [x] destructive-install_zulu8-64bit
- [x] destructive-install_zulu11-64bit
- [x] destructive-install_zulu8-32bit
- [x] destructive-install_zulu11-32bit
- [x] destructive-install_adopt
- [x] installation-Frontail_is_running

- [x] destructive-homegear_install

- [x] destructive-influxDB_install 
- [x] destructive-grafana_install (pinging @holgerfriedrich)

- [ ] java tests - decide if we want to test openhab startup with this java version, then CI test needs to install openhab as well (and stop in teardown)

ecdye:
Its failing the Java installation test because it does not have a full openHABian image properly set up and booted. Basically that means that it fails because Java is not installed

mstormi:
Why would it need a full image ? The BATS test is supposed to be a unit test, i.e. it should install Java, start it up for a smallish test (java -version or the like) and that's it ...

holgerfriedrich:
@mstormi not easy to debug without any output.
How do we proceed here? Log file as artifact possible? Use $output?

mstormi:
Huh? Did you see #918 ?
Rebase on master and you should see your output ...

ecdye:
It doesn't because it is running in the bare docker image and the openHABian install is what installs Java, not docker.

mstormi:
@ecdye I don't understand at all what you mean. Please quote lines if you refer to some earlier post,
and don't jump between issues, that's highly confusing.

As replied [here](https://github.com/openhab/openhabian/pull/869#discussion_r430685837), the BATS test itself is to install Java.

holgerfriedrich:
> 
> 
> Huh? Didn't you see #918 ?
> Rebase on master and you should see your output ...

what I see in my tests is just 
```
not ok 1 destructive-influxDB_install
# (in test file ./functions/influxdb+grafana.bats, line 12)
#   `[ "$status" -eq 0 ]' failed with status 3
```

I can track this down to the command run before, but I do not see what happened.
E.g. grafana install failed with exit code. Good to know. But hard to debug.

mstormi:
I see. Line 12 is systemctl ...
[...]

PR #927, opened 2020-05-28
Debug output for bats tests and fix for Grafana installer

holgerfriedrich:
* add output for all failed tests
* add lsb-release, unzip to Docker image used for bats #917
* fixed influxdb+grafana installation to run properly in CI #917 #921

### DIFF
--- a/Dockerfile.openhabian-aarch64-native
+++ b/Dockerfile.openhabian-aarch64-native
@@ -12,7 +12,7 @@ ENV APT_KEY_DONT_WARN_ON_DANGEROUS_USAGE=DontWarn
 # bats tests on a minimal install will require additional packages to run properly:
 # - lsb-release (influx, homegear)
 # - psmisc, i.e. killall (homegear)
-# - unzip (jave)
+# - unzip (java)
 # those packages would normally be included in our standard install
 RUN apt-get update -qq \
     && apt-get install --yes -qq --no-install-recommends systemd systemd-sysv git wget python3 python3-pip apt-utils jq virt-what lsb-release unzip psmisc \
--- a/Dockerfile.openhabian-aarch64-native
+++ b/Dockerfile.openhabian-aarch64-native
@@ -8,8 +8,14 @@ ENV container docker
 ENV INITSYSTEM on
 ENV APT_KEY_DONT_WARN_ON_DANGEROUS_USAGE=DontWarn
 
+# Currently this config is used for BATS tests.
+# bats tests on a minimal install will require additional packages to run properly:
+# - lsb-release (influx, homegear)
+# - psmisc, i.e. killall (homegear)
+# - unzip (jave)
+# those packages would normally be included in our standard install
 RUN apt-get update -qq \
-    && apt-get install --yes -qq --no-install-recommends systemd systemd-sysv git wget python3 python3-pip apt-utils jq virt-what \
+    && apt-get install --yes -qq --no-install-recommends systemd systemd-sysv git wget python3 python3-pip apt-utils jq virt-what lsb-release unzip psmisc \
     && rm -rf /var/lib/apt/lists/*
 RUN git clone https://github.com/bats-core/bats-core.git && \
     cd bats-core && \
--- a/functions/packages.bats
+++ b/functions/packages.bats
@@ -3,12 +3,29 @@
 load packages
 load helpers
 
+
+teardown_file() {
+  echo -e "# \e[36mHomegear test cleanup..." >&3
+  systemctl stop homegear.service || true
+  # TODO: fix homegear shutdown handling, is does not properly shut down and will cause bats tests to hang
+  #ps aux|grep -i homeg >&3
+  # workaround in place:
+  killall -9 homegear || true
+}
+
 @test "destructive-homegear_install" {
   echo -e "# \e[36mHomegear installation starting..." >&3
   run homegear_setup
+  if [ "$status" -ne 0 ]; then echo "$output" >&3; fi
   [ "$status" -eq 0 ]
   echo -e "# \e[32mHomegear installation successful." >&3
-  run systemctl is-active --quiet homegear.service
+  # TODO: fix homegear service, currently it does not detect running instance in CI
+  #run systemctl status homegear.service
+  #echo "$output" >&3
+  #run systemctl is-active --quiet homegear.service
+  #if [ "$status" -ne 0 ]; then echo "$output" >&3; fi
+  # workaround in place: (pgrep would be better, but in doubt that it is installed in CI)
+  run bash -c "ps aux | grep homegear | grep -v grep"
   [ "$status" -eq 0 ]
   echo -e "# \e[32mHomegear service running." >&3
 }
--- a/functions/influxdb+grafana.bash
+++ b/functions/influxdb+grafana.bash
@@ -284,52 +284,73 @@ influxdb_install() {
 ##    influxdb_install(String admin_password)
 ##
 
+grafana_debug_info() {
+  echo
+  date
+  echo ---
+  cond_redirect tail -n40 /var/log/grafana/grafana.log | sed "s/^/DEBUG  /"
+  echo ---
+  local TMP
+  TMP=$(pgrep -a grafana)
+  # shellcheck disable=SC2001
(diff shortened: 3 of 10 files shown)


holgerfriedrich:
@mstormi getting closer, pls. take a look at the CI. Added output to all bats checks.

mstormi:
> take a look at the CI

what do you mean?

holgerfriedrich:
https://travis-ci.org/github/holgerfriedrich/openhabian/jobs/692049537 (hope you can open this, the PR build is still running)
https://travis-ci.com/github/openhab/openhabian/jobs/340822671

mstormi:
Ok but what doe you mean by your comment ? That you can now produce output in BATS tests to appear in CI logs and that we should look at your Grafana bats code to see how ?

holgerfriedrich:
I added it to all tests
```
  if [ "$status" -ne 0 ]; then echo "$output" >&3; fi
  [ "$status" -eq 0 ]
```

Now you can see what the other tests are complaining about, e.g.
```
# Zulu 64-bit Java installation is being (test-)installed...

# Zulu 64-bit Java installation successful.

$ systemctl stop openhab2.service 

ERROR:systemctl:Unit openhab2.service could not be found.

/opt/openhabian/./functions/java-jre.bash: line 65: timestamp: command not found
```

But still I cannot get Grafan install to start up the service properly. Will have to look at it again in the evening...

holgerfriedrich:
@ecdye any idea why it stall hangs after rebasing on #930?

ecdye:
I have my theories but I'll wait to make any statements until I have confirmed them on my end.

ecdye:
Ok, so I think it has to do with adding the packages directly to the dockerfiles. Rather install the packages in the script that you are testing. For example, instead of adding `software-properties-common` directly to the Dockerfile (as when I did this was when Bats would hang, don't ask why because I have no clue) I added this to my install script for AdoptOpenJDK (what required the `software-properties-common` package):
``` bash
if ! dpkg -s 'software-properties-common' > /dev/null 2>&1; then
    cond_redirect apt-get install --yes software-properties-common
fi
```
Then after discovering the error mentioned below relating to influx-db through the use of your debug command (very useful), and working around it, my latest build worked correctly without any hanging. See <https://travis-ci.com/github/openhab/openhabian/jobs/341218969> for the results.

As an aside, it appears that influx-db install is not correct for the `apt` lists as I had to work around it and that was why my Adopt test was failing. See <https://travis-ci.com/github/openhab/openhabian/jobs/341215297#L459> for what I mean. I have no idea if you already fixed this here but I just used a setup script in bats to clear the apt lists directory before running the Java tests and that solved the issue.

That is what worked for me, however, your mileage may vary. Let me know if what I said makes no sense and I can try to clarify.

mstormi:
> Then after discovering the error mentioned below relating to influx-db through the use of your debug command (very useful), and working around it, my latest build worked correctly without any hanging.

Then your builds were probably hung because of influx ? And you also fixed some ";" in influx install which made that going.
That now sounds more reasonable to be a reason than your move of a single package installation from Docker into the code.

ecdye:
Sort of APT was complaining about being unable to read the file relating infulx so I made sure it would be deleted when starting my install script to reduce the error sources.

holgerfriedrich:
@ecdye can you show where?
@ecdye influx apt problem due to missing lsb-release package, then apt line will be generated wrong. will add a check for this later.

I did a bit testing: adding bats debug output did not make the tests hang. adding the missing packages (unzip, lsb-release) both did (also separately), so I suppose it is related to the content of the tests which are then able to proceed.

I am aware of https://github.com/bats-core/bats-core#file-descriptor-3-read-this-if-bats-hangs, but I do not see where this could help.
What is the thing about missing ;  ???

ecdye:
Show where what? The apt fail I mentioned it earlier its here: https://travis-ci.com/github/openhab/openhabian/jobs/341215297#L459 

The workaround for bats I used is the setup function (bats provides this and it runs before the tests) in my case it is very simple
``` bash
setup() {
  rm -rf /etc/apt/sources.list.d/*
}
```
the ; missing was just a theory of mine as to why the tests were hanging but apparently it was wrong.

mstormi:
will you make it by tomorrow ? Would love to have BATS tests show green when we announce java beta test tomorrow but don't hurry, reviews and PR fixes for tomorrow are more important if in question.

mstormi:
closure intentional?

holgerfriedrich:
@mstormi will get this done soon. Bats tests were hanging once the installer worked.
Seems to be caused by services started by the test. Now added a teardown function.
Trying to finalize.

holgerfriedrich:
@mstormi what should we do about tests like this, which currently fail:
```
@test "installation-java_exist" {
  run java -version
  if [ "$status" -ne 0 ]; then echo "$output" >&3; fi
  [ "$status" -eq 0 ]
  [[ $output == *"Zulu"* ]]
}
```
Current output:
```
/usr/local/lib/bats-core/test_functions.bash: line 38: java: command not found
not ok 1 installation-java_exist
```

mstormi:
that particular one should be removed, does not make sense anyway

mstormi:
Fix frontail test while you're at it.
I consider this to be enough. Else to have the service running we would need java installed

[22:12:37] openhabian@openhabianpi:~/openhabian$ cat functions/nodejs-apps.bats
#!/usr/bin/env bats

load nodejs-apps
load helpers

@test "installation-Frontail_is_running" {
  run frontail_setup
  run frontail -V
  if [ "$status" -ne 0 ]; then echo "$output" >&3; fi
  [ "$status" -eq 0 ]

  echo -e "# \e[32mFrontail properly installed." >&3
}

mstormi:
you said you install lsb-release now ? That's also needed for the homegear test

holgerfriedrich:
frontail test fails, see https://travis-ci.org/github/holgerfriedrich/openhabian/jobs/693227967

and yes, need to install lsb-release.
currently trying to trace which test still causes bats to hang after completion.

mstormi:
did you fix the InfluxDB repo ? For me that caused the next bats test in line to fail on apt-update

holgerfriedrich:
Rebased, now all tests are hopefully green again.
Grafana install needs to be checked on real hardware and influxdb needs to be tested it it started after reboot.

holgerfriedrich:
@mstormi finally rebased and ready for review.
Just purged influx and grafana on my test install, reinstall worked fine and influx still works after reboot.

Dockerfile.openhabian-aarch64-native
```diff
@@ -9,7 +9,7 @@ ENV INITSYSTEM on
 ENV APT_KEY_DONT_WARN_ON_DANGEROUS_USAGE=DontWarn
 
 RUN apt-get update -qq \
-    && apt-get install --yes -qq --no-install-recommends systemd systemd-sysv git wget python3 python3-pip apt-utils jq virt-what \
+    && apt-get install --yes -qq --no-install-recommends systemd systemd-sysv git wget python3 python3-pip apt-utils jq virt-what lsb-release unzip psmisc \
```
mstormi, line 12: Maybe create a separate Dockerfile for BATS, else we'll forget why this line is different across Dockerfiles.

functions/packages.bats
```diff
@@ -3,12 +3,29 @@
...
   [ "$status" -eq 0 ]
   echo -e "# \e[32mHomegear installation successful." >&3
-  run systemctl is-active --quiet homegear.service
+  # TODO: fix homegear service, currently it does not detect running instance in CI
+  #run systemctl status homegear.service
+  #echo "$output" >&3
+  #run systemctl is-active --quiet homegear.service
+  #if [ "$status" -ne 0 ]; then echo "$output" >&3; fi
+  # workaround in place:
+  run bash -c "ps aux | grep homegear"
```
mstormi, line 28: will always find the grep process so no good test

holgerfriedrich: right, now fixed

Dockerfile.openhabian-aarch64-native
holgerfriedrich: Added a comment block, IMHO better than creating another config.
Maybe we should silently extend the installed packages also in all other config files.

Issue #912, opened 2020-05-24
WiFi gone after HW upgrade RPi3 -> RPi4


holgerfriedrich:
Tried to upgrade my hardware from RPi3b to RPi4 _without a fresh install_.
WiFi was enabled and working before, LAN not connected.
System booted fine, but WiFi function was gone. Reboot did not help.

Since this is a typical upgrade situation, we should have a look if we can improve the behaviour.

ecdye:
Did you have WiFi enabled on your RPi3b when you upgraded? If so, perhaps it just is a simple fix of checking the HW on boot and adjusting the settings for network if needed.

mstormi:
While I disagree formally (you SHOULD install from scratch and migrate config only), you are correct that it would be a good idea if that worked at least in the most common setup (boot from SD). We should not aim to support external SSD, net boot or the like. 

Could you please review #861 so I can merge that asap, it has some changes to affect WiFi.

Is there anything about Wifi in your /boot/config.txt ?

mstormi:
Just booted from a virgin image and noticed some startup messages about a
"Load/Save RF Kill Switch Status" right below "Starting Blueetooth service".
Probably worth to check out.

mstormi:
Not sure if that's related but see https://github.com/openhab/openhabian/pull/931#issuecomment-641815609

mstormi:
Holger can you retry with your new image and eventually use menu option 33 ?

mstormi:
@holgerfriedrich ping

mstormi:
Just upgraded the 'enclosure' of my SD card from a PRi2 to a RPi4 and WiFi is still there (with the default config in /etc/network/interfaces).

Let's close this one. You can reopen should you hit that again.

Issue #825, opened 2020-04-03
Offline install


mstormi:
Is openHABian install supposed to finish even when Internet access isn't available ?

At least a number of users expect it to. I myself, frankly, I don't know. @ThomDietrich ?
We have a number of reported install fails at
https://github.com/openhab/openhabian/blob/master/functions/openhab.bash#L60

While not fully analyzed yet, it all boils down to the introductory question.
If yes, we would also need to test install runs under this condition.
If yes, we would need to identify the basicmost packages we need to get openHABian+openHAB running and include them with the image:
That would at least be needed_packages() - probably only python - PLUS Java PLUS openHAB packages (current release).

holgerfriedrich:
@mstormi This would be a completely differnt approach compared to what we do now.
Currently, we have only sporadic releases of the base image and connect to the internet to upgrade the base system as well as openhabian and openhab. Latest is greatest approach. Following an offline approach for installation would require - at least from my point of view - regular releases of the base image including all updates up to that time. Jenkins/Travis could do this, but.... Considering the effort, I would vote against this. Let us deliver what we have and focus on stability. Developer community is too small for this.

Running openhabian-config without network is something different, at least for the part which changes some settings. Some people decide not to connect the home server to the internet and want it this way. So maybe the start of the script could skip the upgrade if not connected to the internet. Though, not sure how helpful this is at all. Nearly all functions are to install additional features, which is impossible withhout adding everything to the base image. I would suggest not to add error handling to the functions, maybe just a disclaimer/reminder when openhabian-config is started without internet connection.

mstormi:
It's really meant as an question in the first place. _Was_ it supposed to work like that ?

mstormi:
I'd like to move this forward. Aiming for an offline install capable image.

Components we would need to validate / enable caching or other pre-storage method for
(cast "in order of appearance"):

FireMotD
Java see #869 
openHAB2 stable release
vimrc, nanorc
Samba (is there a version included with Raspbian already ?)
Frontail

holgerfriedrich:
Really serious about this?

From my point of view this would require executing the first-boot script before shipping the image.
Otherwise we will end up with two different types of installations with different sets of installed packages.

mstormi:
Granted I have not analyzed what this would mean in detail but yes I'm serious about this because

> This could help in attracting users (or more precisely, to not lose them on first contact) because install fails.

[...]

PR #931, opened 2020-05-29
Introduction of Raspberry Pi OS

holgerfriedrich:
* Replace raspbian by raspios due to renaming of the OS
* (beta) Added initial support for building 64-bit images

Resolves: #946


ecdye:
RFC: Do we want to consider renaming the project, or do we keep our name and move on?

build-image/openhabian.pi-raspbian64.conf
```diff
@@ -0,0 +1,58 @@
...
+# values: "enable", "disable"
+ipv6=enable
+ 
+# repo to clone from (if not master for build and test)
+repositoryurl=https://github.com/openhab/openhabian.git
+clonebranch=master
+
+# debug verbosity mode
+# values: "unattended", "unattended_debug" (verbose output in log) or "debug_maximum" (show every command in either mode)
+mode=unattended
```
mstormi, line 38: called debugmode now

You need to rebase on master

mstormi:
I'm just enjoying the "offline installation" use case #825 and while it's pretty hard to develop with (almost) no Internet access, this helped me identify the parts of the code we will need to rework for offline install to work (I hope, hard to test in this situation)

Will upload a PR when I'm back online. Meanwhile below the relevant code for you to include.
Can you insert that plus calls to `install_zram_code` and `java_zulu_fetch Zulu8-32` into build.bash ?

That should pre-load the code versions that are current at the time of image build. Our current java and my new ZRAM code should work to install these from storage instead of Inet if Inet is down.
Make sure that `/opt/zram` and `/opt/jdk` are also included with the image.
```

install_zram_code() {
  mkdir -p "$1"
  /usr/bin/git clone -q --branch "$TAG" "$ZRAMGIT" "$1"

  cond_redirect apt-get install -y -q --no-install-recommends make
  cd "$1" || return 1
  /bin/sh ./install.sh
}

ZRamInstallLocation=/opt/zram
install_zram_code "$ZRamInstallLocation"

```

holgerfriedrich:
@mstormi .
- .xz.zip is caused by GitHub, all artifacts will be zipped during download. There seems no way to avoid it right now, but others are complaining about this as well. If we put the img file in the archive, you will have to wait for the img file to be compressed on the server side, this takes some time until the download starts.
- 64bit image is still experimental, there is not "lite" version available. GUI is now disabled by patching the standard systemd target.
- pulling stuff in build.bash - no problem. BUT: compiling! This would require cross-compiling for the ARM platform (since I run the build on x86). Have you thought about creating deb packages for the zram build, this would ease the handling a lot?

Noticed some problems with during install, webif did not work and zulu8 64bit did not work. Trying the next build...

mstormi:
>     * Have you thought about creating deb packages for the zram build, this would ease the handling a lot?

Yes but I want to avoid that. Just like the original zram author I'm no expert at building packages. And the compilation is not even for the ZRAM code itself but for a single tool (overlayfstools) which is in turn YET another upstream. That I don't want to mess with.
Then again you just made me aware that we would need to pre-fetch overlayfstools, too, if that's supposed to install offline. .... so probably I moved that off the zram install routine.
But you need not compile it when building the image. Just call the fetch routine. With a copy of the repo in place, compilation will only take place at _installation_ time (see Travis logs).

PS: mein Internet soll frühestens Dienstag wieder funktionieren ... seufz, bis dahin ist das alles etwas sehr eingeschränkt.

mstormi:
did you set the number of tryUntil() attempts on "ping 9.9.9.9|| wget ...." to 60 from 10 ?
10 is way enough (every wget does retries already)

mstormi:
Also remember to replace names in the docs

holgerfriedrich:
@mstormi @ecdye willing to play around with the new 64bit image?
I just installed successfully on my Pi3b. A few points to clarify:
- webif not tests, did not seem to work during last install
- azul8-64 did not work with OH, complained about library issue. Switched to azul11-64 in latest image
- graphical splash screen of the new OS still pops up during install before showing the command prompt
- is_rapbian to be checked if it still works

@mstormi yes, more retries on slow WLAN APs....
@mstormi doc update done on pr branch, not applied to current image

holgerfriedrich:
> Yes but I want to avoid that. Just like the original zram author I'm no expert at building packages. And the compilation is not even for the ZRAM code itself but for a single tool (overlayfstools) which is in turn YET another upstream. That I don't want to mess with.

An alternative approach to cross compiling is cache the two git repos, and start compilation on the PI during installer run (first-boot.bash). Any special compile flags derived from the HW platform you compile on (e.g. optimizations only suitable for Pi3)?

mstormi:
> An alternative approach to cross compiling is cache the two git repos, and start compilation on the PI during installer run (first-boot.bash).

Yes that's what I'm out for. Check out #950. Can you call install_zram_code() during build to have the repo "cached", and include /opt/zram with the image ?

> Any special compile flags derived from the HW platform you compile on (e.g. optimizations only suitable for Pi3)?

no, dead simple

mstormi:
> willing to play around with the new 64bit image?

where is it ? Link in #946 is still the old one, boots into desktop

mstormi:
I just interrupted install but on reboot it fails to continue on "apt-get --yes upgrade"
Manual executing of that resulted in "E: dpkg was interrupted, you must manually run 'sudo dpkg --configure -a' to correct the problem."
So while of course a user should not interrupt, and this problem probably is not specific to the new image, some users do like I did, and we should take care of that.

holgerfriedrich:
@mstormi, sorry pls check latest one at

https://github.com/holgerfriedrich/openhabian/actions?query=workflow%3Abuild-image

mstormi:
BTW I have added #950 (offline install enhancements) to the "testbuild" branch, so why not test that with Internet blocked.

Just did and it does not work because /boot/first-boot.bash (which is run) is different from /opt/openhabian/build-image/first-boot.bash (which is newer).
How's that? Is that copied on image build ?

holgerfriedrich:
@mstormi yes, build.bash is patching the file - but for RPi it is only adding the content of helpers.bash:
`sed -i -e '1r functions/helpers.bash' $buildfolder/boot/first-boot.bash`

Regarding the offline install: I would prefer to do things separately. Currently, I already merge 3 PRs + fixes to my master branch to build the image. I think we could start merging #890, for me the new installation worked fine in all of my tests. It will anyway only be rolled out when we release the new image.

mstormi:
>  yes, build.bash is patching the file -but for RPi it is only adding the content of helpers.bash:

Then why is /boot/first-boot.bash so different from build-image/first-boot.bash ?

Re: offline install: ok, finish image build first, then we'll review what's left missing for offline install

Did you mean to say #890 is ready to be reviewed and merged ?

holgerfriedrich:
> Just did and it does not work because /boot/first-boot.bash (which is run) is different from /opt/openhabian/build-image/first-boot.bash (which is newer with my changes).
> How's that? Is that copied on image build ?

Did you also apply the changes from this PR? It is mainly changing build.bash...?

mstormi:
> Did you also apply the changes from this PR? It is mainly changing build.bash...?

No. As far as I can see it does not change first-boot.bash or anything related to copying or creating that.
I admit I've never built an image or changed anything about build.bash so am unaware of how it works.

/boot/first-boot.bash applies BEFORE git retrieves the current openhabian repo. So it is as it is because it was built like that on image creation. _My_ changes cannot change that (they change the copy inside build-image but not the one in /boot).

Could you apply all of your image-build related changes (you spoke of 3 PRs but I don't know what these are) to the testbuild branch ? That's nothing other than today's master plus #950 . Thanks!

mstormi:
FWIW, I tried the 64 bit image on my brand new RPi4. Note the last two lines of the output.
(I did nothing other than install via Ethernet).
Possibly related to any of the open WiFi issues such as #912 ?

```
openhabian@192.168.178.61's password:
Linux openhab 5.4.42-v8+ #1319 SMP PREEMPT Wed May 20 14:18:56 BST 2020 aarch64

The programs included with the Debian GNU/Linux system are free software;
the exact distribution terms for each program are described in the
individual files in /usr/share/doc/*/copyright.

Debian GNU/Linux comes with ABSOLUTELY NO WARRANTY, to the extent
permitted by applicable law.
Last login: Tue Jun  9 17:59:22 2020

Wi-Fi is currently blocked by rfkill.
Use raspi-config to set the country before use.

```

mstormi:
By the way, many of your commits have an invalid signature, please correct. it
see https://github.com/openhab/openhabian/pull/953/checks

mstormi:
> @mstormi yes, build.bash is patching the file - but for RPi it is only adding the content of helpers.bash:

I saw this results in the first TWO lines to start with #!.

mstormi:
We also need these packages to be there offline:
python3-pip python3-wheel python3-setuptools 
(and its dependencies)

mstormi:
Hi Holger,
could you review #961 please and build another image to include that ?
What about the promised hook from https://github.com/openhab/openhabian/pull/950#issuecomment-642290136 for all those packages we need to have available on offline installs ?

holgerfriedrich:
@mstormi I think we should go step by step. This changes basically only the download loacation and adds basic support for the 64bit beta image. The introduction of 64bit image will cause more issues (especially in external components) and should be handled in a separate issue. WDYT?

holgerfriedrich:
@mstormi @ecdye new images with latest commits from today are available

build-image/first-boot.bash
```diff
@@ -18,7 +18,9 @@ fail_inprogress() {
   exit 1
 }
 
-echo "$(timestamp) [openHABian] Starting the openHABian initial setup."
+###### start ######
+sleep 5
```
ecdye, line 22: Why add a delay before start?

build-image/openhabian.pi-raspios64beta.conf
```diff
@@ -0,0 +1,59 @@
...
+
+# OS Release
+# ubuntu, raspbian, debian
+# stretch, buster
+# bionic, focal
+#release=buster
+
+# Java architecture mode
+# Valid arguments: "Adopt11", "Zulu8-32", "Zulu8-64", "Zulu11-32", or "Zulu11-64"
+java_opt=Zulu11-64
```
ecdye, line 57: Java 11 default? Are you planning on this being the openHAB 3 recommended setup for a RPi3+?

build.bash
```diff
@@ -265,7 +269,17 @@ if [ "$hw_platform" == "pine64-xenial" ]; then
   if mv $buildfolder/xenial-pine64-*.img $imagefile; then echo "OK"; else echo "FAILED"; exit 1; fi
 
 # Build Raspberry Pi image
-elif [ "$hw_platform" == "pi-raspbian" ]; then
+elif [ "$hw_platform" == "pi-raspios32" ] || [ "$hw_platform" == "pi-raspios64beta" ]; then
+  if [ "$hw_platform" == "pi-raspios64beta" ]; then
+    zipfile=raspios_lite_arm64_latest.zip
+    baseurl="https://downloads.raspberrypi.org/raspios_arm64/images/raspios_arm64-2020-05-28/2020-05-27-raspios-buster-arm64.zip"
```
ecdye, line 275: You should add a note to change the link to a permanent redirect to the latest version when it becomes available.

build-image/first-boot.bash
holgerfriedrich: systemd might start more services, this improves the output

build-image/openhabian.pi-raspios64beta.conf
holgerfriedrich: See above, zulu8-64 did not work in my last test. This might have changed in the meantime, since java installation improved since then. Not tested yet.

build.bash
holgerfriedrich: sure. just the one beta image available as of now.

mstormi:
> The introduction of 64bit image will cause more issues (especially in external components) and should be handled in a separate issue. WDYT?

I'm fine to start with the 32bit image only.
We could offer the 64bit one marked experimental/unsupported. Or not at all for the time being ?

holgerfriedrich:
I would out it into the code base but not explicitly mention it in the documentation.

build.bash
```diff
@@ -265,7 +269,17 @@ if [ "$hw_platform" == "pine64-xenial" ]; then
...
+    bits=64
+  else
+    zipfile=raspios_lite_armhf_latest.zip
+    baseurl="https://downloads.raspberrypi.org/raspios_lite_armhf_latest"
+    bits=32
+  fi
+
   # Prerequisites
   echo_process "Checking prerequisites... "
   REQ_COMMANDS="git wget unzip crc32 dos2unix xz"
```
mstormi, line 285: is that my hook for offline? If so add libattr1-dev

build-image/first-boot.bash
ecdye: OK

PR #935, opened 2020-05-31
Patch day 2020-05-31 master => stable

mstormi:
Forwarding patches as of 2020/05/31 from master to stable branch


ecdye:
Is there a way to turn of DCO for these patch requests?

mstormi:
yes but DCO is not the problem here
Need to check tomorrow. It's 3am over here really need to go to bed now. Will be back in ~10hrs

ecdye:
> yes but DCO is not the problem here

 I know that, I was just making an observation. You may have to resort to manually making the merge by hand to fix the conflicts.

PR #941, opened 2020-06-01
Grafana installer to call grafana-cli as user grafana

holgerfriedrich:
* resolve permission problems which caused Grafana install issue #921


holgerfriedrich:
(small patch exception)

PR #947, opened 2020-06-02
DRAFT replace nrjavaserial by reversioned 5.2.1

mstormi:
Replace nrjavaserial 3.15.0 as used in OH2.5 frozen core with a reversioned 5.2.1 bundle
from https://github.com/wborn/nrjavaserial to fix issues with serial interfaces on Java 11


bodiroga:
Hi @mstormi!

I'm not sure this should be done, as nrjavaserial 5.2.1 has some breaking changes and perhaps OH2.5 bindings do not support them (AFAIK, ZWave 2.5.x binding does not work correctly with nrjavaserial 5.2.1).

What do you think @wborn?

mstormi:
Wouter already made me aware of that. Also, there seems to be no need to patch anything in Linux so will retreat this PR.

PR #950, opened 2020-06-08
Offline installation

mstormi:
WIP


holgerfriedrich:
@mstormi did you consider that the base image might not contain everything you need to compile the zram packages?

My requirements would be
- no cross compilation during image build.
- no package installation during build (cannot call apt of the target image)
- separate what is to be done during image build (x86) and what during first-boot (arm/arm64)

 If really needed, we could separate the compilation step out and build deb packes for zram separately. Much simpler would be to clone on build and compile during first-boot script.

For req 2 we might need to cache some Deb packages (but sure what is missing, might be a set of dependencies required)

mstormi:
I'm targeting to (git-)_download_ the (source only) code on image build and to compile during installation. That would satisfy your requirements I guess.

Compiler tools are part the big (non-lite) image, checking next if they also are for the small one.

PR #961, opened 2020-06-11
more offline install fixes

### LINKED ISSUE #313: Reverse proxy on port 80 as default component
The nginx reverse proxy is currently missing in the optional components menu

### LINKED ISSUE #825: Offline install

mstormi:
Fixes #313 
Fixes #825 

### DIFF
--- a/build-image/first-boot.bash
+++ b/build-image/first-boot.bash
@@ -147,7 +147,7 @@ if tryUntil "ping -c1 9.9.9.9 >/dev/null || wget -S -t 3 --waitretry=4 http://ww
 echo "OK"
 
 echo -n "$(timestamp) [openHABian] Waiting for dpkg/apt to get ready... "
-if tryUntil "apt-get update &>/dev/null" 10 0; then echo "OK"; else echo "FAILED"; fi
+if tryUntil "apt-get update &>/dev/null" 3 1; then echo "OK"; else echo "FAILED"; fi
 echo "OK"
 
 echo -n "$(timestamp) [openHABian] Updating repositories and upgrading installed packages... "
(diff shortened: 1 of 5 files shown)


build-image/first-boot.bash
```diff
@@ -161,10 +162,11 @@ if ! cond_redirect dpkg -s "git" &>/dev/null; then
   if apt-get install --yes git &>/dev/null; then echo "OK"; else echo "FAILED"; fail_inprogress; fi
 fi
 
-if [ -d /opt/openhabian ]; then cd /opt && rm -rf /opt/openhabian; fi
+# must not remove for offline to work
+#if [ -d /opt/openhabian ]; then cd /opt && rm -rf /opt/openhabian; fi
```
holgerfriedrich, line 166: I suppose that removing the delete will cause `git clone` command to fail....
```
fatal: destination path '/opt/openhabian already' exists and is not an empty directory.
```

mstormi: right, good you spotted that.
How about this now ?

mstormi:
> Did you test?

Unit testing only.
Will try a full install on my test box before merging.

holgerfriedrich:
@mstormi are you going tomerge this after the tests have completed? I would build another test image later today.

mstormi:
Successfully tested that it does not break installation with Inet available.
Tested with Inet blocked, too, found and fixed a couple of locations that offline install choked on.
Now install successfully finishes, too.
There's some tiles missing on the resulting dashboard but OH is running.

On a sidenote, @holgerfriedrich @ecdye does anyone know a HTML trick to have a tail -f like display of the web install log like in Travis rather than to reload the page every n seconds ?

PR #962, opened 2020-06-11
continue install when broken due to interruption

### LINKED ISSUE #952: Interrupting unattended installation can break it
I just interrupted an unattended install but on reboot it fails to continue on "apt-get --yes upgrade"
Manual execution of that resulted in "E: dpkg was interrupted, you must manually run 'sudo dpkg --configure -a' to correct the problem."
After I did and rebooted, installation continued.
So while of course a user should not interrupt, and this problem probably is not specific to the new image, some users do like I did, and we should take care of that, i.e run dpkg --configure -a if needed.

mstormi:
Fixes #952 

### DIFF
--- a/build-image/first-boot.bash
+++ b/build-image/first-boot.bash
@@ -151,8 +151,14 @@ sleep 10  # Related to: https://github.com/openhab/openhabian/issues/441#issueco
 echo "OK"
 
 echo -n "$(timestamp) [openHABian] Updating repositories and upgrading installed packages... "
-apt-get --yes --fix-broken install &>/dev/null
-if apt-get --yes upgrade &>/dev/null; then echo "OK"; else echo "FAILED"; fail_inprogress; fi
+apt --yes --fix-broken install &>/dev/null
+apt-get --yes upgrade &>/dev/null
+res=$?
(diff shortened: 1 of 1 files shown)


build-image/first-boot.bash
```diff
@@ -151,8 +151,14 @@ sleep 10  # Related to: https://github.com/openhab/openhabian/issues/441#issueco
...
-apt-get --yes --fix-broken install &>/dev/null
-if apt-get --yes upgrade &>/dev/null; then echo "OK"; else echo "FAILED"; fail_inprogress; fi
+apt --yes --fix-broken install &>/dev/null
+apt-get --yes upgrade &>/dev/null
+res=$?
+if [[ $res -eq 100 ]]; then 
+  dpkg --configure -a
+  apt --yes --fix-broken install &>/dev/null
+  if apt-get --yes upgrade &>/dev/null; then echo "OK"; else echo "FAILED"; fi
+fi
```
holgerfriedrich, line 161: approach itself looks good.

`else echo "OK"` for the normal use case

mstormi: like so?
Rest ok?

PR #966, opened 2020-06-14
do not run apt-get update in background on unattended install

### LINKED ISSUE #965: [openHABian] Updating Linux package information ...   FAILED
on unattended install in Travis. Let's find and eliminate the reason.

mstormi:
Fixes #965 

I don't remember why it was done like this - anyone ?

### DIFF
--- a/Dockerfile.openhabian-aarch64-64bit
+++ b/Dockerfile.openhabian-aarch64-64bit
@@ -1,7 +1,7 @@
 # Note RPi3 base (first RPi to have aarch64 available) WITH 64bit OS
 #FROM balenalib/raspberrypi3-debian-node:latest-buster
 #FROM balenalib/generic-aarch64-debian-node
-FROM balenalib/raspberrypi4-64-debian-node-build
+FROM balenalib/raspberrypi4-64-debian-node:latest-buster-build
 
 
 ENV container docker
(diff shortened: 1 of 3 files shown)


ecdye:
Not sure

Dockerfile.openhabian-aarch64-64bit
```diff
@@ -1,7 +1,7 @@
 # Note RPi3 base (first RPi to have aarch64 available) WITH 64bit OS
 #FROM balenalib/raspberrypi3-debian-node:latest-buster
 #FROM balenalib/generic-aarch64-debian-node
-FROM balenalib/raspberrypi4-64-debian-node-build
+FROM balenalib/raspberrypi4-64-debian-node:latest-buster-build
```
holgerfriedrich, line 4: Never know why you change between different images for the CI. Seems unrelated to the topic of the PR...
But still ok for me.

mstormi: Yes this is unrelated Sorry I should have separated that. This was meant to just be a test because one CI test in this PR continously failed because of that, then again now that it works I'll put it in.

mstormi:
> I don't know if we are consitent here.

That was my thinking, too. It was an inconsistency causing delays and annoying messaging so I wanted to remove it. it was me who has put it there in #724. I just could not remember why I placed it there in the first place, though.
Then again I tested menu option 33 with `apt-get update` running in background which is the only use case I believe and that worked fine.

PR #938, opened 2020-06-01
Begin Standardization and Rewrite

ecdye:
Begins a rewrite of the openHABian core code to make it overall safer
and more verbose in failure. Additionally this helps to standardize the
coding format.

This contains major a major rewrite of 'packages.bash',
'nodejs-apps.bash', and other minor fixes in other files as well. This
also contains the proper testing functions using BATS for the rewritten
code.

Fixes #936
Related #937 #949
Fixes #856 
Fixes #633 

Signed-off-by: Ethan Dye <mrtops03@gmail.com>

PR #968, opened 2020-06-15
remove wlan0 config lines from /etc/network/interfaces on wifi disable

### LINKED ISSUE #608: openhabian-config disables eth0 interface while configuring WIFI
Hi,

I have a couple 1.4.1 installs on raspberry pi 3b v1.2. I wanted to get one on wifi instead of ethernet (for my camper) and running openhabian-config to configure wifi seems to have disabled the ethernet port. I suspect I could replicate it. 

I can work-around the issue but figured I'd let you know here as well.

mstormi:
Fixes #608 

I think the proper way to handle #608 is to enable Eth (X)OR WiFi exclusively.
This PR to remove wlan0 config from /etc/network/interfaces on disabling WiFi so that should remain empty hence Ethernet gets reenabled.

### DIFF
--- a/functions/system.bash
+++ b/functions/system.bash
@@ -276,8 +276,8 @@ enable_rpi_audio() {
 }
 
 prepare_serial_port() {
-  introtext="Proceeding with this routine, the serial console normally provided by a Raspberry Pi can be disabled for the sake of a usable serial port. The provided port can henceforth be used by devices like Razberry, UZB or Busware SCC.
-On a Raspberry Pi 3 the Bluetooth module can be disabled, ensuring the operation of a RaZberry or other HAT (usage of BT and HATs to use serial is mutually exclusive).
+  introtext="Proceeding with this routine, the serial console normally provided by a Raspberry Pi can be disabled for the sake of a usable serial port. The provided port can henceforth be used by devices like RaZberry, UZB or Busware SCC.
+On a Raspberry Pi 3 and 4 the Bluetooth module should be disabled to ensure the operation of a RaZberry or other HAT. Usage of BT and HATs to use serial is mutually exclusive.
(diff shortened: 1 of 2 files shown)


mstormi:
@ecdye @holgerfriedrich any idea why the "OK\n" is missing in https://travis-ci.com/github/mstormi/openhabian/jobs/349633271#L456 ?

ecdye:
I fixed that in #938 as well. Java output has been working for me in #938 as well.

mstormi:
> 
> 
> I fixed that in #938 as well. 

did you? that lines to output  'Installing git' in your repo look the same as mine.
And can you explain why the OK is missing ?

> Java output has been working for me in #938 as well.

By "no Java output" I meant there's no echo line stating Java got installed. Just for openHAB and lesser important components.

ecdye:
At the end of the install, Java should output its version info indicating it was installed.

Let me look into the installing git piece. I thought that I fixed it.

ecdye:
Travis shows the OK I did fix it, did you look at the fixes-2 branch? https://travis-ci.com/github/openhab/openhabian/jobs/349920702#L432

mstormi:
> 
> 
> Travis shows the OK I did fix it, did you look at the fixes-2 branch? https://travis-ci.com/github/openhab/openhabian/jobs/349920702#L432

Yes. The difference is in https://github.com/ecdye/openhabian/blob/fixes-2/build-image/first-boot.bash#L170 you print another OK when the outer if fails. In https://github.com/mstormi/openhabian/blob/disablewifi/build-image/first-boot.bash#L168 I don't. I was so convinced that the outer diff would apply, then I was wondering why the inner if did not produce neither OK nor FAILED. Now that's alright. Let's take your fix.

Issue #984, opened 2020-06-18
Prefetch code and packages for offline install


holgerfriedrich:
Refs: #825 

For an "offline install", i.e. first boot of the openHABian image without internet access, we need to cache some code and packages during creation of the image.
I.e. cache ZRAM code repos, cache JAVA, cache essential packages for building ZRAM and all the dependencies.

As this is done during image creation, it cannot use the built-in mechanisms of the target image (apt, ect).

```
install_zram_code $buildfolder/root/opt/zram
java_zulu_fetch $buildfolder/root
...
```

List what is required including dependencies, _only if_ not included in the Raspberry Pi OS lite image (make, gcc, git is fine....).
- libattr1-dev

holgerfriedrich:
@mstormi zram did not work, we have an apt install in install_zram_code

we need to download the deb files manually from correct respian repo and install them during first-boot :-/
This is why I did not like the idea of offline install in the first place.....

zulu java seems to work, azul has the same problem as zram but even worse....

mstormi:
> we have an apt install in install_zram_code

which is just the libattr1-dev that you've asked for (what to install upfront) 

> zulu java seems to work, azul has the same problem as zram but even worse....

what do you mean ? it works or not ?

mstormi:
wrong click

holgerfriedrich:
@mstormi  We have to distinguish between build system (the one which executes build.bash)  and target image. When I call `install_zram_code`, the library gets installed on the build system.
I have not way of executing apt of the mounted target image during install.
A simple but ugly way is to wget the correct deb file in `install_zram_code` and install it during first-boot (I was thinking to create a cache folder and installing all packages during first boot...)

mstormi:
I think @ecdye worked around exactly that for Zulu, check his code

ecdye:
No, for Zulu we use the API so we then avoid APT altogether. Honestly, I don't think that an elegant solution for offline install is possible. We may be able to fail gracefully and leave the user with a partial system but not full one. 

Zulu offline may not even truly work because the install still has apt packages that it requires as prerequisites.

mstormi:
> Zulu offline may not even truly work because the install still has apt packages that it requires as prerequisites.

which ones? That's Holger's intention of this issue to collect these.

mstormi:
there's an 'offline' competitor now: https://ubuntu.com/appliance/openhab

holgerfriedrich:
[...]

Issue #665, opened 2019-08-11
User openhabian not created due to failure to detect RPi model


massimilianocasini:
Hi,

in this image can be an error because the user openhabian is not created. There is only pi user.
There are more error on install log.

Can you check?

mstormi:
This is intentional. What error do you think this creates?

massimilianocasini:
A more scripts during first install to refer to openhabian user.
In documentation there is not any refer to PI user.
So, the openhabian user will be not used?

Il lun 12 ago 2019, 00:29 Markus Storm <notifications@github.com> ha
scritto:

EliasGabrielsson:
> This is intentional. What error do you think this creates?

@mstormi This is _not_ correct. The `openhabian` user shall be created.

@massimilianocasini Hard to say whats went wrong with your installation. Could you try to reinstall it again and see if it works? (Reflashing your image from scratch.)

massimilianocasini:
@elias, I have just tried with the same result. Everyone can see right away
that they can't log in during the first boot to see the log /boot/firstboot
with user openhabian as described.



Il giorno sab 17 ago 2019 alle ore 12:46 Elias Gabrielsson <
notifications@github.com> ha scritto:

massimilianocasini:
There are a more errors during the first boot:

2019-08-17_16:22:56_CEST [openHABian] Adding slightly tuned bash config
files to system... cp: cannot create regular file
'/home/openhabian/.bash_profile': No such file or directory
chown: invalid user: ‘openhabian:openhabian’
OK
2019-08-17_16:22:56_CEST [openHABian] Adding slightly tuned vim config file
to system... OK
2019-08-17_16:22:56_CEST [openHABian] Downloading and setting up
FireMotD... grep: /home/openhabian/.bash_profile: No such file or directory
/opt/openhabian/functions/packages.bash: line 37:
/home/openhabian/.bash_profile: No such file or directory
OK 2019-08-17_16:23:58_CEST [openHABian] Installing Zulu Embedded
OpenJDK... OK
2019-08-17_16:25:16_CEST [openHABian] Installing or upgrading to latest
openHAB release (stable)... OK
2019-08-17_16:26:20_CEST [openHABian] Adding an openHAB dashboard tile for
'openhabiandocs'... OK
2019-08-17_16:26:21_CEST [openHABian] Adding openHAB syntax to vim
editor... chown: invalid user: ‘openhabian:openhabian’
OK
[...]

PR #985, opened 2020-06-18
Caching essential packages for offline install, #984.

holgerfriedrich:
* initial functionality added to build.bash
* directory prefix for java_zulu_fetch


mstormi:
Just tried. Seems it's missing git so it's quickly skipping a major part of installation
Yeah I also thought it's there but it is not. Could you preinstall that ?


Could you also please add copying helpers.bash and openhabian.bash to /boot ?
Then also remove the line to paste helpers.bash into /boot/first-boot.bash

ecdye:
Several of those fixes are included in #1008

mstormi:
We really should get the new image out asap. See https://github.com/openhab/openhabian/issues/665#issuecomment-653270634
that reminds me that first-boot.bash is still that from the time the image was created, including all bugs fixed since such as support of some RPi models.

#1008 now fixes what I wrote in my last comment, and IIRC you said prefetching works now, too ?
Maybe change current PR to store prefetched packages to install at installation time in /var/cache/apt/archives, that's more elegant IMHO.

Please let's finalze this PR this weekend and get the image out.

holgerfriedrich:
@mstormi latest image build considers your comment regarding git. Not sure if it works on real HW. We may need to shift the code which installs the packages since git seems to be used before it is installed (maybe not a problem at all, because it won't be able to clone anyway in offline mode)
I did not have time to check this out carefully this week. Same for the dependencies which might be missing.

mstormi:
are you ok with #1008, can I merge that? Please build another image when that's in so we can test on real HW.

mstormi:
My recent attempt with the latest image failed because apt/dpkg complained about a lock when "Updating repositories and upgrading installed packages". I'm not sure it is the only reason but I believe it is because the last action of standard bootup is a systemd timer unit to handle apt cleanup stuff, and this was overlapping with the first apt actions in our install scripts.
So #721 is not just a cosmetic issue.
As you already correctly said a proper fix (rather than delay) would be to use systemd After= setting in our install unit file.
I believe the last timer (it's a systemd unit as other are, too) which is also the last bootup action to take place (not sure if always, that's possibly depending on HW speed and some coincidence) is `apt-daily-upgrade`.

I'll add `After=apt-daily-upgrade` to `/etc/systemd/system/openhabian-installer.service`.

mstormi:
So now with 1008 merged, what's left to be done here ?
To recap, Ethan pre-stores Java (which is not apt packages) in /opt/jdk and you can (IMHO should) pre-store .deb's in /var/cache/apt/archive where apt will automatically pick them up.

holgerfriedrich:
@mstormi I had to rebase and resolve conflicts.
Image build works on my local machine, but not in the github action - apt used for downloading the packages seems to ignore target system package lists....

mstormi:
> apt used for downloading the packages seems to ignore target system package lists....

I guess it's using /etc/apt/sources.list and /etc/apt/sources.list.d/* just as on any Un*x, but those are whatever they are in Github
(in the Travis VM which is Ubuntu you could overwrite those but I don't know about Github action).
I don't know off the top of my head but there's apt options to specify which repo to use.

mstormi:
On a sidenote, do you (or you @ecdye) know what's wrong with vpn.bats?
It seems to be skipping the ` install_wireguard install` which creates the config files that BATS complains about to be missing when setup_wireguard is run. If you know a fix, just apply it.

ecdye:
> On a sidenote, do you (or you @ecdye) know what's wrong with vpn.bats?
> It seems to be skipping the ` install_wireguard install` which creates the config files that BATS complains about to be missing when setup_wireguard is run. If you know a fix, just apply it.

I'll take a look at it, I noticed it was failing but wasn't too worried about it.

ecdye:
@holgerfriedrich have you looked at this method: <https://www.raspberrypi.org/forums/viewtopic.php?f=63&t=231762#p1423008>?

holgerfriedrich:
@mstormi Does not help, I cannot chroot from amd64/pc into an arm image. They do it on the same platform, and this is something I do not want to.
The problem I have is that my patched apt.conf which points into the mounted image is not working properly in GitHub. Locally it works perfectly (buster).

ecdye:
> @ecdye Does not help, I cannot chroot from amd64/pc into an arm image. They do it on the same platform, and this is something I do not want to.
> The problem I have is that my patched apt.conf which points into the mounted image is not working properly in GitHub. Locally it works perfectly (buster).

Is it perhaps because GitHub is running on Ubuntu?

holgerfriedrich:
Not figured out yet. I pretty much liked the approach to use apt for downloading the packages because it does the job of selecting the latest version from the repo and checks the signature...

If I cannot get it working we need a different approach.

mstormi:
> The problem I have is that my patched apt.conf which points into the mounted image is not working properly in GitHub.

Maybe a silly question, but why download into the mounted image? Download locally and just copy over.

holgerfriedrich:
@mstormi where to download to is not relevant, getting apt to use the arm repos of the target system and fetching the right versions of the packages is important.
I do not want to trash the apt of the building machine by adding RPi sources list.

mstormi:
>  I do not want to trash the apt of the building machine by adding RPi sources list.

but that's just a temporary one in Github, to just exist for the time of the build, so why not ?

ecdye:
> but that's just a temporary one in Github, to just exist for the time of the build, so why not ?

Correct, whatever you do on the build machine is undone after the operation is done. You could do whatever you want and it should work just fine. I think it might not be a bad idea.

holgerfriedrich:
Test image built fine. Could you test installation to check if the offline install works?
Not sure if I inserted the install at the right place in first-boot and if I got all the dependencies....

Too bad that we do not have a chance to test the new image directly in CI.....

mstormi:
It installed but the java package arch is wrong (i686 on a Raspi wont work)

BTW if you have a Fritz!box it's easy to test, just block Inet in Internet-Filter

ecdye:
I don't think that is caused by me, Holger is it pulling from the Host machine? You may need to change that.

holgerfriedrich:
@mstormi @ecdye java_zulu_fetch internally decides what to download using is_arm - not matter if I source the correct hw_platform config. Now setting hwarch explicitly to any value which makes is_arm return true.
New image build is just running - I do not have time to test on real hw, sorry for that.

holgerfriedrich:
Just for documentation: this is linked to #825 (already closed)

holgerfriedrich:
nice blocker, image reports no disk space while java is extracted....

mstormi:
> 
> 
> nice blocker, image reports no disk space while java is extracted - image needs resize :-/

Ouch. Right, on real HW that's what it does in the first place.
Does "apt autoremove" or "cleanup" help?
Can you set $TMPDIR to point elsewhere where there's space ?

holgerfriedrich:
@mstormi 
Moved extraction from java_zulu_fetch to java_zulu_install (extracting needs too much space, even when I put the zip to my local tmp folder).
Hope this does not cause any unwanted problems. @ecdye pls check
Free space before caching: 275MB, after caching: 95MB.

mstormi:
It failed to install java this time.. I see errors "find: X/opt/jdk/*/binX: No such file or directory"  **with X being nonprintable characters**. Double-

I also noted repositoryurl set to your Github repo and clonebranch=pr-prefetch in the final image, 

Please also rebase on master before you build the next image

BTW it's a lot easier to debug if you run it on hardware so I suggest you put up a RPi for that
You can live-edit the failing install and simply reboot to retry.

mstormi:
will now be off for some hours so don't wait for feedback

holgerfriedrich:
@mstormi @ecdye 
I think we should give up at this point.
* Caching a few packages, ok, fine. Install in first-boot before installer starts. Check.
* Caching Java, ok done (w/o crypto) but when will we install during fist-boot or from the installer script.?
* Caching OH packages - not complicated to implement, but same questuion - when to install? Requires part of the install routine to be launched during build.bash (adding sources list). Same approach as in 1)
...finally we are running out of space in the standard raspbian images. I don't want to re-package everything.

Maybe the target is too ambitious... Sounds like we are aiming at running the first boot during install. Doing so in CI could also be a solution but will cause other issues (ssh keys, resize on first boot to target media size, etc).

Maybe pick from this PR what is useful (signature, curl) and give up on the caching part.

ecdye:
@holgerfriedrich I agree, I think that the best we can offer at the moment without a lot more effort is a partial install that leaves the user with a somewhat functioning system when internet is not available. If a better approach becomes available in the future we can return to the idea.

@mstormi thoughts?

@openhab/architecture-council any guidance?

mstormi:
Mmh.  Maybe I wasn't explicit enough in stating, but the goal has always been to fetch packages at build time and to install them at boot (installation) time only on the target system.
[of course, if that had worked at build time, too, we could have done that as well, but it is not a requirement.]

I'm ok with a "somewhat functioning system" as Ethan called it.
That's a running OH. We need the image to contain  /opt/openhabian, java (Zulu11-32 only) and OH packages for that. 

Frankly I don't understand the part about the need of a part of OH installation to take place at build time.
Just put the current OH packages into /var/cache/apt/packages at build time, nothing more.

Do I get that right: essentially you have implemented all of it (or can with minor efforts) but all that's missing is space ?

ecdye:
@mstormi I think what @holgerfriedrich is more of getting at is that we are starting to enter a new territory in the image process where we would almost be better using something like [pi-gen](https://github.com/RPi-Distro/pi-gen) because we are making such substantial modifications to the base image in such a hacky manner to accomplish the end result.

Before we simply injected a few scripts to optimize the system to run our tool with openHAB. This is almost like trying to create our own mini distro. 

That is my take on the subject.

mstormi:
I agree we must not start hacking too much when things turn out to be way more difficult than we hoped for, and you're right we do not want to end up building our own distro or fundamentally redesign the image build process.

But I'd still like to *understand* what the difficulty is in first before making a decision on giving up.
I still believe it should be as simple as to place the packages in the cache dir hence my question ...

holgerfriedrich:
@mstormi so, what is actually missing:
- add space. ~+300MB. Without looking at existing tooling to ease the process, it would mean: after extracting the image, use dd to create a new file of zeros which is larger by the needed amount, use dd to copy over the extracted image, use gparted to extract data for the last partition, delete it and re-create it to mach bigger image file. No magic, but a bit tricky and needs time to implement.
- adding OH repo to `/etc/apt/sources.list.d/openhab2.list`, do the apt update as already done, apt download. Just putting the packages is not enough. Without updating the cache, install will not know that there is something. Process has been shown for the other apt packages and works in principle (just lacking ~300MB for OH packages).
- rework of installation routines - first-boot to install all cached packages might help to ensure we have the dependencies in place, but doing so for all packages and java installation will not work. So you might be right to use the apt cache dir (and then do not execute the apt-get clean from my PR, and pls check if apt download will put the package to the dest folder and the cache). Java is still a special case, no deb package, but install routines to be executed. Installer would need special handling for  cached zip file. Java crypto extension still missing.

A lot of work and complicating everything.

mstormi:
Alright, thanks for summarizing.
I don't want to force either of you into something you think that cannot work, then again sounds like you agree it *is* doable but a lot of work and fumbling. So do I. Although #3 I don't understand, I it's about Java only and I think Java has been working already, as well as ZRAM ... whatever.

I think it's time for a break. My suggestion is:
Let's not abandon this as we still believe it can be done and you have already invested quite some we don't want to lose.
But let's step back for the time being until we (either of us) find the leisure and inspiration to give it another try.

So for now, let's focus to get the image out (without offline install). Then pick on the next stuff you have been wanting to pick on.

mstormi:
I like these little challenges ... check out #1037

mstormi:
@holgerfriedrich could you please state what remains to be done for a stable *image release*.
Getting back to my last comment, it does not have to fully work if installed under offline conditions.
But it must not misbehave or fail under online conditions either.

Referring to your last comment:
* #1 space is available now (since #1037)
* on #2 you said that it works in principle, so eventually bring the code up to date to include that.
  Do we have a place in the code where to define pkgs that are pre-fetched during build ? I'm only talking about those to work with apt where we can place the .deb in the apt cache dir.
* on #3, Java offline install at least should have been working before, so please check that the download part is executed at image build time. For apps other than Java (and other than #2) I don't care for this image release.

mstormi:
@holgerfriedrich are you around ? Please check my last post.

ecdye:
If I understand correctly, we should be ready for a new image, and then when offline has matured and is ready we can create a patch image or new release. If @holgerfriedrich agrees, then I suggest that we trigger a manual build of the stable branch and test and finalize it.

mstormi:
@holgerfriedrich could you make a statement on feasibility, would you be able to get this included in the next days so we can provide it as part of the image ?

holgerfriedrich:
@mstormi lets start with a rebase and see if we have some free space... Then it should be straight forward to add the OH packages.

But I won't be able to complete this the next days. I suppose we have to do modifications in the installer routines to fallback to the cached packages. My approach would always be to install current versions if internet is available, and fall back to cached ones (maybe with some exceptions, as the cached deb packages, as installing won't hurt and updates will be handled by the system). Maybe We can take this as a starting point for you or Ethan to take over the installation part.

mstormi:
So which packages are we talking about ?

* git, gcc, libattr-dev1 (more ?) .deb packages we need during the install process  -> either in image (gcc I believe) or should work if .deb (for the proper ARM architecture) is downloaded and pre-stored in /var/cache/apt/archives inside the image
* zulu java -> should work today if .tgz is pre-stored in /opt/jdk in the image, the install routine should work to handle that
* zram + overlayfstools -> install builds from source, install should work today if repo clone is pre-stored in /opt/zram[/overlayfs-tools]

I might overlook some packages but these are the important ones we want to work offline, aren't they ?

mstormi:
Holger, any news ?

mstormi:
@holgerfriedrich any chance for a completion soon?

mstormi:
thanks for your efforts!

PR #987, opened 2020-06-19
replace /sbin/runlevel in Travis build to avoid package install error msgs

### LINKED ISSUE #840: Errors in unattended install
various errors seen in Travis deployments, see e.g.
https://travis-ci.com/github/mstormi/openhabian/jobs/350591314#L984
https://travis-ci.com/github/mstormi/openhabian/jobs/350591314#L1579

This may be a rc.local compatibility related problem rather than with Samba or other app that gets installed

mstormi:
avahi-daemon and samba to throw errors because runlevel is undefined inside Docker

Fixes #840 

@ecdye pls review because of Docker stuff


Dockerfile.i386
```diff
@@ -1,61 +0,0 @@
-FROM balenalib/i386-debian:buster-build
```
ecdye, line 1: Why delete?

mstormi: why keep, do we need it? Thought it was forgotten.

ecdye: I don't think we need it, I added for people who may want it but I don't mind if you remove it.

PR #988, opened 2020-06-19
Expand swap

### LINKED ISSUE #982: expand swap
swap is a 100 MB file as per Raspbian default.
That's easily filled by OH with it's current Java parameters to preallocate and page-out quite some memory.

Even with ZRAM swap on, people can quickly run into trouble when there's a memleak.

I'd suggest to enlarge swap on SD to ~1GB or even the typical recommendation of 2*mem.

mstormi:
Fixes #982

Please test with various HW to have != 1 GB.
I've patched this into the "testbuild" branch so set that in /boot/openhabian.conf


mstormi:
no idea why the Adopt build failed

ecdye:
Try restarting the build it seemed to fail for some random reason.

functions/system.bash
```diff
@@ -213,6 +213,33 @@ misc_system_settings() {
   echo "OK"
 }
 
+### change swap (int size in MB)
```
ecdye, line 216: Use 2 `#` i.e. `## change swap (int size in MB)`.

mstormi:
ok to merge?

functions/system.bash
```diff
@@ -213,6 +213,34 @@ misc_system_settings() {
...
+  free=$(df -hk / | awk '/dev/ { print $3 }')
+  if [ "$free" -ge "$minfree" ]; then
+    size=$swap
+  elif [ "$free" -ge "$swap" ]; then
+    size=$totalMemory
+  else
+    return 0
+  fi
+
+  if is_raspbian; then
```
holgerfriedrich, line 238: Return at begin of function of !Raspbian. Add echo setting swap size to ... MB.

Dockerfile.raspberry-pi
```diff
@@ -42,6 +42,10 @@ RUN systemctl enable /etc/systemd/system/resin.service
 RUN git clone https://github.com/gdraheim/docker-systemctl-replacement
 RUN cp docker-systemctl-replacement/files/docker/systemctl3.py /bin/systemctl
 
+RUN rm -f /sbin/runlevel
```
holgerfriedrich, line 45: Can you leave this out or does COPY require that target file does not exist?

mstormi: yes it does because it's a link and I don't wanna overwrite what it points to

functions/system.bash
mstormi: done

functions/system.bash
```diff
@@ -213,6 +213,37 @@ misc_system_settings() {
...
+
+
+  if ! is_raspbian; then return 0; fi
+
+  totalMemory="$(grep MemTotal /proc/meminfo | awk '{print $2}')"
+  if [ -z "$totalMemory" ]; then return 1; fi
+
+  swap=$((2*totalMemory))
+  minfree=$((2*swap))
+  free=$(df -hk / | awk '/dev/ { print $3 }')
```
holgerfriedrich, line 232: $3 is used,  free $4.

holgerfriedrich:
@mstormi if you change the code above to $4, it looks good to me.
No time for testing on real hardware, though.

functions/system.bash
mstormi: Ouch, yep. Sun is shiny, and cold beer too much of a distraction ;-)
That's what code reviews are good for - thanks.

PR #993, opened 2020-06-21
Update Zulu API v1

### LINKED ISSUE #878: make Java install resilient
Following up on #874, there is a need to make (first & foremost but not exclusively) Java install more resilient.

Need to consider both a short term and a strategic approach in parallel.

On short term, we need to evaluate the _list_ of available packages and select hf ones.
@ecdye I didn't have the time to dissect this in detail at 2am so may be wrong, but unlike what you wrote in #874 the URL OHian requested returned MORE than one package and your code seemed to select the last out of those. So it was bad luck that that was the sf one. Double-check that please
and eventually come up with a (non-hot) fix.

On strategic, see #655. I stand by my opinion to keep copies of latest packages known to be good
(= have a copy in openhabian repo). That we should do at least for Azul and Adopt Java.
[NOTE: we may need to have several copies for different HW architectures]

Mind #692: given AC's 'tie break' that took place there, I'd like to get this implemented asap.
It means we'll have a new "stable" branch in addition to today's master branch.
What it means for this issue here is that changed Java install code should evaluate the branch first and if in 'stable', it should download the known good package off the openHABian repo. 
Also mind the connection to #825 (offline install), our image would need to contain a copy as well for that to work. So even with Inet available, on the system is the first place place the code should look for.
If in master it should continue to retrieve the package directly from the source, BUT it should be extended to fall back to the repo cache if the install routine can determine that downloading from source fails or is bad somehow [it obviously cannot detect all instances of 'bad' such as the aarch32sf one, but it _can_ detect some of them].

PS: we should duplicate this approach for other external sources once it has proven to work
PPS: there would have been a chance to spot #874 early: actually all test builds made since the Azul change were failing.

### LINKED ISSUE #879: correct references to Azul java versions
We download and use the "Community" version for both, x86 and ARM.

Code and docs however sometimes still refer to these as "Enterprise" and "Embedded" version s, respectively.

Need to remove/correct these referrals (java-jre.bash, docs/openhabian.md)

ecdye:
Additionally fix Java installation output.

Fixes #878
Fixes #879

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,7 +53,7 @@ jobs:
         - docker build -t openhabian/${HW}-openhabian -f Dockerfile.amd64 .
         - echo -e "\n\e[36mopenHABian test installation using 64bit Azul Enterprise Java 8 running on a\e[31m 64bit OS\e[36m running on native ${HW}...\e[0m\n"
         - docker run --name "openhabian-${HW}" -d openhabian/${HW}-openhabian;
-        - docker exec -i "openhabian-${HW}" bash -c './build.bash local-test && mv ~/.profile ~/.bash_profile && /etc/rc.local'
+        - docker exec -i "openhabian-${HW}" bash -c './build.bash local-test && mv ~/.profile ~/.bash_profile && /boot/first-boot.bash'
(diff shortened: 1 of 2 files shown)


ecdye:
@holgerfriedrich I separated it as you asked. Merge?

PR #994, opened 2020-06-21
adapt Amanda default backup dirs

### LINKED ISSUE #992: fix Amanda config for persistence
Add /var/lib/openhab2/persistence to disklist.
It ain't dumped since ZRAM is an own filesystem.

See also 
https://serverfault.com/questions/915726/difference-between-amgtar-and-gnutar
to fix "STRANGE" errors

mstormi:
to match  to ZRAM persistence

Fixes #992
 
### DIFF
--- a/functions/system.bash
+++ b/functions/system.bash
@@ -221,7 +221,6 @@ misc_system_settings() {
 change_swapsize() {
   local totalMemory
 
-
   if ! is_raspbian; then return 0; fi
 
   totalMemory="$(grep MemTotal /proc/meminfo | awk '{print $2}')"
--- a/functions/backup.bash
+++ b/functions/backup.bash
@@ -39,6 +39,7 @@ create_backup_config() {
   local S3bucket=$8
   local S3accesskey=$9
   local S3secretkey=${10}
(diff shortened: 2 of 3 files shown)


functions/system.bash
```diff
@@ -213,6 +213,7 @@ misc_system_settings() {
   echo "OK"
 }
 
+
```
ecdye, line 216: Why the random newline addition?

mstormi: (*puzzled*) see below how the code looks on *my* machine, the extra newline is to separate variable declarations from code. But I removed it now, too, to be in line with the rest of the code.

  echo "OK"
}

## change system swap size dependent on free space on /
## swap on SD (/var/swap per default) only used after ZRAM swap full if that exists
##
##    change_swapsize(int size in MB)
##
change_swapsize() {
  local totalMemory


  if ! is_raspbian; then return 0; fi

ecdye: Looks better, sometimes GitHub makes it really hard to tell what is going on from the diff. Either way it will be standardized once I get around to doing the backup functions.

PR #996, opened 2020-06-22
copy openhabian_update() to /boot to be available to first-boot.bash

### LINKED ISSUE #986: Always use local repo on install?
@mstormi 

After #961 if I am correct the installer will always use the local repo. This is caused because of this:
``` bash
# must not remove for offline to work
#if [ -d /opt/openhabian ]; then cd /opt && rm -rf /opt/openhabian; fi
# shellcheck disable=SC2154
echo -n "$(timestamp) [openHABian] Cloning myself from ${repositoryurl}, ${clonebranch} branch... "
if ! openhabian_update; then
  echo "$(timestamp) [openHABian] The git repository on the public internet is not reachable."
  echo "$(timestamp) [openHABian] We will continue trying to get your system installed, but this is not guaranteed to work."
fi
ln -sfn /opt/openhabian/openhabian-setup.sh /usr/local/bin/openhabian-config
```
This would work, but `openhabian_update` is defined in `functions/openhabian.bash` which as far as I can tell is never sourced or included in `first-boot.bash` which means that it will always fail and default to the local repo, but this is not the right thing to do because then we won't update to the latest available version when we install from an image.

Is my logic right?

mstormi:
Fixes #986 

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,6 +49,7 @@ jobs:
       arch: amd64
       script:
         # 64bit Azul Java 8 on a 64bit amd64 virtual platform, running in a 64bit OS on Travis AMD64 HW
+        - sed -i 's/^debugmode=.*$/debugmode=maximum/' build-image/openhabian.conf
         - sed -i 's/^java_opt=.*$/java_opt=Zulu8-64/' build-image/openhabian.conf
         - docker build -t openhabian/${HW}-openhabian -f Dockerfile.amd64 .
         - echo -e "\n\e[36mopenHABian test installation using 64bit Azul Enterprise Java 8 running on a\e[31m 64bit OS\e[36m running on native ${HW}...\e[0m\n"
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,7 +50,6 @@ jobs:
       script:
         # 64bit Azul Java 8 on a 64bit amd64 virtual platform, running in a 64bit OS on Travis AMD64 HW
         - sed -i 's/^debugmode=.*$/debugmode=maximum/' build-image/openhabian.conf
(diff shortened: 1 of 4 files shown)


ecdye:
Looks like a fix in theory. Now we need to see if it will actually work. Try getting rid of line 160 in `build.bash` so we can see in CI if it works CI should clone the right repo now anyway.

mstormi:
like so?

ecdye:
That should work.

mstormi:
not there yet https://travis-ci.com/github/mstormi/openhabian/jobs/352403370#L423

ecdye:
Hmm, make sure you bypass the stable default on first install and use whatever has been set in openhabian config and also make sure you are copying the .git directory of the openhabian repo when creating the image.

mstormi:
It isn't about bypassing stable, that's what we normally want to use (except in CI).
When I introduced the stable branch, clonebranch is honored if set else stable is used as the default.
It was about clonebranch (to point to the CI branch) not being used.

So how about this now ? See first test build for maximum output.

mstormi:
I'm still unsure how that'll work "in real life" for both cases, Inet there or Inet gone.

I've flashed a card with the beta image and copied first-boot.bash, helpers and openhabian.bash over.

With Inet on, install fails because "cannot change to /opt/openhabian" because "no such file or dir".
Duh. Yes of course it isn't there ....

ecdye:
Well, that is a start. At least we are giving a partially complete system now. Looks like we are getting close. That seems to be working now. Now just refine it.

mstormi:
Merge this ? The current state is broken IIRC.
Then refine what exactly ? Maybe I merge and you try yourself to see why java ain't there.

ecdye:
Java doesn't install without internet because it us not included in the image yet and we have yet to resole dependency issues for it and zRam

PR #1003, opened 2020-06-25
silence systemctl (daemon-)reload

### LINKED ISSUE #1002: systemd errors in CI
See https://travis-ci.com/github/openhab/openhabian/jobs/353508710#L1186

mstormi:
also dont add openhab user to gpio group twice

Fixes #1002

### DIFF
--- a/functions/openhab.bash
+++ b/functions/openhab.bash
@@ -10,7 +10,7 @@ delayed_rules() {
   else
     rm ${targetdir}/override.conf
   fi
-  cond_redirect systemctl daemon-reload
+  cond_redirect systemctl -q daemon-reload
   cond_redirect systemctl restart openhab2.service
 }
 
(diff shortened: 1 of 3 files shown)


mstormi:
So Ethan bad luck for you: I was able to select you as a reviewer now

ecdye:
Glad that it works. They are silenced with `cond_redirect` and it is intentional that the output is displayed in this case. I believe there is a quiet option for `update-alternatives` but I don't know if it can be set globally.

functions/openhab.bash
```diff
@@ -68,8 +68,7 @@ Check the \"openHAB Release Notes\" and the official announcements to learn abou
   openhabVersion="$(apt-cache madison openhab2 | head -n 1 | cut -d'|' -f2 | xargs)"
 
   if ! cond_redirect apt-get -y --allow-downgrades install "openhab2=${openhabVersion}"; then echo "FAILED (apt)"; return 1; fi
-  cond_redirect adduser openhab gpio
```
ecdye, line 71: Why remove?

mstormi: it fails because it's done in some other place before

PR #1005, opened 2020-06-28
Fix `first-boot.bash` logic

### LINKED ISSUE #1004: Another `first-boot.bash` error?
Once again, during my rewriting efforts I seem to have spotted some faulty logic, I believe:
``` bash
# must not remove /opt/openhabian for offline to work
if [ -d /opt/openhabian ]; then git clone https://github.com/openhab/openhabian.git /opt/openhabian; fi
```
should be 
``` bash
# must not remove /opt/openhabian for offline to work
if ! [ -d /opt/openhabian ]; then git clone https://github.com/openhab/openhabian.git /opt/openhabian; fi
```
As the `-d` option checks for the directory existing, why would we want to clone it if it it already exists?

ecdye:
Fixes #1004

### DIFF
--- a/build-image/first-boot.bash
+++ b/build-image/first-boot.bash
@@ -179,7 +179,7 @@ else
 fi
 
 # must not remove /opt/openhabian for offline to work
-if [ -d /opt/openhabian ]; then git clone https://github.com/openhab/openhabian.git /opt/openhabian; fi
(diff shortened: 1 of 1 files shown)


ecdye:
That is so much more convenient now that I can actually request a review from you @mstormi so that bad luck goes both ways. 😄

PR #1008, opened 2020-06-29
Add build image function to GitHub Actions

ecdye:
The 'Build Image' action was primarily written by Holger Friedrich and
then edited for better functionality in the openHABian core repo.

This also contains other changes to the CI setup, such as moving the
ShellCheck checks out of Travis and instead enforcing them in GitHub
Actions.

Along with theses changes, some improvements to the build script and
install output were also made. These changes to the build script include
using cURL over wget for downloading needed files and also cloning the
openHABian repo in the build process rather than the install process.
This change was made because the repo must be present for offline
install to work, and as such it makes sense to provide it with the image
to reduce points of failure on install.

Fixes #991

Co-authored-by: Holger Friedrich <mail@holger-friedrich.de>
Co-authored-by: Markus Storm <markus.storm@gmx.net>
Signed-off-by: Ethan Dye <mrtops03@gmail.com>

PR #1016, opened 2020-07-07
Fix IPv6 enabling/disabling

ecdye:
The current implementation of IPv6 enabling/disabling had many issues
and could not enable IPv6 after it was disabled because of broken
functionality. This contains a simpler and fixed method to
enable/disable IPv6.

### DIFF
--- a/functions/openhabian.bash
+++ b/functions/openhabian.bash
@@ -193,86 +193,29 @@ ua-netinst_check() {
   fi
 }
 
-choose_ipv6() {
-  local IPV6_DISABLE="net.ipv6.conf.all.disable_ipv6"
+## Enable / Disable IPv6 according to the users configured option in '$CONFIGFILE'
+##
+##    config_ipv6()
+##
+config_ipv6() {
+  local aptConf
+  local sysctlConf
 
-  # shellcheck disable=SC2154
-  if [[ "$ipv6" == "disable" ]]; then
-    if ! grep -qE "^${IPV6_DISABLE}" "${SYSCTL_INIT}"; then
-      disable_ipv6 yes
+  aptConf="/etc/apt/apt.conf/S90force-ipv4"
+  sysctlConf="/etc/sysctl.d/99-sysctl.conf"
+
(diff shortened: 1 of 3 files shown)


functions/openhabian.bash
```diff
@@ -193,86 +193,33 @@ ua-netinst_check() {
...
+  sysctlConf="/etc/sysctl.d/99-sysctl.conf"
+
+  if [[ "${ipv6:-enable}" == "disable" ]]; then
+    echo -n "$(timestamp) [openHABian] Disabling IPv6... "
+    {
+      echo "# Disable all IPv6 functionality"
+      echo "net.ipv6.conf.all.disable_ipv6=1"
+      echo "net.ipv6.conf.default.disable_ipv6=1"
+      echo "net.ipv6.conf.lo.disable_ipv6=1"
+    } >> "$sysctlConf"
```
mstormi, line 214: '>>' or '>' ? what if that's executed for the 2nd time ?

maybe not today but if there was a menu option ....

ecdye: I didn't know if there was any other stuff that `/etc/sysctl.d/99-sysctl.conf` was used for so I didn't want to override it. If there is nothing else then I'll just create a includes file and use that.

ecdye: Changed. @mstormi is that better?

ecdye:
Also #1008 needs to be merged asap as PR builds are failing because it contains CI-fixes to the build process.

mstormi:
Seems you deleted a comment about PR not building hence a need to merge #1008 asap ?
So is that the case or not ? This one built fine ...

ecdye:
> Seems you deleted a comment about PR not building hence a need to merge #1008 asap ?
> So is that the case or not ? This one built fine ...

Sort of, there are some fixed errors in output in #1008 for ci builds.

mstormi:
So are builds broken (=need to merge 1008 asap) or can we wait for Holger to review ?

ecdye:
They are not completely broken, they just are not correctly reflecting what is happening in their output as some things need to be changed. We can wait.

PR #1026, opened 2020-07-10
Fix Wireguard BATS tests output

ecdye:


mstormi:
thanks but frankly I still don't understand why the install_wireguard routine wasn't started (that's at least what I think because I got absolutely no output no matter what I tried)

ecdye:
It started but no output was shown because the output was overwritten by the next command.

PR #1030, opened 2020-07-12
Verify signature of Raspberry Pi OS image.

holgerfriedrich:
* verify signature of download images using key from raspberrypi.org website
* silence install_zram_code warnings regarding detached head
* move installation of libattr1-dev to init_zram_mounts


holgerfriedrich:
separate out all parts of #985 which are not related to caching

PR #1035, opened 2020-07-15
Patch day 2020-07-15 master => stable

mstormi:
Forwarding patches as of 2020/07/15 from master to stable branch

### DIFF
--- a/functions/packages.bash
+++ b/functions/packages.bash
@@ -182,7 +182,7 @@ homegear_setup() {
   if ! cond_redirect adduser "${username:-openhabian}" homegear; then echo "FAILED"; return 1; fi
   if cond_redirect adduser openhab homegear; then echo "OK"; else echo "FAILED"; return 1; fi
   echo -n "$(timestamp) [openHABian] Setting up Homegear service... "
-  cp "$BASEDIR"/includes/homegear*.service /lib/systemd/system/
+  cp "$BASEDIR"/includes/homegear*.service /etc/systemd/system/
   if running_in_docker; then sed -i '/RuntimeDirectory/d' /lib/systemd/system/homegear*; fi
   cond_redirect systemctl -q daemon-reload &>/dev/null
   if ! systemctl enable --now homegear homegear-management; then echo "FAILED (enable service)"; return 1; fi


mstormi:
All I did was a PR with the diff from master to stable
Why are there so many conflicts ? 

I believe I must be overlooking something substantial.

ecdye:
Why not just reset stable to master? it would be easier and our git update process should handle it just fine.

mstormi:
> Why not just reset stable to master? it would be easier and our git update process should handle it just fine.

Yes that would work but I hate stuff I don't understand. Also, the idea is to have a changelog.

ecdye:
It will have a changelog let me try to make an example

ecdye:
So it looks like there may not be a really good way to do it beacuse of the way GitHub compares the trees. THe cleanest way to do it may be to force reset stable to master and add a tag to stable indicating the patch day on the last commit.

mstormi:
ok, forced stable to master and added a tag "patchday-20200715"

ecdye:
Hmm, thats funny. I don't see the tag, you should add it to master and to stable it will help to give refrence and make it so it does not overwrite when we do future patchdays.

mstormi:
You need to explicitly force transfer of tags .. either you forgot, or I did :)

ecdye:
How could I have done it? I don't have write access to the repo. :)

mstormi:
No I mean I think you also need to explicitly fetch them.

ecdye:
No I was looking at github

mstormi:
yeah it was me.
Can you see them now?

ecdye:
Yep

mstormi:
I've given you write access. Please use responsibly.

ecdye:
I was not expecting this, however, I will.

mstormi:
> I was not expecting this, however, I will.

You could review & eventually merge #1037 for a test.

BTW, we've been [iced](https://archiveprogram.github.com/).

ecdye:
Will do.

> BTW, we've been iced.

No kidding, I guess as a final deadline for our next image we should strive for 2/2/2021 so it is preserved in the vault. That seems like a very achievable target, right?

mstormi:
I'd like to get that out a little earlier 🧊

PR #1039, opened 2020-07-18
Add support for FIND3

### LINKED ISSUE #632: Update FIND optional package to version 3
A new major rewrite of the FIND framework have been released. Would be nice have our script updated as well to install that version.

ecdye:
Fixes #632


mstormi:
needs a rebase

mstormi:
does running FIND3 work for you?
I have it running but not really working, DB doesn't fill

mstormi:
Glancing at the time it takes and offline install, would you be able to compile binaries on Github for both, ARMv7l and amd64, then just package those ?

ecdye:
@mstormi No, not really, offline install for Find3 would be very difficult. But because it is an optional package, I am less concerned about it. It runs using this setup. There are three pieces. the server, the ai server, and the cli fingerprinting client. The fingerprinting client is what acually does the work the UI options won't work correctly unless you have the CLI scanner running properly.

The cli scanner has two modes Active (easy to set up), and passive (takes monumental effort to get working) which requires modifying your wifi chipset firmware to enable monitor mode and after that is works flawlessly. The menu options provided by the server interface with the cli-client only during passive scanning. 

That is my understanding. I am installing the three parts in this however I am not going to make a monitor mode patch unless there is a high demand for it because it is a very involved process and would require a lot of work. It is not to difficult to set up on your own using Google to figure it out.

ecdye:
Correction. My desire for a challenge drove me to create an install routine for enabling Monitor Mode. I have very clearly labeled it as ALPHA and in no way guaranteed to work. I can only test that it works on my RPI4 so any additional testing would be very helpful. 

After it is installed the patch makes regular WiFi not work without manual configuration and setup so it will not work for regular AP connection. It may be possible to run monitor mode and regular mode concurrently, however I have not figured it out.

mstormi:
did you check out the existing FIND2 installation method ? I'd think it remains mostly applicable.

ecdye:
> did you check out the existing FIND2 installation method ? I'd think it remains mostly applicable.

No it is not, @schollz building releases anymore.

mstormi:
I was referring to the deployment method via systemd I used in openHABian

ecdye:
I am still using systemctl to launch the service. But the fingerprinting client need manual configuration just like last time.

mstormi:
I believe the fingerprinting client ain't needed. You won't be carrying your server around for the purpose of learning but only your mobile.

ecdye:
> I believe the fingerprinting client ain't needed. You won't be carrying your server around for the purpose of learning but only your mobile.

You can use it in passive mode to scan all over, it is needed so that you can scan anything.

functions/find.bash
```diff
@@ -0,0 +1,266 @@
...
+
+  if (whiptail --title "FIND3 installation?" --yes-button "Continue" --no-button "Cancel" --yesno "$introText" 22 80); then echo "OK"; else echo "CANCELED"; return 0; fi
+
+  echo -n "$(timestamp) [openHABian] Configuring FIND3... "
+  if ! MQTT_SERVER="$(whiptail --title "FIND3 Setup" --inputbox "\\nPlease enter the hostname and port of the device your MQTT broker is running on:" 10 80 localhost:1883 3>&1 1>&2 2>&3)"; then echo "CANCELED"; return 0; fi
+  if [[ $MQTT_SERVER != "localhost:1883" ]]; then
+    if ! (whiptail --title "MQTT Broker Notice" --yes-button "Continue" --no-button "Cancel" --yesno "$brokerText" 12 80); then echo "CANCELED"; return 0; fi
+  fi
+  if [[ -f $mqttPass ]]; then
+    if ! MQTT_ADMIN=$(whiptail --title "FIND3 MQTT Setup" --inputbox "\\nEnter a username for FIND3 to connect with on your MQTT broker:" 9 80 find 3>&1 1>&2 2>&3); then echo "CANCELED"; return 0; fi
```
mstormi, line 44: is that proper English ? "to connect with on your MQTT broker"

functions/find.bash
```diff
@@ -0,0 +1,266 @@
...
+    if ! cond_redirect apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 648ACFD622F3D138; then echo "FAILED (add keys)"; return 1; fi
+
+    echo -e "Package: *\\nPin: release a=buster-backports\\nPin-Priority: 90\\n" > /etc/apt/preferences.d/limit-buster-backports
+
+    if ! cond_redirect apt-get update; then echo "FAILED (update apt lists)"; return 1; fi
+    if cond_redirect apt-get install --yes --target-release "buster-backports" golang-go; then echo "OK"; else echo "FAILED"; return 1; fi
+  fi
+}
+
+## Function to enable montior mode on RPi
```
mstormi, line 140: montior

includes/90-wireless.rules
```diff
@@ -0,0 +1 @@
+SUBSYSTEM=="ieee80211", ACTION=="add", KERNEL=="%KERNEL", RUN+="/sbin/iw phy %k interface add mon0 type monitor", RUN+="/sbin/ip link set mon0 up"
```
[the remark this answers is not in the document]
ecdye: This is for the monitor mode patch to ensure that the interface is correctly made on reboot.

includes/find3server.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=Framework for Internal Navigation and Discovery server
+Documentation=https://www.internalpositioning.com/doc/
+After=mosquitto.service find3ai.service
+
+[Service]
+EnvironmentFile=-/etc/default/find3server
```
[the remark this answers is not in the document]
ecdye: No idea, however I observed no issues with submitting fingerprints and DB access in my tests.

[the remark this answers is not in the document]
mstormi: ... because you run FIND as root. Shouldn't it have its own user or run as openhab ?

functions/find.bash
ecdye: No it should probably be "to connect to your MQTT broker with" I'll change it.

functions/packages.bash
```diff
@@ -258,109 +258,6 @@ mqtt_setup() {
   fi
 }
 
-## Function for installing FIND to allow for indoor localization of WiFi devices.
```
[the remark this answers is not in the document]
ecdye: Makes sense.

includes/find3server.service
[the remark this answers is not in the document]
ecdye: No, it never needed it before, try using this new setup and see if it gives you issues.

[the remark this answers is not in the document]
mstormi: I'm sure it works as root. The challenge is rather to run with _minimum_ rights hence as an ordinary user.
That's a security principle I'd like not to drop.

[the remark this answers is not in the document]
ecdye: So essentially, create a new find user and find group, add that to the service files and what else? I have little experience with services.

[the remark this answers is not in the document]
mstormi: That's it more or less. See https://github.com/openhab/openhabian/blob/master/includes/findserver.service for a start

User yes, Group I don't think so we can use openhabian
Rundir (to store the DB) I'm not sure if we need ZRAM (once learnt there should be little write access, right?).
But we will want to have the DB dir in the Amanda backup.

User=find
Group=openhabian
RuntimeDirectory=/var/lib/findserver                # or /var/lib/openhab2/findserver to have ZRAM ?
RuntimeDirectoryMode=0775

mstormi:
just wanted to install FIND3 from your repo to my test box so I changed repositoryurl and clonebranch ...
but I forgot we (intentionally) cannot clone from non-openhabian repo.
Can you patch your code into testbuild ?

includes/find3server.service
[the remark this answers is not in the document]
ecdye: ZRAM is probably a good idea because it is continuously learning even once original learning is done because it uses AI. 

I will work on that.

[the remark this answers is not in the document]
mstormi: to be consistent we would need to add another line to /etc/ztab. Would need to know how much RAM we need to prepare for.

mstormi:
$ adduser openhabian find3
+ adduser openhabian find3
adduser: The group `find3' does not exist.
+ return 1

includes/find3server.service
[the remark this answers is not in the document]
ecdye: No idea on that one

ecdye:
I later changed and got rid of that and probably overwrote your changes as well. That was before you told me the proper way to do it.

mstormi:
```
Fresh Installation...
+ cond_redirect git clone https://github.com/schollz/find3.git /opt/find3
+ [[ -n '' ]]
+ echo -e '\n\033[90;01m$ git clone https://github.com/schollz/find3.git /opt/find3 \033[39;49;00m'

$ git clone https://github.com/schollz/find3.git /opt/find3
+ git clone https://github.com/schollz/find3.git /opt/find3
Cloning into '/opt/find3'...
remote: Enumerating objects: 15, done.
remote: Counting objects: 100% (15/15), done.
remote: Compressing objects: 100% (15/15), done.
remote: Total 5968 (delta 4), reused 1 (delta 0), pack-reused 5953
Receiving objects: 100% (5968/5968), 20.20 MiB | 9.98 MiB/s, done.
Resolving deltas: 100% (3188/3188), done.
+ return 0
+ echo OK
OK
++ command -v go
+ [[ -x /usr/bin/go ]]
++ timestamp
++ date +%F_%T_%Z
+ echo -n '2020-07-28_23:44:29_CEST [openHABian] Building FIND3 server... '
2020-07-28_23:44:29_CEST [openHABian] Building FIND3 server... + cond_redirect cd /opt/find3/server/main
+ [[ -n '' ]]
+ echo -e '\n\033[90;01m$ cd /opt/find3/server/main \033[39;49;00m'

$ cd /opt/find3/server/main
+ cd /opt/find3/server/main
+ return 0
+ cond_redirect go build -o find3server -v
+ [[ -n '' ]]
+ echo -e '\n\033[90;01m$ go build -o find3server -v \033[39;49;00m'

$ go build -o find3server -v
+ go build -o find3server -v
build github.com/schollz/find3/server/main: cannot find module for path archive/tar
+ return 1
+ echo 'FAILED (build)'
FAILED (build)
+ return 1
```

ecdye:
No idea what that is being caused by, try again in a day and see if it is just upstream issues.

mstormi:
> 
> 
> No idea what that is being caused by, try again in a day and see if it is just upstream issues.

but it doesn't try to download anythng at that stage does it ?

ecdye:
I'm not entirely sure, I have no real experience with GO. When I originally tested the install it worked fine for me, does it still give you the same error? If so I'll look into it and see if I can figure it out.

mstormi:
yes still same error

ecdye:
What version of Go do you have installed?

mstormi:
[20:37:33] openhabian@devpi:~$ go version
go version go1.11.6 linux/arm

wasn't it part of your code to install it ?  I think 1.11 was/is the current stable version.

ecdye:
No, that is why. You must have previously installed Go I am using backports to install Go 1.14 the current stable version because there are some annoying bugs with 1.11 try purging golang-* and reinstalling.

mstormi:
Ok install went fine this time.
find3server.service still failing. Am I correct in guessing that it cannot connect to mosquitto with the credentials I gave ?

[21:22:23] root@devpi:/home/openhabian# journalctl -xu find3server
-- Logs begin at Tue 2020-07-28 21:17:02 CEST, end at Wed 2020-07-29 21:21:58 CEST. --
Jul 29 21:19:42 devpi systemd[1]: /etc/systemd/system/find3server.service:8: RuntimeDirectory= path is absolute, ignoring: /var/lib/find3se
Jul 29 21:19:42 devpi systemd[1]: Started Framework for Internal Navigation and Discovery server.
-- Subject: A start job for unit find3server.service has finished successfully
-- Defined-By: systemd
-- Support: https://www.debian.org/support
--
-- A start job for unit find3server.service has finished successfully.
--
-- The job identifier is 2294.
Jul 29 21:19:42 devpi systemd[11342]: find3server.service: Failed to determine user credentials: No such process
Jul 29 21:19:42 devpi systemd[11342]: find3server.service: Failed at step USER spawning /usr/sbin/find3server: No such process
-- Subject: Process /usr/sbin/find3server could not be executed
-- Defined-By: systemd
-- Support: https://www.debian.org/support
--
-- The process /usr/sbin/find3server could not be executed and failed.
--
-- The error number returned by this process is ERRNO.
Jul 29 21:19:42 devpi systemd[1]: find3server.service: Main process exited, code=exited, status=217/USER
-- Subject: Unit process exited
-- Defined-By: systemd
-- Support: https://www.debian.org/support
--
-- An ExecStart= process belonging to unit find3server.service has exited.
--
-- The process' exit code is 'exited' and its exit status is 217.
Jul 29 21:19:42 devpi systemd[1]: find3server.service: Failed with result 'exit-code'.
-- Subject: Unit failed
-- Defined-By: systemd
-- Support: https://www.debian.org/support
--
-- The unit find3server.service has entered the 'failed' state with result 'exit-code'.
Jul 29 21:19:42 devpi systemd[1]: find3server.service: Service RestartSec=100ms expired, scheduling restart.
Jul 29 21:19:42 devpi systemd[1]: find3server.service: Scheduled restart job, restart counter is at 1.
-- Subject: Automatic restarting of a unit has been scheduled
-- Defined-By: systemd
-- Support: https://www.debian.org/support
--
-- Automatic restarting of the unit find3server.service has been scheduled, as the result for
-- the configured Restart= setting for the unit.
Jul 29 21:19:42 devpi systemd[1]: Stopped Framework for Internal Navigation and Discovery server.
-- Subject: A stop job for unit find3server.service has finished
-- Defined-By: systemd
-- Support: https://www.debian.org/support
--
-- A stop job for unit find3server.service has finished.
--
-- The job identifier is 2426 and the job result is done.
Jul 29 21:19:42 devpi systemd[1]: Started Framework for Internal Navigation and Discovery server.
-- Subject: A start job for unit find3server.service has finished successfully
-- Defined-By: systemd
-- Support: https://www.debian.org/support
--
-- A start job for unit find3server.service has finished successfully.
--

ecdye:
That or the user changes to the service file are not correct. The MQTT authentication defaults have worked fine for me in the past.

```
Jul 29 21:19:42 devpi systemd[1]: /etc/systemd/system/find3server.service:8: RuntimeDirectory= path is absolute, ignoring: /var/lib/find3se
```
Why does it say that?

mstormi:
> Why does it say that?

RuntimeDirectory has to be a relative dirname, to be created by systemd below /run

ecdye:
> RuntimeDirectory has to be a relative dirname, to be created by systemd below /run

So what should I make it?

mstormi:
> So what should I make it?

RuntimeDirectory=find3server
this is where PIDfile etc can go.

Not to be confused with WorkingDirectory to contain the DB which should be on ZRAM.
That you should add to /etc/ztab as well.

ecdye:
What would you suggest the zram config string to be?

mstormi:
>  I added a best guess in the code at the ZRAM config, what do you think?

Hmm, zram-config restart will stop OH, too. A simple /usr/local/bin/zram-config start will add without (in my testing it duplicated one of the existing dirs but that's only until next reboot so I don't mind)

We should also add this to /etc/amanda/openhab-{dir,aws}/disklist if either exists
and in backup.bash (if Amanda is installed later than FIND)

includes/find3ai.service
```diff
@@ -0,0 +1,16 @@
+[Unit]
+Description=Framework for Internal Navigation and Discovery AI
+Documentation=https://www.internalpositioning.com/doc/
+After=mosquitto.service
+
+[Service]
+WorkingDirectory=/opt/find3/server/ai
+RuntimeDirectory=find3ai
```
mstormi, line 8: I think it should be the same rundir as find3

functions/zram.bash
```diff
@@ -74,6 +74,12 @@ init_zram_mounts() {
     if ! cond_redirect install -m 644 "$zramInstallLocation"/openhabian-zram/ro-root.sh /usr/local/share/zram-config/ro-root.sh; then echo "FAILED (ro-root)"; return 1; fi
     if cond_redirect install -m 644 "$zramInstallLocation"/openhabian-zram/zram-config.logrotate /etc/logrotate.d/zram-config; then echo "OK"; else echo "FAILED (logrotate)"; return 1; fi
 
+    if [[ -f /etc/systemd/system/find3server.service ]]; then
+      echo -n "$(timestamp) [openHABian] Adding FIND3 to ZRAM... "
+      if ! cond_redirect systemctl stop zram-config.service; then echo "FAILED (stop service)"; return 1; fi
```
mstormi, line 79: no stop and no start but a one-shot call of /usr/local/bin/zram-config start

includes/find3ai.service
ecdye: ok

functions/zram.bash
ecdye: Done

ecdye:
> We should also add this to /etc/amanda/openhab-{dir,aws}/disklist if either exists
> and in backup.bash (if Amanda is installed later than FIND)

I think that I'll do that as part of #1055 as I am making other major changes to the Amanda script there and it will be simpler to do it in that one.

functions/zram.bash
```diff
@@ -74,6 +74,10 @@ init_zram_mounts() {
     if ! cond_redirect install -m 644 "$zramInstallLocation"/openhabian-zram/ro-root.sh /usr/local/share/zram-config/ro-root.sh; then echo "FAILED (ro-root)"; return 1; fi
     if cond_redirect install -m 644 "$zramInstallLocation"/openhabian-zram/zram-config.logrotate /etc/logrotate.d/zram-config; then echo "OK"; else echo "FAILED (logrotate)"; return 1; fi
 
+    if [[ -f /etc/systemd/system/find3server.service ]]; then
```
mstormi, line 77: That won't start the dir as ZRAM, only on next reboot.
If we want it right away as I said we need a one-time zram-config run for the single new dir like so
mv /etc/ztab /tmp
echo "...">/etc/ztab
/usr/local/bin/zram-config start
mv /tmp/ztab /etc

ecdye: That is before ZRAM has been initialized so it should be fine. This is in the ZRAM install script.

mstormi: that won't work.
it is in ZRAM *install* script which is run at install time only. FIND install is not run.
FIND will be added later, interactively

mstormi:
[23:09:52] root@devpi:/opt/find3/server/main# su - find3
su: user find3 does not exist

but you run the service via systemd as user "find3"

functions/zram.bash
ecdye: It is also run from the menu if a user decides to uninstall and reinstall.

mstormi: I don't think it makes sense that way:
On unattended install, the find3server unit file is not there thus nothing is added to ztab
When a user decides to install FIND, it won't (and shouldn't) call the zram install routine, will it?
And even if we did, it won't start the new ZRAM dir without a reboot.
So better use my sniplet.

ecdye: No, the code to add it appears in both places so that it will be added in both cases. Check find.bash again to see what I mean.

mstormi: ok, it's added. But the point is it still is not activated.
That's why you need to run zram-config start with a one-line ztab. Or reboot, but that's not necessary.

ecdye: Yes it will be, it is activated in both. In `zram.bash` the line is added to `/etc/ztab` before ZRAM is started for the first time which means it will automatically be added after the first start.

In `find.bash` it is activated using a `zram-config start` command directly following it. I see no issues with this. 

Quite frankly, I prefer the original method which was simply stopping zram and restarting because that is the recommend method to do it anyway. It shouldn't matter if openhab is restarted because this will only be run during server maintenance or setup to begin with.

mstormi: > In `find.bash` it is activated using a `zram-config start` command directly following it. I see no issues with this.

zram-config isn't built to cope with changes to ztab. It duplicates at least some 'log' ztab lines resulting in extra zram devices we don't want. The method to prevent is to use a temporary one-line ztab.
ZRAM restarts are ugly (stops syslog, does lazy unmounting and probably has more potential breaking issues) so I want to avoid that.

On a similar note, we also need to ensure we don't add to ztab more than once e.g. if a user runs FIND install twice.

ecdye: Perhaps a better solution is to change the upstream ZRAM code to be safer and more reliable? IMHO if a service (in this case ZRAM) is intended to be used in a specific way (in this case stop and restart to allow for safe and proper handling) it should function properly in that way and the end user should not have to worry about it.

I personally have never observed any issues with restarting ZRAM and it has always been fairly quick in my experience. I would prefer to use the service method because it avoids the duplication of ZRAM which could lead to not enough RAM being available, and it avoids the messy process of adding to a live ZRAM config.

Because of this, if you don't mind I would prefer to stick with starting and stopping the service and if issues arise we can look into using your method.

mstormi: > Perhaps a better solution is to change the upstream ZRAM code to be safer and more reliable?

Yes but that's tough. It was written by someone who no longer maintains it and I've taken maintainership and it took me many hours to reverse engineer but the code really is not built to handle that, it would need a major refactoring which I'm not really keen to take on.

> I personally have never observed any issues with restarting ZRAM

But I have, and I've spent countless hours supporting others that did.
Just spent many more with a non-UNIX guy (duh!) and yes, sometimes even after zram stop there's a ZRAM device left.
I agree that we must avoid duplication (not sure if it really uses RAM twice but a fair chance it does) but I hence disagree service stop is a better or safe way.
This is exactly why I've put testing efforts into this and came up with that other method so I really prefer that.

ecdye: Very well, I have added it.

ecdye:
@mstormi I have updated the `testbuild` branch with all of the most recent changes, test it again now and see if everything is now working properly. 

My tests of the latest changes have finished successfully and I am ready to merge if you approve.

mstormi:
Did you mean to say latest *FIND3 changes* are now in testbuild or ALL latest changes ?

I just ran an install of testbuild on HW and it seems it is missing the latest ZRAM change (or that one is bad)?
Let's track image issues in #1063

ecdye:
It should be all the changes made in this branch. Let me look into it.

ecdye:
The ZRAM change appears to be included and working. Not sure why it is not working for you, was this a fresh install?

PR #1065, opened 2020-07-29
Reboot at the end of a fresh install

### LINKED ISSUE #1058: Reboot on fresh install?
I think that after a fresh install we should automatically reboot the system. I suggest this for a few reasons,

1.   Several operations performed on first install require reboot for changes to take effect (locale, wifi, ipv6...).
2.  Acts as a way to refresh the network connection for instances where that may have changed.
3.  It is a very common operation to perform after system upgrades and our install process often installs kernel upgrades or other significant upgrades to the system during setup.

Thoughts?

ecdye:
Fixes #1058

### DIFF
--- a/build-image/first-boot.bash
+++ b/build-image/first-boot.bash
@@ -210,12 +210,11 @@ echo -n "$(timestamp) [openHABian] Waiting for openHAB to become ready on ${HOST
 # this took ~130 seconds on a RPi2
 if ! tryUntil "curl --silent --head http://${HOSTNAME:-openhab}:8080/start/index |& grep -qs 'HTTP/1.1 200 OK'" 20 10; then echo "OK"; else echo "FAILED"; exit 1; fi
 
-echo "$(timestamp) [openHABian] Visit the openHAB dashboard now: http://${HOSTNAME:-openhab}:8080"
-echo "$(timestamp) [openHABian] To gain access to a console, simply reconnect."
-echo "$(timestamp) [openHABian] First time setup successfully finished."
+echo "$(timestamp) [openHABian] First time setup successfully finished. Rebooting your system!"
+echo "$(timestamp) [openHABian] After rebooting the openHAB dashboard will be available at: http://${HOSTNAME:-openhab}:8080"
+echo "$(timestamp) [openHABian] After rebooting to gain access to a console, simply reconnect using ssh."
 sleep 12
 if [[ -x $(command -v python3) ]]; then bash /boot/webif.bash inst_done; fi
-sleep 12
 
 if running_in_docker; then
   PID="/var/lib/openhab2/tmp/karaf.pid"
@@ -227,5 +226,7 @@ if running_in_docker; then
   fi
   echo -e "$COL_DEF"
 fi
+
+reboot
 
 # vim: filetype=sh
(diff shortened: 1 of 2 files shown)


ecdye:
The problem is that the kernel patch won't fix the issues listed in #1058. I'll go change the statements to be more accurate as well.

build-image/first-boot.bash
```diff
@@ -212,12 +212,13 @@ echo -n "$(timestamp) [openHABian] Waiting for openHAB to become ready on ${HOST
 # this took ~130 seconds on a RPi2
 if ! tryUntil "curl --silent --head http://${HOSTNAME:-openhab}:8080/start/index |& grep -qs 'HTTP/1.1 200 OK'" 20 10; then echo "OK"; else echo "FAILED"; exit 1; fi
 
-echo "$(timestamp) [openHABian] Visit the openHAB dashboard now: http://${HOSTNAME:-openhab}:8080"
-echo "$(timestamp) [openHABian] To gain access to a console, simply reconnect."
-echo "$(timestamp) [openHABian] First time setup successfully finished."
+echo "$(timestamp) [openHABian] First time setup successfully finished. Rebooting your system!"
+echo "$(timestamp) [openHABian] After rebooting the openHAB dashboard will be available at: http://${HOSTNAME:-openhab}:8080"
+echo "$(timestamp) [openHABian] After rebooting to gain access to a console, simply reconnect using ssh."
 sleep 12
```
mstormi, line 218: remove the sleeps ?

build-image/webif.bash
```diff
@@ -45,12 +45,12 @@ if [ "$1" = "inst_done" ]; then
   # shellcheck disable=SC2016
   echo '<html>
 <head>
-<meta http-equiv='"refresh"' content='"10;url=http://${HOSTNAME:-$hostname}:8080/"' />
+<meta http-equiv='"refresh"' content='"120;url=http://${HOSTNAME:-openhab}:8080/"' />
```
mstormi, line 48: is 120 the right number ? Ok for a start.
On a sidenote, the install log is refreshed every 10 secs, but it's always jumping back to the top of file.
Do you know how to jump to the bottom instead ?

ecdye: No clue, it was just a random value I chose. If it is not enough and we observe it often then we can change it.

No, I looked into it a week or two ago but didn't find anything.

build-image/first-boot.bash
ecdye: Why are they there?

mstormi: no clue, I guess they originate from the guy mues-lee that wrote the webif stuff.

ecdye: K. I'll remove them, I see no reason to keep them.

ecdye: Actually the first one has a purpose, to make sure the last couple of messages actually show in the log on the website. Therefore, I will leave it but remove the other one.

build-image/webif.bash
mstormi: http://sourcetricks.com/2010/07/javascript-scroll-to-bottom-of-page.html
but it doesn't work :(

build-image/first-boot.bash
mstormi: they will either way as "reboot" stops OH first, ZRAM next so enough time

ecdye: No they won't because what is happening is he is delaying it so the final messages appear in the Web log then turning off the server. The first sleep is needed for UX the second one can go though.

PR #1070, opened 2020-07-31
Start ZRAM and Frontail before OH2 also on OH2 stop/start/restart

### LINKED ISSUE #1069: OH restart makes ZRAM fail restart
see https://community.openhab.org/t/mapdb-doesnt-restoreonstartup/102754/76

My preliminary idea is that OH stop/start runs ZRAM stop as ZRAM is configured "PartOf=openhab2.service" but it does not execute ZRAM start again. (I'm just wondering why the /var/log ZRAM dir is left at that guy's system)

See also https://community.openhab.org/t/frontail-log-not-running-get-a-connection-refused/102873/20 for similar problem with OH and frontail.

I think in both cases the service1 to be a PartOf service2 (=oh2) is stopped when service2 is stopped.
But it remains stopped when service2 is started.

@BClark09 do you have any idea how we could  cope with that ?
I wouldn't know how to do that other than to make openhab2.service dependent (?) on all of these services but IIRC you don't want that ...

cc @holgerfriedrich 

PS: And oh2 stop/start should not change config like in the link above hence systemd should not ask for reload !?

mstormi:
Fixes #1069 

### DIFF
--- a/functions/openhab.bash
+++ b/functions/openhab.bash
@@ -1,6 +1,22 @@
 #!/usr/bin/env bash
 
-## Function to quicky rename openHAB rules back and forth after two minutes to
+## Generate systemd dependencies for ZRAM, Frontail and others to start together with OH2
+## This is done using /etc/systemd/system/openhab2.service.d/override.conf
+##
+##    create_sys_dependencies()
+##
+create_sys_dependencies() {
+  local targetDir
+
+  targetDir="/etc/systemd/system/openhab2.service.d"
+
+  if ! cond_redirect mkdir -p $targetDir; then echo "FAILED (prepare directory)"; return 1; fi
+  if ! cond_redirect rm -f "${targetDir}"/override.conf; then echo "FAILED (clean directory)"; return 1; fi
+  if cond_redirect cp "${BASEDIR:-/opt/openhabian}"/includes/openhab2-override.conf "${targetDir}"/override.conf; then echo "OK"; else echo "FAILED (copy configuration)"; return 1; fi
+}
+
+
+## Function to quickly rename openHAB rules back and forth after two minutes to
 ## speed up startup of openHAB.
 ## This is done using /etc/systemd/system/openhab2.service.d/override.conf
 ## Valid arguments: "yes" or "no"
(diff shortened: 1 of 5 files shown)


ecdye:
That only works if the user leaves delayed rules enabled, right?

mstormi:
good point I didn't think of that.... will create a reduced unit file rather than rm it on disabling delayed rules

mstormi:
ok now?

functions/openhab.bash
```diff
@@ -14,14 +30,12 @@ delayed_rules() {
...
   targetDir="/etc/systemd/system/openhab2.service.d"
 
+  create_sys_dependencies
   if [[ $1 == "yes" ]]; then
     echo -n "$(timestamp) [openHABian] Adding delay on loading openHAB rules... "
-    if ! cond_redirect mkdir -p $targetDir; then echo "FAILED (prepare  directory)"; return 1; fi
-    if ! cond_redirect rm -f "${targetDir}"/override.conf; then echo "FAILED (clean directory)"; return 1; fi
-    if cond_redirect cp "${BASEDIR:-/opt/openhabian}"/includes/systemd-override.conf "${targetDir}"/override.conf; then echo "OK"; else echo "FAILED (copy configuration)"; return 1; fi
+    if cond_redirect cat "${BASEDIR:-/opt/openhabian}"/includes/delayed-rules.conf >>"${targetDir}"/override.conf; then echo "OK"; else echo "FAILED (copy configuration)"; return 1; fi
   elif [[ $1 == "no" ]]; then
```
ecdye, line 37: Please use sed or some other program to make this actually work and disable delayed rules.

mstormi: I thought about but sed is really hard to get right because there's multi-line dependencies ... I possibly would not get that right right away. So this is the first attempt at a multi-line building block approach. Works for delayed rules but no additional blocks ATM. Let's keep that and extend to multiple blocks (to use a generator function with a table of functions to include) in a future patch.

ecdye: Perhaps delete the current and restore the original `override.conf` without adding the delayed rules.

mstormi: that's what the current code does, doesn't it? It always (oninstall,  add or remove rules) creates it as the original `override.conf`, then if add it appends the rules part.

ecdye: Oh, I see what you are saying now, I did not notice that before.

PR #1071, opened 2020-07-31
remove homegear package-originated service file (use own)

mstormi:


ecdye:
It shouldn't matter, as the services in the `/etc/systemd/system/` directory should take precedence, right?

mstormi:
correct, but I feel this is a cleaner, less confusing setup

mstormi:
yes it is a bit awkward as the packages come with /lib/systemd/system/homegear.service which has been working in a long time (I believe).
It runs homegear as root (it's changing its effective uid/group) but I dislike that concept from a security pov and prefer to start it right away as an ordinary UID.

PR #1072, opened 2020-07-31
Universal fix for kernel upgrade on first boot

ecdye:

### DIFF
--- a/build-image/first-boot.bash
+++ b/build-image/first-boot.bash
@@ -33,7 +33,6 @@ if ! source "$CONFIGFILE"; then echo "FAILED (source config)"; fail_inprogress; 
 if ! source "/opt/openhabian/functions/helpers.bash"; then echo "FAILED (source helpers)"; fail_inprogress; fi
 if source "/opt/openhabian/functions/openhabian.bash"; then echo "OK"; else echo "FAILED (source openhabian)"; fail_inprogress; fi
 
-
 if [[ "${debugmode:-on}" == "on" ]]; then
   unset SILENT
   unset DEBUGMAX
@@ -130,7 +129,6 @@ else
   fi
 fi
 
-
 echo -n "$(timestamp) [openHABian] Ensuring network connectivity... "
 if tryUntil "ping -c1 www.example.com &> /dev/null || curl --silent --head http://www.example.com |& grep -qs 'HTTP/1.1 200 OK'" 30 1; then
     echo "FAILED"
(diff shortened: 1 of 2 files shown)


build-image/first-boot.bash
```diff
@@ -151,26 +149,29 @@ echo "OK"
...
+firmwareBefore="$(uname -r)"
 echo -n "$(timestamp) [openHABian] Updating repositories and upgrading installed packages... "
 apt-get install --fix-broken --yes &> /dev/null
 if [[ $(eval "$(apt-get --yes upgrade &> /dev/null)") -eq 100 ]]; then
   echo -n "CONTINUING... "
   dpkg --configure --pending &> /dev/null
   apt-get install --fix-broken --yes &> /dev/null
   if apt-get upgrade --yes &> /dev/null; then
-    if is_pi; then
+    if [[ $firmwareBefore != "$(uname -r)" ]]; then
```
mstormi, line 160: will it change the output of uname -r without a reboot ?

BTW why apt-get --yes upgrade twice ?

ecdye: Yes it will, because it will fail the first time if any packages are broken so it uses `dpkg` to fix any broken packages if it does fail and then completes the upgrade fully the second time. It is essentially a safety precaution that was implemented before I joined Dev and prevents weird edge case errors so I left it in.

mstormi: > Yes it will.

Did you test that on real HW ?
Sorry for being sceptic but better safe than sorry. I had to re-build the image three times until it worked properly, and you're about to change that piece of code now.

ecdye: It should, I will double check now.

ecdye: It didn't but I found a different solution.

build-image/first-boot.bash
```diff
@@ -151,26 +149,29 @@ echo "OK"
...
+firmwareBefore="$(dpkg -s raspberrypi-kernel | grep "Version:[[:space:]]")"
 echo -n "$(timestamp) [openHABian] Updating repositories and upgrading installed packages... "
 apt-get install --fix-broken --yes &> /dev/null
 if [[ $(eval "$(apt-get --yes upgrade &> /dev/null)") -eq 100 ]]; then
   echo -n "CONTINUING... "
   dpkg --configure --pending &> /dev/null
   apt-get install --fix-broken --yes &> /dev/null
   if apt-get upgrade --yes &> /dev/null; then
-    if is_pi; then
+    if [[ $firmwareBefore != "$(dpkg -s raspberrypi-kernel | grep "Version:[[:space:]]")" ]]; then
```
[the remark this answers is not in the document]
ecdye: No it can't. Not to mention if that were to happen, the user would likely have a broken system anyway because of a incomplete kernel upgrade.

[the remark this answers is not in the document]
mstormi: ok then

PR #1074, opened 2020-08-03
Add absolute link to frontail bootstrap css

ecdye:

### DIFF
--- a/includes/openhab2-override.conf
+++ b/includes/openhab2-override.conf
@@ -1,5 +1,3 @@
 [Unit]
 Description=openHAB2 instance, reachable at http://%H:8080
-Wants=zram-config.service
-Wants=frontail.service
-
+Wants=zram-config.service frontail.service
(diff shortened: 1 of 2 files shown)


includes/openhab2-override.conf
```diff
@@ -1,5 +1,3 @@
 [Unit]
 Description=openHAB2 instance, reachable at http://%H:8080
-Wants=zram-config.service
-Wants=frontail.service
-
+Wants=zram-config.service frontail.service
```
mstormi, line 3: are you sure this works? I looked around the Inet when I've added it but did not find a confirmation so went the safe route (it's also better to have separate lines should we decide to dynamically generate/change this one day via script)

ecdye: I looked at the MAN page and it said it works.

mstormi: Please validate first.

mstormi: you mean you did validate?

ecdye: Not yet. I will this afternoon.

ecdye: It works, no issues for me.

PR #1055, opened 2020-07-26
Finish Standardization and Rewrite

ecdye:
Fixes #937
Fixes #625

Signed-off-by: Ethan Dye <mrtops03@gmail.com>

Issue #625, opened 2019-08-04
Drop fixed path in favor of relative when calling binaries


EliasGabrielsson:
In order to get some more headroom for folder changes should the scripts in what extent is possible use relative calling on binaries. i.e. `git` in favor of `/local/bin/git`.

- [x] Update CONTRIBUTE.md with coding style
- [ ] Review all files in the whole repository

mstormi:
For proper understanding: you mean in DISfavor of using the whole path, i.e. just use "git" in our scripts - correct ?

PR #1077, opened 2020-08-05
auto backup

### LINKED ISSUE #1044: Auto-backup
Just an idea to track and discuss

We equip a Pi with a say 16GB internal SD card plus an external card reader(!) with a 64GB one.
First two partitions on the external card will be mirrors of the two partitions on the internal one, and the rest will be dedicated to provide backup storage for use by Amanda.
Also add cronjobs to rsync config and persistence DB daily and copy partitions say monthly (eventually mount+unmount after).
Now if the internal card fails for whatever reason, we have a ready-to-run backup on the external card, including daily snapshots of persistence and other data in ZRAM(!) that otherwise gets lost.

Implementation: Make card reader device a parameter in openhabian.conf.
In unattended install, check for partition sizes, partition external stick and call amanda_setup with suitable parameters. 

wdyt?

mstormi:
Closes #1044

Signed-Off-By: Markus Storm markus.storm@gmx.net


holgerfriedrich:
@mstormi I am in doubt that copying a mounted (writable) partition of a running system on block basis (dd) could work. You will not get a snapshot, but likely in inconsistent state of the filesystem (it might still look ok and by mountable, but the content of the files might point to something completely different due to modifications of the fs during copy).

My recommendation would be to try an rsync (file basis) with some special handling of system directories plus some extra efforts to get it bootable etc.

mstormi:
Thanks for the warning. Yes I know all those horror stories as I've been doing backups for >20 years in different organizations.
But I'm sure it works fine. I've successfully dd'ed `/dev/mmcblk0` a couple of times in the past on my production system and never had issues.
First, it's done during low-traffic times, and most remaining traffic will be on ZRAM'ed directories (the lower FS which is what I dump is not changed during that time).
Also, if you look at the code (beware, that's rapidly prototyped only so not ready yet), you'll see that I prepared for a combination of dd and rsync. My idea is to do raw dumps just once a month and rsync just the diffs all the other days.

mstormi:
Fixes #1044

ecdye:
While making this be sure to keep #625 in mind as it will soon be the standard across all the code once #1055 is merged.

ecdye:
We should create a size chart for storage space recommendations for the backup device. with descriptions of how much space will be used for each piece of the backup solution.

ecdye:
Also, why not use Amanda to create the RAW disk backups?

mstormi:
> Also, why not use Amanda to create the RAW disk backups?

Thought about that but that would require additional storage space thus raise HW requirements

mstormi:
Ok this was quite some twiddling but I think I made it.
Please get yourself an external card reader and an SD card at least twice the size of the internal card and test this feature by activating any of the storageXXX options in (/boot/)openhabian.conf. Eventually you can also use the 50 menu options.

functions/backup.bash
```diff
@@ -237,9 +215,156 @@ amanda_setup() {
...
+      dest="/dev/$retval"
+    else
+      dest=${dest}2
+    fi
+    # Auswahl des src dir nötig ?? nein immer "/"
+    # Auswahl des dest dir nötig ?? nein immer syncmount und dest wird oben ausgewählt
+    #if [[ -n "$INTERACTIVE" ]]; then
+    mount "$dest" "${syncMount:-/storage/syncmount}"
+    rsync --one-file-system -avRh "/" "$syncMount"
+    if ! umount "$syncMount" &> /dev/null; then
```
ecdye, line 270: Use `()` in if statements if output redirection is used to help avoid ambiguity.

functions/backup.bash
```diff
@@ -237,9 +215,156 @@ amanda_setup() {
...
 
+  if [[ "$1" == "diff" ]]; then
+    if [[ ! -d $syncMount ]]; then
+      mkdir -p "$syncMount"
+    fi
+    if [[ -n "$INTERACTIVE" ]]; then
+      select_blkdev "^-sd" "select partition" "Select the partition to copy the internal SD card data to"
+      dest="/dev/$retval"
+    else
+      dest=${dest}2
```
ecdye, line 263: Double quote all variable assignments.

functions/backup.bash
```diff
@@ -237,9 +215,156 @@ amanda_setup() {
...
+##
+##   setup_mirror_SD()
+##
+setup_mirror_SD() {
+  local dest
+  local srcSize
+  local destSize
+  local minSize
+  local targetDir="/etc/systemd/system/"
+  local storageDir=${storagedir:-/storage}
```
ecdye, line 288: Double quote all variable assignments.

functions/backup.bash
```diff
@@ -237,9 +215,156 @@ amanda_setup() {
...
+    echo "FAILED"; return 1
+  fi
+
+  # shellcheck disable=SC2154
+  if [[ -n "$UNATTENDED" ]] && [[ -z "$backupdrive" ]]; then return 0; fi
+
+  if [[ -n "$INTERACTIVE" ]]; then
+    select_blkdev "^sd" "Setup SD mirroring" "Select USB device to copy the internal SD card data to"
+    if [[ -z "$retval" ]]; then return 0; fi
+    dest="/dev/$retval"
```
ecdye, line 314: Use `{}` in a variable assignment involving another variable to avoid ambiguity.

functions/backup.bash
```diff
@@ -237,9 +215,156 @@ amanda_setup() {
...
+  else
+    # shellcheck disable=SC2154
+    dest=${backupdrive:-/dev/sda}
```
ecdye, line 317: Double quote all variable assignments.

functions/backup.bash
```diff
@@ -237,9 +215,156 @@ amanda_setup() {
...
+  fi
+
+  size=$(fdisk -l /dev/mmcblk0 | head -1 | cut -d' ' -f3)	# in GBytes
+  infoText="$infoText1 $dest $infoText2"
+  srcSize=$(blockdev --getsize64 /dev/mmcblk0)
+  minSize="$((19 * srcSize / 10))"	# to accomodate for slight differences in SD sizes
+  destSize=$(blockdev --getsize64 "$dest")
```
ecdye, line 324: Double quote all variable assignments.

functions/backup.bash
```diff
@@ -237,9 +215,156 @@ amanda_setup() {
...
+  if [[ "$destSize" -lt "$minSize" ]]; then
+    if [[ -n "$INTERACTIVE" ]]; then
+      whiptail --title "insufficient space" --msgbox "$sizeError" 9 80
+    fi
+    echo "FAILED"; return 1;
+  fi
+
+  # copy partition table and create storage partition for the rest
+  sfdisk -d /dev/mmcblk0 | sfdisk "$dest"
+  start=$(fdisk -l /dev/mmcblk0 | head -1 | cut -d' ' -f7)
```
ecdye, line 334: Double quote all variable assignments.

functions/backup.bash
```diff
@@ -237,9 +215,156 @@ amanda_setup() {
...
+w
+EOF
+
+  partprobe
+  cond_redirect mke2fs -F -t ext4 "${dest}3"
+  mkdir -p "${storageDir}"
+  if ! sed -e "s|%DEVICE|${backupdrive:-/dev/sda}3|g" -e "s|%BKPDIR|${storageDir}|g" "${BASEDIR:-/opt/openhabian}"/includes/storage.mount >${targetDir}/storage.mount; then echo "FAILED (create storage mount)"; fi
+  if ! cond_redirect systemctl enable --now storage.mount; then echo "FAILED (enable storage mount)"; return 1; fi
+
+  size=$(fdisk -l "${dest}3" | head -1 | cut -d' ' -f3)
```
ecdye, line 353: Double quote all variable assignments.

functions/backup.bash
```diff
@@ -237,9 +215,156 @@ amanda_setup() {
...
+##    mirror_SD(String method)
+##
+mirror_SD() {
+  local src
+  local dest
+  local start
+  local syncMount="/storage/syncmount"
+
+
+  src=/dev/mmcblk0
```
ecdye, line 239: Double quote all variable assignments.

functions/backup.bash
```diff
@@ -237,9 +215,156 @@ amanda_setup() {
...
+
+
+## "raw" copy partition using dd or mount and use rsync to sync "diff"
+## Valid arguments: "raw", "diff"
+## Periodically activated by systemd timer units (raw copy on 1st of month, rsync else)
+##
+##    mirror_SD(String method)
+##
+mirror_SD() {
+  local src
```
ecdye, line 233: Stick to one variable declaration style per function, either one line or declare and assign on different lines.

docs/NEWSLOG.md
```diff
@@ -1,5 +1,12 @@
+## August 15, 2020
+## Auto-backup (ALPHA)
+Have openHABian automatically take daily syncs of your internal SD card to
```
ecdye, line 3: openHABian can automatically take daily backup syncs of your internal SD card to another device in an external port. This allows for fast swapping of cards to reduce impact of a failed SD card. The external device will also be used to setup an Amanda backup system.

openhabian-setup.sh
```diff
@@ -108,6 +108,8 @@ if [[ -n "$UNATTENDED" ]]; then
   clean_config_userpw
   frontail_setup
   zram_setup
+  amanda_setup
```
ecdye, line 111: Remove, as it is not necessary.

functions/backup.bash
```diff
@@ -154,75 +138,69 @@ create_backup_config() {
 }
 
 amanda_setup() {
+  local matched
```
ecdye, line 141: Remove changes from previous iteration of this PR that are no longer necessary.

functions/helpers.bash
```diff
@@ -411,3 +411,19 @@ wait_for_apt_to_finish_update() {
...
+##    select_blkdev(String, String, String)
+##
+select_blkdev() {
+    declare -a array=()
+    while read -r id foo{,} size foo{,,}; do
+      array+=("$id"     "$size" )
+    done < <(lsblk -i | tr -d '`\\|' | grep -E "${1}" | tr -d '\\-')
+
+    # shellcheck disable=SC2034
+    retval=$(whiptail --title "$2" --cancel-button Cancel --ok-button Select --menu "$3" 12 76 4 "${array[@]}" 3>&1 1>&2 2>&3)
```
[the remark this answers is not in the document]
mstormi: ok (although you shouldn't call this if you're not interactive)

functions/backup.bash
```diff
@@ -237,9 +215,156 @@ amanda_setup() {
...
+    if [[ ! -d $syncMount ]]; then
+      mkdir -p "$syncMount"
+    fi
+    if [[ -n "$INTERACTIVE" ]]; then
+      select_blkdev "^-sd" "select partition" "Select the partition to copy the internal SD card data to"
+      dest="/dev/$retval"
+    else
+      dest=${dest}2
+    fi
+    # Auswahl des src dir nötig ?? nein immer "/"
```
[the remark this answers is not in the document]
mstormi: just forgot to remove

functions/system.bash
```diff
@@ -236,9 +236,7 @@ create_mount() {
 
   echo -n "$(timestamp) [openHABian] Creating mount $source in '/srv/openhab2-${destination}'... "
   if ! sed -e 's|%SRC|'"${source}"'|g; s|%DEST|'"${destination}"'|g' "${BASEDIR:-/opt/openhabian}"/includes/mount_template > /etc/systemd/system/"$mountPoint"; then echo "FAILED (sed)"; return 1; fi
-  cond_redirect systemctl -q daemon-reload &> /dev/null
-  if ! cond_redirect systemctl enable "$mountPoint"; then echo "FAILED (enable service)"; return 1; fi
-  if cond_redirect systemctl restart "$mountPoint"; then echo "OK"; else echo "FAILED (restart service)"; return 1; fi
+  if ! cond_redirect systemctl enable --now "$mountPoint"; then echo "FAILED (enable service)"; return 1; fi
```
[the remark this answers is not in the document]
mstormi: it's a short form of the previous 2 lines, I don't recall why they're in

includes/sdrsync.service
```diff
@@ -0,0 +1,9 @@
+[Unit]
+Description=mount external card reader device and rsync / over there
+
+[Service]
+Type=oneshot
+#User=root
+#Group=root
+ExecStart=source /opt/openhabian/functions/backup.bash; mirror_SD "diff" %DEST
```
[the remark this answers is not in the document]
mstormi: Frankly, I don't know yet

includes/storage.mount
```diff
@@ -0,0 +1,14 @@
+[Unit]
+Description=%BKPDIR mount
+Before=zram-config.service
+
+[Mount]
+What=%DEVICE
+Where=%BKPDIR
```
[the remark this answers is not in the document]
mstormi: fstab has been given up earlier already. Reason is some mounts must happen before and some after specific services, mount units allow to be more fine-grained.

functions/backup.bash
```diff
@@ -237,9 +215,156 @@ amanda_setup() {
...
+    fi
+  fi
+}
+
+
+## setup mirror/sync of boot and / partitions
+##
+##   setup_mirror_SD()
+##
+setup_mirror_SD() {
```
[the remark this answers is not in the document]
mstormi: ok

functions/backup.bash
mstormi: you're more insisting than shellcheck ;-) but ok

functions/backup.bash
mstormi: we had that already. I use the short form for constants, longer else

docs/NEWSLOG.md
mstormi: ok

functions/backup.bash
mstormi: huh? "matched" is used or what do you mean ?

openhabian-setup.sh
mstormi: it is! will return if $backupdrive unset

functions/backup.bash
```diff
@@ -237,9 +215,156 @@ amanda_setup() {
...
+    if [[ -n "$INTERACTIVE" ]]; then
+      whiptail --title "insufficient space" --msgbox "$sizeError" 9 80
+    fi
+    echo "FAILED"; return 1;
+  fi
+
+  # copy partition table and create storage partition for the rest
+  sfdisk -d /dev/mmcblk0 | sfdisk "$dest"
+  start=$(fdisk -l /dev/mmcblk0 | head -1 | cut -d' ' -f7)
+  fdisk "$dest" &> /dev/null <<EOF
```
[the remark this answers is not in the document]
mstormi: no there's no feasible alternative I know of.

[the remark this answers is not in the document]
ecdye: Better yet, can you explain exactly what this command is doing? Especially since there is no real way to do it otherwise, add it to the comments in the code to help explain what is going on.

includes/storage.mount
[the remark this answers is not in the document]
ecdye: Fair enough.

functions/backup.bash
ecdye: Its all about style :)

openhabian-setup.sh
ecdye: But it does not actaully setup the amanda backup? right?

functions/backup.bash
ecdye: I was referring to the function in general, in a sense from looking at the comment in `openhabian-setup.sh`

functions/backup.bash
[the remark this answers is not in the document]
mstormi: see comment you even quoted

openhabian-setup.sh
mstormi: yes it does, with default params

functions/backup.bash
mstormi: be more specific please what would need to be removed?

functions/backup.bash
mstormi: what's ambiguous about that ?
() is subshell. Not sure it would work the same.

ecdye: it does.

openhabian-setup.sh
ecdye: But then it would be overwritten by the mirror SD which creates another amanda config, right?

functions/backup.bash
mstormi: ok then

openhabian-setup.sh
mstormi: you're right, I removed that call in setup_mirror_SD

ecdye: I would actually prefer it to take place there, it makes incompatibility with #1055 much simpler.

functions/backup.bash
ecdye: Never-mind, it was confusion because of other changes.

openhabian-setup.sh
mstormi: I understand but that sort of nesting is a no-go to me from a SW architecture POV.
All install routines should be capable of being called directly and without dependencies on each other, for both, unattended and interactive install, and handle interativeness (or not) themselves.

functions/backup.bash
mstormi: ok

openhabian-setup.sh
ecdye: It's not really nesting. The create Amanda config function is intended for that purpose and should function just fine. Whereas the Amanda setup function should be used for an interactive setup.

mstormi: No it's insufficient to create the config only, and the code for the missing parts exists in amanda_setup() and used to be duplicated in mirror_SD. Not desirable from the SW arch perspective either.

ecdye: SW Arch?

mstormi: software architecture
Ok a big word for little code like this but still valid IMHO

ecdye: Very well, however for a more object oriented design, it might be better to create a new function to install and setup necessary Amanda components (ie `amanda_install`). And then keep `amanda_setup` as a interactive configuration setup and use those to create a more modular and simpler design.

mstormi: yeah but bash programming isn't OO, pure imperative procedural that is, and I don't want to start mixing styles either.

ecdye: But in this case I think it works out to a good compromise and allows for much more modular code.

mstormi: In the first place, I'm happy to have it working and I don't see the point in (more) modularization there so I don't want to refactor that upfront ... but I'm ok with when you change it after the initial merge as part of your ongoing standardization efforts. Are you ok with that, too ?

ecdye: Fair enough.

mstormi:
did anyone test ?
Please do, use this testbuild branch image

includes/sdrsync.service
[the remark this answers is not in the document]
mstormi: it didn't :) but now that I fixed it I think it does

PR #1086, opened 2020-08-10
build.bash improvements

mstormi:
build on pi, too
intro TMPDIR
trap on exit/err to remove mounts and temp folder

### DIFF
--- a/build.bash
+++ b/build.bash
@@ -111,7 +111,7 @@ check_command_availability_and_exit() {
 
 # mount rpi image using userspace tools, in docker use privileged mount via kpartx
 mount_image_file_boot() { # imagefile buildfolder
-  if ! running_in_docker && ! running_on_github; then
+  if ! running_in_docker && ! running_on_github || is_pi; then
     guestmount --format=raw -o uid=$EUID -a "$1" -m /dev/sda1 "$2/boot"
   else
     loop_prefix=$(kpartx -asv "$1" | grep -oE "loop([0-9]+)" | head -n 1)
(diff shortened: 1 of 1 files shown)


build.bash
```diff
@@ -177,6 +186,7 @@ EOF
 #### Build script start ####
 ############################
 
+trap cleanup_build EXIT ERR
```
holgerfriedrich, line 189: installing the trap before setting $buildfolder might introduce some problems (actually unmounting `/boot` and `/root`...

build.bash
```diff
@@ -100,7 +109,7 @@ check_command_availability_and_exit() {
 
 # mount rpi image using userspace tools, in docker use privileged mount via kpartx
 mount_image_file_boot() { # imagefile buildfolder
-  if ! running_in_docker && ! running_on_github && ! is_pi; then
+  if ! running_in_docker && ! running_on_github; then
```
holgerfriedrich, line 112: why removing the is_pi here? I think the guestmount package which is based on qemu/libvirt which is not available on the PI?

holgerfriedrich: the intention was to use "guestmount" on normal Linux PCs (in user mode), and for all others (including the RPI) the classic losetup/kpartx way (root required).... 
Do you actually build the images on a PI?

mstormi: yes of course

build.bash
mstormi: that's supposed to work, $buildfolder will be evaluated at runtime.
But I've added a check in cleanup_build() for $buildfolder empty just in case

build.bash
holgerfriedrich: I have the feeling that using classic ("root") way is better suited for building on the PI. Typical installations do not provide libguestfs-tools. For my production system, this would need ~600MB of extra packages to install.
But actually, I do not mind at all. Keep it or remove it, as you like....

mstormi: I wasn't aware of that so changed it back to build as root

build.bash
```diff
@@ -100,7 +111,7 @@ check_command_availability_and_exit() {
 
 # mount rpi image using userspace tools, in docker use privileged mount via kpartx
 mount_image_file_boot() { # imagefile buildfolder
-  if ! running_in_docker && ! running_on_github && ! is_pi; then
+  if ! running_in_docker && ! running_on_github || is_pi; then
```
holgerfriedrich: && !   is not the same as ||,   why changing the logic here?

mstormi: typo, changed

PR #1124, opened 2020-09-15
Use lite image for arm64 builds

ecdye:
Raspberry Pi has provided a lite image for arm64 and as such, like with
our armhf builds we will prefer it over the full image as it is less
bloated with things we don't need for openHABian images.


mstormi (approved):
a smallish contribution to save the earth
well or watch another hour of Netflix instead ;)

Issue #316, opened 2018-01-16
Frontail service: exchange user


ThomDietrich:
Related to https://github.com/openhab/openhab-linuxpkg/issues/108
Turns out using the user `openhab` as the user for the frontail service was a bad idea. We should switch to one of the default system users. @mhaack what do you think?

CC @BClark09

mhaack:
@ThomDietrich I have no strong opinion on that. It might help to mitigate https://github.com/openhab/openhab-linuxpkg/issues/108, but other components like homegear or node red also run with the `openhab` user. There is no guarantee to solve that.

mstormi:
still relevant ? If no please close.

ThomDietrich:
@BClark09 what's the latest on the openHAB side of things?

mstormi:
pinging @BClark09

mstormi:
@BClark09 openhab user still in use.
What about making frontail PartOf=openhab2.service in systemd ?

BClark09:
Sorry, I've let this slip. I agree that using the `openhab` user might not be the best approach for services that aren't directly associated with `openhab`. The easiest fix, and one I really should have implemented a long time a go is the one we discuss in   openhab/openhab-linuxpkg#108

mstormi:
> , but I'm not sure having a list of overrides in different systemd files is very maintainable.

Well that just applies to openHABian, and for anything not-openhab2 that's no packages you would need to fiddle tohat into but just files we place during installation anyway, or am I wrong there ?
And is there more than frontail.service ?

mstormi:
@BClark09 does closing openhab/openhab-linuxpkg#108 mean this now works and I can close it, too?

BClark09:
The issue fixed in linuxpkg was about the errors caused when another service uses the `openhab` user. 

Ideally, these services should use a default or new user to manage their permissions. So there is still something to improve on openHABian's end I think.

ecdye:
If I understand this correctly. What you are saying is that we should use a new user for frontail perhaps even just a new user called frontail? 

If so then it should be an easy fix right?

mstormi:
That's how I, too, understood it. just go for it

mstormi:
upps, wrong button

PR #1133, opened 2020-09-18
use dedicated user to run frontail

mstormi:
Fixes: #316 


ecdye (approved):
LGTM, I did not test however, everything looks fine so I have no issue with merging if you tested.

mstormi:
I tested but as that's not a full end-2-end testing something might have slipped.
As frontail is apparently in widespread use I guess we'll quickly get a feedback if there's anything wrong with this patch.
So you have been warned :)

PR #1142, opened 2020-09-21
fix PARTUUID after dd copy of raw device && skip ZRAM on RPi4 8GB

mstormi:
To avoid that the external partition is mounted on boot because it has the same PARTUUID as the internal SD.

Apologies for the zram issue I mixed in (I forgot to branch again).
Everybody with a 8GB RPi seems to have this problem for the time being so I thought I'd drop ZRAM on this HW until a fix or workaround is found.

### DIFF
--- a/functions/backup.bash
+++ b/functions/backup.bash
@@ -378,7 +378,7 @@ mirror_SD() {
   if [[ "$1" == "raw" ]]; then
     echo "Creating a raw partition copy, be prepared this may take long such as 20-30 minutes for a 16 GB SD card"
     if ! cond_redirect dd if="${src}" bs=1M of="${dest}"; then echo "FAILED (raw device copy)"; dirty="yes"; fi
-    if ! cond_redirect set-partuuid "${dest}" random; then echo "FAILED (set random PARTUUID)"; dirty="yes"; fi
+    if ! cond_redirect echo "y" | set-partuuid "${dest}" random; then echo "FAILED (set random PARTUUID)"; dirty="yes"; fi
     if ! cond_redirect fsck -y -t ext4 "${dest}2"; then echo "OK (dirty bit on fsck ${dest}2 is normal)"; dirty="yes"; fi
     if [[ "$dirty" == "no" ]]; then
       echo "OK"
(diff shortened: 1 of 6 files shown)


functions/backup.bash
```diff
@@ -377,7 +377,8 @@ mirror_SD() {
   fi
   if [[ "$1" == "raw" ]]; then
     echo "Creating a raw partition copy, be prepared this may take long such as 20-30 minutes for a 16 GB SD card"
-    if ! cond_redirect dd if="${src}" bs=1M of="${dest}"; then echo "FAILED (raw device copy)"; return 1; fi
+    if ! cond_redirect dd if="${src}" bs=1M of="${dest}"; then echo "FAILED (raw device copy)"; dirty="yes"; fi
+    if ! cond_redirect echo "y" | set-partuuid "${dest}" random; then echo "FAILED (set random PARTUUID)"; dirty="yes"; fi
```
ecdye, line 381: don't use `echo "y"` rather use the `yes` command

includes/set-partuuid
```diff
@@ -0,0 +1,196 @@
+#!/bin/bash
+
+# shamelessly copied from https://www.raspberrypi.org/forums/viewtopic.php?f=29&t=196778
```
ecdye, line 3: Are we allowed to do this from a licensing standpoint? Just checking.

ecdye (changes requested):

includes/set-partuuid
mstormi: Dunno. It's a post offering the code for download, with no license to apply (at least I have not seen any).
So crediting the origin is _more_ than the author asked for in terms of licensing, isn't it ?

functions/backup.bash
mstormi: didn't know it. at your own risk ...

includes/set-partuuid
ecdye: Sure, it is now. I don't think it matters too much I was more of just curious.

functions/backup.bash
ecdye: Yep

ecdye (approved):
LGTM

PR #1144, opened 2020-09-21
Set MQTT values from default file for FIND3

ecdye:

### DIFF
--- a/functions/find.bash
+++ b/functions/find.bash
@@ -113,7 +113,7 @@ find3_setup() {
   if ! (id -u find3 &> /dev/null || cond_redirect useradd --groups openhabian find3); then echo "FAILED (adduser)"; return 1; fi
   if ! cond_redirect chown -R find3:openhabian /opt/find3; then echo "FAILED (permissions)"; return 1; fi
   if ! cond_redirect install -m 644 "${BASEDIR:-/opt/openhabian}"/includes/find3ai.service /etc/systemd/system/find3ai.service; then echo "FAILED (copy service)"; return 1; fi
-  if ! (sed -e 's|%MQTT_SERVER|'"${MQTT_SERVER}"'|g; s|%MQTT_ADMIN|'"${MQTT_ADMIN}"'|g; s|%MQTT_PASS|'"${MQTT_PASS}"'|g; s|%FIND3_PORT|8003|g' "${BASEDIR:-/opt/openhabian}"/includes/find3server.service > /etc/systemd/system/find3server.service); then echo "FAILED (service file creation)"; return 1; fi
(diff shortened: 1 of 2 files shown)


functions/find.bash
```diff
@@ -196,7 +196,7 @@ setup_monitor_mode() {
       echo "$(timestamp) [openHABian] Monitor Mode patch has not been updated for kernel version 5.4 on your RPi... CANCELED"
       return 0
     elif is_pithreeplus || is_pifour; then
-      firmwareVersion="bcm43455c0/7_45_189"
+      firmwareVersion="bcm43455c0/7_45_206"
```
mstormi, line 199: uhh. Now I don't know the context but static firmware strings are not an optimum solution

includes/find3server.service
```diff
@@ -10,7 +10,7 @@ RuntimeDirectory=find3server
 RuntimeDirectoryMode=0775
 User=find3
 Group=openhabian
-ExecStart=/usr/sbin/find3server -mqtt-server %MQTT_SERVER -mqtt-admin %MQTT_ADMIN -mqtt-pass %MQTT_PASS -port %FIND3_PORT
+ExecStart=/usr/sbin/find3server -port %FIND3_PORT
```
mstormi, line 13: Are you sure it'll take the remaining parameters off /etc/default/find3server ? Why ? And why don't you put the port number there, too ?

functions/find.bash
ecdye: It is required because of the way that nexmon does their versioning and patches. Dynamic firmware would result in issues because they only support certain versions.

includes/find3server.service
ecdye: It does, I tested. I looked at the source code and it does not support setting the port from the env.

functions/find.bash
mstormi: just sayin' . I'm sure you would have chosen a better means if it existed.

ecdye: Believe me, I wanted to. I just couldn't find any way to do it with the way nexmon does their patches.

includes/find3server.service
mstormi: I'm confused. Are you saying it takes all those MQTT_* env variables but not the port ?

[18:49:17] root@openhab:/opt/openhabian#  cat /etc/default/find3server
# FIND server service configuration
# see https://www.internalpositioning.com/server/
#
MQTT_SERVER=openhabianpi:1883
MQTT_ADMIN=openhab2
MQTT_PASS=

ecdye: Yes, that is what I am saying.

mstormi: well ok then
can we use $(grep PORT /etc/default/find3server) or similar in .service ?
if no I'm fine with this, too.

ecdye: No we can't, that is a bash substitution which won't work in .service files. I am not aware of any way that would work in .service files.

Issue #1151, opened 2020-09-23
dd / setup mirroring (option 53) kills partition 3 ?


mstormi:
We dd the (complete) raw device /dev/mmcblk0 to the destination raw device (/dev/sda or some such).
There's hints that any third partition to exist on the external device (such as the partition that auto backup sets up for Amanda) are gone after using dd this way.
Does this overwrite the partition table?
If no, would it still work if the internal SD is (slightly) larger than the external one ? That would probably overwrite partition 3 with the tail end of internal's partition 2.

If so:
~~Change to dump partition 2 only~~
~~Verify not to dump if src partition(not device !) larger than dest.~~

~~Run dd semiannually, add to send mail to admin when running~~
~~Run rsync weekly only~~

PR #1121, opened 2020-09-14
Add inital offline install capabilities

ecdye:
Supersedes #985
Fixes #984

Signed-off-by: Ethan Dye <mrtops03@gmail.com>

PR #1152, opened 2020-09-23
modify SD dump to dump single partitions only 

mstormi:
Fixes: #1151  
(hopefully)
raw device partitions 1 (/boot) and 2 (/) only to avoid overwriting partition table or subsequent partitions

### DIFF
--- a/functions/backup.bash
+++ b/functions/backup.bash
@@ -355,7 +355,6 @@ mirror_SD() {
   local dirty="no"
   local dumpInfoText="For your information as the operator of this openHABian system:\\nA timed background job to run semiannually has just created a full raw device copy of your RPI's internal SD card.\\nOnly partitions to contain openHABian (/boot and / partitions 1 & 2) were copied."
   local partUUID
-  local origUUID
 
   if [[ "${src}" == "${dest}" ]]; then
     echo "FAILED (source = destination)"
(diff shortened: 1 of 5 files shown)


ecdye:
Did you test and confirm it works for you?

functions/backup.bash
```diff
@@ -353,7 +353,10 @@ mirror_SD() {
   local storageDir="${storagedir:-/storage}"
   local syncMount="${storageDir}/syncmount"
   local dirty="no"
-  
+  local dumpInfoText="For your information as the operator of this openHABian system:\\nA timed background job to run semiannually has just created a full raw device copy of your RPI's internal SD card.\\nOnly partitions to contain openHABian (/boot and / partitions 1 & 2) were copied."
```
ecdye, line 356: Twice a year?

mstormi: yep. We can change that later. Just killed an SD card so got more cautious, rsync also just once a week

mstormi:
> 
> 
> Did you test and confirm it works for you?

still testing

mstormi:
it doesn't work out.
you can merge the code but it doesn't boot from the alternative disk.
Then again, the previous code didn't either.
Trouble is the duplicated UUID when both cards are in (and they need to be else we cannot automatically keep the external one up to date). It's coincidence which device is booted from when both are in.

ecdye:
Huh, odd...  Make sure you fix shellcheck errors before merging

mstormi:
Found the issue, fstab also references PARTUUIDs not instead of device names so I had to replace those entries, too.
Right now busy testing.

mstormi:
It's working fine now. Please review again and also test when your reader has arrived.

ecdye:
> It's working fine now. Please review again and also test when your reader has arrived.

I actually have them here now and have started testing, I just have not had enough time to test this yet as I have been trying to finish up #1121

functions/backup.bash
```diff
@@ -375,30 +369,56 @@ mirror_SD() {
...
+    #origPartUUID=$(blkid "${src}p2" | sed -n 's|^.*PARTUUID="\(\S\+\)".*|\1|p')
+    origPartUUID=$(blkid "${src}p2" | sed -n 's|^.*PARTUUID="\(\S\+\)".*|\1|p' | sed -e 's/-02//g')
+    if ! partUUID=$(yes | cond_redirect set-partuuid "${dest}2" random | awk '/^PARTUUID/ { print substr($7,1,length($7) - 3) }'); then echo "FAILED (set random PARTUUID)"; dirty="yes"; fi
+    if ! cond_redirect tune2fs "${dest}2" -U random; then echo "FAILED (set random UUID)"; dirty="yes"; fi
+    mount "${dest}1" "$syncMount"
+    sed -i "s|${origPartUUID}|${partUUID}|g" "${syncMount}"/cmdline.txt
+    umount "$syncMount"
+    mount "${dest}2" "$syncMount"
+    sed -i "s|${origPartUUID}|${partUUID}|g" "${syncMount}"/etc/fstab
+    sed -i 's|^What=.*|What=/dev/mmcblk0p3|g' "${syncMount}/etc/systemd/system/${storageDir}".mount
```
ecdye, line 397: Will this get reset by syncs?

ecdye (approved):
LGTM

functions/backup.bash
mstormi: This code *is* what is being executed on every sync so it's effectively kept.
That's the storage mount which would point to the (no more existing) external SD after we had to exchange cards.

ecdye: OK, I just misread the code.

PR #1157, opened 2020-09-25
Fix primary group

### LINKED ISSUE #1140: user openhabian not in group openhab ?
should it be in there?
User `openhabian` has primary group `openhabian` so created files belong to that group.
So should we add `openhab` user to `openhabian` group or should the user `openhabian` be given `openhab` as the primary group ?

in first-boot.bash, the existing user gets renamed and assigns the user-provided name to BOTH, user AND group.
Shouldn't it rather assign to group "openhab" and add the group with the user-provided username ?

mstormi:
Fixes #1140 

Please give it a second thought before approving

### DIFF
--- a/build-image/first-boot.bash
+++ b/build-image/first-boot.bash
@@ -62,10 +62,9 @@ echo -n "$(timestamp) [openHABian] Changing default username and password... "
 if [[ -z "${username+x}" ]] || ! id $userdef &> /dev/null || id "$username" &> /dev/null; then
   echo "SKIPPED"
 else
-  usermod -l "$username" "$userdef"
+  usermod -l "$username" -g "openhab" "$userdef"
   usermod -m -d "/home/$username" "$username"
-  groupmod -n "openhab" "$userdef"
-  usermod --append --groups "$username" "$userdef"
+  groupmod -n "$username" "$userdef"
   echo "${username}:${userpw:-$username}" | chpasswd
   echo "OK"
 fi
--- a/build-image/first-boot.bash
+++ b/build-image/first-boot.bash
@@ -62,7 +62,9 @@ echo -n "$(timestamp) [openHABian] Changing default username and password... "
 if [[ -z "${username+x}" ]] || ! id $userdef &> /dev/null || id "$username" &> /dev/null; then
   echo "SKIPPED"
 else
+  gid=$(id -g "$userdef")
   usermod -l "$username" -g "openhab" "$userdef"
+  usermod -aG "$gid" "$username"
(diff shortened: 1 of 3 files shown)


build-image/first-boot.bash
```diff
@@ -64,7 +64,8 @@ if [[ -z "${username+x}" ]] || ! id $userdef &> /dev/null || id "$username" &> /
 else
   usermod -l "$username" "$userdef"
   usermod -m -d "/home/$username" "$username"
-  groupmod -n "$username" "$userdef"
+  groupmod -n "openhab" "$userdef"
```
ecdye, line 67: Doesn't the `openhab` group get created upon openHAB install?

build-image/first-boot.bash
```diff
@@ -64,7 +64,8 @@ if [[ -z "${username+x}" ]] || ! id $userdef &> /dev/null || id "$username" &> /
...
+  usermod --append --groups "$username" "$userdef"
```
ecdye, line 68: What exactly is `$userdef` for, I can see that it is selected based on if it is a pi or not but what for?

build-image/first-boot.bash
mstormi: right it doesn't make sense that way

build-image/first-boot.bash
mstormi: IDK for sure didn't write that... it's set some lines up.
Bottom line is that the user of that name already exists, has an environment, sudo rights etc.

build-image/first-boot.bash
ecdye: But actually, it wont work correctly without it because at this point the openhab group has not been created. That is one of the reasons that Permissions Corrections was made, right? It runs after the openhab install and adds the user to the openhab group, right?

mstormi: IDK if permissions corrections was made for that reason and IDK where exactly openhab group is created but if you're right
then I would need to move changing the primary group into permissions_corrections(), right ?

ecdye: Probably, if I understand what is happening here correctly. @ThomDietrich any insight?

mstormi: moved it to after openhab install - wdyt?

mstormi:
how to proceed ? Merge this version ?

mstormi:
@ecdye @holgerfriedrich wdyt - merge like this and validate as part of our ongoing testing ?

ecdye (approved):
LGTM, yeah it should be fine to test as part of our ongoing changes.

PR #1170, opened 2020-10-01
debug frontail BATS

mstormi:

### DIFF
--- a/functions/nodejs-apps.bash
+++ b/functions/nodejs-apps.bash
@@ -63,7 +63,7 @@ frontail_setup() {
 
   frontailBase="$(npm list -g | head -n 1)/node_modules/frontail"
 
-  if ! (id -u ${frontailUser} &> /dev/null || cond_redirect useradd --groups openhabian,openhab -s /bin/false -d /var/log/openhab2 ${frontailUser}); then echo "FAILED (adduser)"; return 1; fi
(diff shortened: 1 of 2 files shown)


mstormi:
yeah I inserted a journalctl -xu to see startup log but it says it has no journal  :(

ecdye:
Is it maybe because the `openhab` and `openhabian` groups have not been created properly and so the service is failing to start?

mstormi:
let's not guess wildly but insert debug statements that give more hints.
Although I must say I don't know what to look for ATM.

ecdye:
> let's not guess wildly but insert debug statements that give more hints.
> Although I must say I don't know what to look for ATM.

Fair enough, but I have to leave now and probably won't have time to do much more testing today so for now you are on your own.

mstormi:
ok

mstormi:
I didn't get it to work it's as if chmod/setfacl does not take effect.
So I changed login shell and homedir to /var/tmp instead.

mstormi:
So are we set for the image release ?

ecdye:
I think so, how about we let master settle until Saturday (ie try not to merge anything unless critical) and then sync master to stable and create the final image and upload and announce it.

mstormi:
Ok will do like that.
Please cross-check the docs & NEWS in the meantime.

PR #1177, opened 2020-10-08
Add admin user and group for non Raspi OS

mstormi:
A major reason for lack of Ubuntu support (and other Unices) is the absence of the admin user and its group.
It's in the RaspiOS image and we create it in CI but it's missing in the real world.

I'm not fully sure I got this fix for that right. This is why I've spent so many comments on the code.

Please give it an extra thought or three if you would also fix it that way.
Also check out the questions in there. I don't understand how it can have worked on generic Debian so far (which has no "pi" user).

### DIFF
--- a/build-image/openhabian.conf
+++ b/build-image/openhabian.conf
@@ -7,8 +7,6 @@
 hostname=openHABianDevice
 # if exists, the default user (with its working environment) will be renamed to this username given here ...
 adminusername=openhabian
-# ... assigned to this group ...
-admingroupname=openhabian
 # ... and given this password. The password will be removed from this file after completion for security reasons.
 userpw=openhabian
 # set this to download a SSH key and authorize the owner to login as the admin user
--- a/build-image/openhabian.conf
+++ b/build-image/openhabian.conf
@@ -5,8 +5,10 @@
 
 # Hostname to set this one to
 hostname=openHABianDevice
-# the default user (with its working environment) will be renamed to this username given here ...
-username=openhabian
+# if exists, the default user (with its working environment) will be renamed to this username given here ...
+adminusername=openhabian
+# ... assigned to this group ...
+admingroupname=openhabian
 # ... and given this password. The password will be removed from this file after completion for security reasons.
 userpw=openhabian
 # set this to download a SSH key and authorize the owner to login as the admin user
--- a/build-image/openhabian.conf
+++ b/build-image/openhabian.conf
@@ -5,7 +5,7 @@
 
 # Hostname to set this one to
 hostname=openHABianDevice
-# if exists, the default user (with its working environment) will be renamed to this username given here ...
+# if it exists, the default user (with its working environment) will be renamed to this username given here ...
 adminusername=openhabian
 # ... and given this password. The password will be removed from this file after completion for security reasons.
 userpw=openhabian
--- a/build-image/openhabian.conf
+++ b/build-image/openhabian.conf
@@ -6,7 +6,7 @@
 # Hostname to set this one to
 hostname=openHABianDevice
 # if it exists, the default user (with its working environment) will be renamed to this username given here ...
-adminusername=openhabian
+username=openhabian
 # ... and given this password. The password will be removed from this file after completion for security reasons.
 userpw=openhabian
 # set this to download a SSH key and authorize the owner to login as the admin user
--- a/functions/openhabian.bash
+++ b/functions/openhabian.bash
@@ -340,13 +340,11 @@ config_ipv6() {
 
 ## Create UNIX user and group for openHABian administration purposes
 ##
-##    create_user()
+##    create_user_and_group()
 ##
-create_user() {
+create_user_and_group() {
 
-  # TODO: rename to source admin.... from .conf in first-boot.bash
-  #       rework first-boot.bash to not create user there
-  #       add calling this routine in openhabian-setup.bash ()??)
+# TODO: add calling this routine in openhabian-setup.bash (or elsewhere ?)
 
 # IF
 # (1) the string/username that the end user entered as "username=" in openhabian.conf is *empty* OR
--- a/functions/openhabian.bash
+++ b/functions/openhabian.bash
@@ -370,8 +370,8 @@ create_user() {
   if ! [[ $(getent group "$groupName") ]]; then
     if ! cond_redirect groupadd "$groupName"; then echo "FAILED (add default group $groupName)"; return 1; fi
   fi
-  if ! [[ $(id -u $userName &> /dev/null) ]]; then
-    if ! cond_redirect useradd --groups "$groupName",openhab --gecos "openHABian,,,,openHAB admin user" -s /bin/bash -d /var/tmp "$userName"); then echo "FAILED (add default usergroup $userName)"; return 1; fi
+  if ! [[ $(id -u "$userName" &> /dev/null) ]]; then
+    if ! cond_redirect useradd --groups "$groupName",openhab --gecos "openHABian,,,,openHAB admin user" -s /bin/bash -d /var/tmp "$userName"; then echo "FAILED (add default usergroup $userName)"; return 1; fi
   else
     usermod --append --groups "$groupName" "$userName";
   fi
--- a/functions/openhabian.bash
+++ b/functions/openhabian.bash
@@ -365,7 +365,7 @@ create_user_and_group() {
   local userName="${adminusername:-openhabian}"
 
   if ! [[ $(id -u "$userName" &> /dev/null) ]]; then
-    if ! cond_redirect adduser --quiet --disabled-password --gecos "openHABian,,,,openHAB admin user" --shell /bin/bash --home "/home/${username}" "$userName"; then echo "FAILED (add default usergroup $userName)"; return 1; fi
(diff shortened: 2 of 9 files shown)


ecdye:
My question, `first-boot.bash` is only executed in our Docker environments and on the premade images. Therefore, none of this will change anything, I think that we need to start running a `first-boot.bash` or something similar on other platforms in order to get the desired effect. WDYT?

mstormi:
Duh, right, I didn't think of that. (something was naggin' in the back of my head while I wrote it but it didn't want to surface).
Yes we would need to have a piece of code that is guaranteed to get executed on every manual install. Not sure how we get there.
We could generate package and have this execute as a pre-install script but frankly I've never built a package so that would be a challenge for sure. @BClark09 could you stand in or at least take the lead ? 

The alternative could be to build it a check into openhabian-config startup.

mstormi:
Did you find code that queries the user to input the username that @EliasGabrielsson referred to in https://github.com/openhab/openhabian/issues/665#issuecomment-522261443 ?

mstormi:
As it's difficult to test probably best is to use an image or clonebranch at least so I created one

check out branch adminusername

mstormi:
I had forgotten to call create_user if user or group doesn't exist. Now check and eventually add in openhabian-config on every start. Is there a better location to check you can think of ?

build-image/openhabian.conf
```diff
@@ -5,8 +5,8 @@
 
 # Hostname to set this one to
 hostname=openHABianDevice
-# the default user (with its working environment) will be renamed to this username given here ...
-username=openhabian
+# if exists, the default user (with its working environment) will be renamed to this username given here ...
```
ecdye, line 8: change to 'if it exists'

functions/openhabian.bash
```diff
@@ -241,3 +241,38 @@ config_ipv6() {
...
+## Create UNIX user and group for openHABian administration purposes
+##
+##    create_user_and_group()
+##
+create_user_and_group() {
+
+# TODO: add calling this routine in openhabian-setup.bash (or elsewhere ?)
+
+# IF
+# (1) the string/username that the end user entered as "username=" in openhabian.conf is *empty* OR
```
ecdye, line 255: adminusername= not username=

functions/openhabian.bash
```diff
@@ -241,3 +241,38 @@ config_ipv6() {
...
+# (2) the default user ("pi" on RaspiOS, "openhabian" on other OS) does not exist OR
+# (3) the user whose name the end user entered as "username=" in openhabian.conf *exists* (and isn't empty because (1) applies, too)
+# THEN skip
+# ELSE rename the default user and default group to what is defined as username= in openhabian.conf
+#
+# QUESTIONS:
+# (1) will that do what we want it to
```
ecdye, line 262: I think so.

functions/openhabian.bash
```diff
@@ -241,3 +241,38 @@ config_ipv6() {
...
+# (2) did there exist a default user in generic Debian on x86 ? If no why did it work there before ? did it ?
```
ecdye, line 263: The only one i am aware of is `root` but that is not what we should be changing

functions/openhabian.bash
```diff
@@ -241,3 +241,38 @@ config_ipv6() {
...
+# (3) what happens on generic Debian on x86 ? Ubuntu on x86 ? Ubuntu,Armbian on RPi ?
```
ecdye, line 264: No idea

functions/openhabian.bash
```diff
@@ -241,3 +241,38 @@ config_ipv6() {
...
+#
+# according to Elias on non image installs the end user is queried to input the username ?
+# https://github.com/openhab/openhabian/issues/665#issuecomment-522261443
+# is he really ? Is that true for interactive installs ? I have not seen where.
```
ecdye, line 268: I don't see anything that does that, its possible it was removed.

ecdye (changes requested):
So if I understand correctly, this will create a user `openhabian` on systems if it does not exist. I have a few questions:
1. Do we know that the openhabian user will be in the sudoers file? 
2. Does this actually work, have you tested? 
3. Does this actually rename the default user?

functions/openhabian.bash
mstormi: err, yes. I wrote that comment before I renamed username to adminusername

mstormi:
> Do we know that the openhabian user will be in the sudoers file?
it's made a member to the sudo *group* which has sudo rights (see bottom of openhabian.bash)

> Does this actually work, have you tested?
yes but as that's so critical I've built an image, too. Please use this to validate.

> Does this actually rename the default user?
It renames the default user if there is one. That's on Raspi OS only. 
When there's no default or the specified adminusername does not exist it'll create it.

functions/openhabian.bash
mstormi: @holgerfriedrich can you test on Debian/x86 please ?

functions/openhabian.bash
mstormi: right

functions/openhabian.bash
mstormi: ok that was my thinking, too

functions/openhabian.bash
mstormi: corrected

build-image/openhabian.conf
mstormi: done

mstormi:
> Also @mstormi I found the place Elias was referencing, see `config.bash` under `load_create_config`

Good. So that *is* executed every time *if* there's `openhabian.conf` and if not it creates it from either `installer-config.txt` or `openhabian.conf.dist`.
Frankly I don't understand everything there. What's `/boot/installer-config.txt` good for ?

Now on interactive install (as we'll have it when you manually install openhabian, right?), it'll only query for the user if
there's no /opt/openhabian/openhabian.conf.dist - but there always is, so it'll never get to the point to ask for the user.
Am I right? Is that how we want to keep it ?

ecdye:
> > Also @mstormi I found the place Elias was referencing, see `config.bash` under `load_create_config`
> 
> Good. So that _is_ executed every time _if_ there's `openhabian.conf` and if not it creates it from either `installer-config.txt` or `openhabian.conf.dist`.
> Frankly I don't understand everything there. What's `/boot/installer-config.txt` good for ?


I think it was a way for a user to create a custom openhabian.conf without editing the normal one. @EliasGabrielsson any insight? If not we can probably remove.

> Now on interactive install (as we'll have it when you manually install openhabian, right?), it'll only query for the user if
> there's no `/opt/openhabian/openhabian.conf.dist` - but there always is, so it'll never get to the point to ask for the user.
> Am I right? Is that how we want to keep it ? Doesn't make much sense to me.

No, it just asks the user for an input username and then replaces `username` in `openhabian.conf.dist` and copies it to `/etc/openhabian.conf` right?

mstormi:
> No,

is the answer to which of my questions ?

What would you suggest to change about load_create_config()?
Other than to move my new code that creates the user there

ecdye:
About the openhabian.conf.dist always existing stopping it from running. 

Nothing really needs to be changed just update the code to work in tandem with your new method.

mstormi:
> Nothing really needs to be changed just update the code to work in tandem with your new method.

ok but I'd like to cleanup the mess, i.e. we shouldn't have multiple routines with overlapping functionality or else this section will remain to be a brain twister every time there's a new issue with the username related code.

mstormi:
would like to get this in and forward master to stable in the forseeable future

functions/openhabian.bash
```diff
@@ -24,6 +24,7 @@ get_git_revision() {
 ##
 install_cleanup() {
   echo -n "$(timestamp) [openHABian] Cleaning up... "
+  clean_config_userpw
```
ecdye, line 27: That should be run directly from `openhabian-setup.sh`, right?

functions/openhabian.bash
```diff
@@ -336,3 +337,32 @@ config_ipv6() {
...
+# (1) the string/username that the end user entered as "username=" in openhabian.conf is *empty* OR
+# (2) the default user ("pi" on RaspiOS, "openhabian" on other OS) does not exist OR
+# (3) the user whose name the end user entered as "username=" in openhabian.conf *exists*
+#     (and isn't empty because (1) applies, too)
+# THEN skip
+# ELSE rename the default user and default group to what is defined as username= in openhabian.conf
+#
+# COMMENT:
+# according to Elias on non image installs the end user is queried to input the username:
+# https://github.com/openhab/openhabian/issues/665#issuecomment-522261443
```
ecdye, line 356: Update the comments here to remove questions that have been answered.

ecdye (changes requested):
Looks pretty good, just a couple things

functions/openhabian.bash
mstormi: ok

functions/openhabian.bash
mstormi: Right, removed it here. It's already called from openhabian-setup.sh.
Frankly I do not recall my original intention for moving it.

mstormi:
ok, ready to merge then ?

ecdye (approved):
LGTM

PR #1184, opened 2020-10-13
[OH3] Update openHABian logo

ecdye:

### DIFF
--- a/includes/bash_profile
+++ b/includes/bash_profile
@@ -30,26 +30,21 @@ fi
 echo
 FireMotD --theme Gray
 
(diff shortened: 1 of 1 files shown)


mstormi:
shouldn't that go into the openHAB3 branch only ?

ecdye:
Nah, it's a logo change in general. It can be incorporated now.

mstormi:
> Nah, it's a logo change in general.

they won't change the logo on 2.5.x so we should not either I'd think.
It's also nice to have a (big) visual indication which version you're on so please put that onto OH3 branch only.

ecdye:
Ok, I'll just push this to the openHAB3 branch and close this PR.

PR #1186, opened 2020-10-14
Changed cron to systemd.timer

maikjoe:
Changed cron to systemd.timer to align documentation with implementation


mstormi:
thanks for your contribution

PR #1187, opened 2020-10-14
[OH3] Migrate to either OH3 or OH2 incl. existing files on disk

mstormi:
Installs new, removes old packages.
Changes filenames and config on disk to work with either openHAB3 or master/stable branches.
Must be part of master because on migrating (back) *to* either master or stable, all openHABian code is re-downloaded from that target branch *and script will restart* with anything overwritten we had put into the openHAB3 branch.


ecdye (commented):
I need to look into this more to make sure that this is everything but LGTM.

functions/openhabian.bash
```diff
@@ -134,10 +166,10 @@ openhabian_update() {
   echo -n "$(timestamp) [openHABian] Updating myself... "
 
   if [[ -n $INTERACTIVE ]]; then
-    if [[ $current == "stable" || $current == "master" ]]; then
-      if ! selection="$(whiptail --title "openHABian version" --radiolist "$introText" 14 80 2 stable "recommended standard version of openHABian" ON master "very latest version of openHABian" OFF 3>&1 1>&2 2>&3)"; then return 0; fi
+    if [[ $current == "stable" || $current == "master" || $current == "openHAB3" ]]; then
```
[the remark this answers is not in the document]
mstormi: See my introductory comment. Did that in my earlier version but trouble is on downgrade the script downloads master or stable and the migration code is missing there.

mstormi:
> I need to look into this more to make sure that this is everything but LGTM.

I had forgotten to call create_user if user or group doesn't exist. Now check and eventually add in openhabian-config on every start. Is there a better location to check you can think of ?

ecdye:
@mstormi Does this migrate configs too? Or do we expect the user to do that?

mstormi:
> Does this migrate configs too?

No. I believe if you remove the openhab2 package it leaves the config behind (on purge, too?) and the install scripts from the OH3 package will offer to keep or overwrite every file that changes. (@BClark09 correct ?).
But there's no downgrade script to do the inverse. So all we can do is ex+reimport the config.

User needs to do that currently. I'll expand restore_openhab_config() to work with files in both backupdirs, the OH2 and OH3 ones. I was thinking to also automate that step but I'd be happier to have a _working_ openHABian codepiece first.

mstormi:
I'm really having problems with integration testing this because on OH2<->OH3 branch changes, new to-be-tested code is replaced from the "other" old repo.
Best I can think of ATM is to merge #1206 first and then this, then do a full test and identify what's left to work on.
@ecdye Wdyt?

ecdye:
Should be fine, the openHAB3 branch is not being used in any production environments and it is to be expected if things break over there.

mstormi:
ok merged to openHAB3.
OH3 needs Java 11 and complains if that's not installed. We should do a forced upgrade, what would be the proper code snippet for that?

ecdye:
> ok merged to openHAB3.
> OH3 needs Java 11 and complains if that's not installed. We should do a forced upgrade, what would be the proper code snippet for that?

Probably the best way to do that would be to display a notice on startup and ask if the user approves, then if so we just call `java_install_or_update "Zulu11-32"` and update `/etc/openhabian.conf`

mstormi:
This is d*mned complex to work with,always have to patch both branches, prepMigration and openHAB3 but cannot use the same codebase.

mstormi:
> Probably the best way to do that would be to display a notice on startup and ask if the user approves, then if so we just call `java_install_or_update "Zulu11-32"` and update `/etc/openhabian.conf`

This is also on interactive install and a user might have Java 11 running already (Zulu or even Adopt).
What's the most easy check for a compatible (11) Java ? Will java_install_or_update "Zulu11-32" overwrite that ?

mstormi:
> java_install_or_update "Zulu11-32"

I inserted that in my latest code. Now it *downloads* that. Shouldn't it be in the cache ? Or is that just Java 8 ?

mstormi:
I believe it's working now but I still cannot do an integration test of a downgrade, would need to merge to master first.
So at this point I think we need to merge and see what's left to not work.

Could you please review now ?

mstormi:
Hmm, a fresh install from image downloads a wrong branch because of
.git/config:    url = https://github.com/mstormi/openhabian
... and I cannot find the reason for that file to exist.

mstormi:
any feedback, @ecdye @holgerfriedrich @BClark09 anyone ?

ecdye:
> > java_install_or_update "Zulu11-32"
> 
> I inserted that in my latest code. Now it _downloads_ that. Shouldn't it be in the cache ? Or is that just Java 8 ?

Just Java 8 unless it was on 64 bit. We had kept it until OH3 in case of compatibility issues as it was stable for certain on Java 8.

mstormi:
prefetch Zulu11-32 please. In case of incompatibilities a 2.5 user can go back and install Zulu8. An OH3 user cannot anyway.

ecdye:
Seems to be working, needs some polish for cross branch changes but overall seems to be working!

mstormi:
> I had to make a few changes to the migration code

which changes, [these ?](https://github.com/openhab/openhabian/commit/291c87d89213c2eedd8eb2c9986f404a84aaf73e#diff-2b13d353215584eb937ab19a145d01b10448f3abe4119e520f2f23be8aae5bb7) 

Updated my PR, please cross-check if that's fine with you now.

AFAIK there's a means for a reviewer in Github to suggest changes to PR that the author can simply accept. A way safer method than a message like this one. Must admit though I have not figured out myself how that works but maybe you know or find out.

docs/openhabian.md
```diff
@@ -290,6 +224,97 @@ This is not done by default because it requires a number of user inputs, but you
...
+-   Flash the system image to your micro SD card as described, do not remove the SD card yet
+-   Access the first SD card partition. It's a vfat/FAT-32 (Windows) filesystem so just use the file explorer of your client PC.
+-   Open the file `openhabian.conf` in a text editor
+-   Uncomment and complete the lines to contain the parameters you want to set
+-   Save, unmount/eject, remove and insert into the RPi and boot it
+-   Continue with the instructions for your hardware
+
+Mind the comments of a configuration parameter. Browse the next documentation section for further explanations.
+
+#### administration user
```
ecdye, line 254: Capitalize administration

docs/openhabian.md
```diff
@@ -290,6 +224,97 @@ This is not done by default because it requires a number of user inputs, but you
 
 Delayed rules load will also be enabled by default. This is renaming the rules files, then after 2 minutes it renames them back. You can toggle to use this feature in menu option 44.
 
+## Setup notes
+
+### on openHAB3
```
ecdye, line 229: Capitalize on

functions/backup.bash
```diff
@@ -44,15 +42,15 @@ restore_openhab_config() {
   fi
 
   local backupList
-  local backupPath="/var/lib/openhab2/backups"
+  local backupPath="${OPENHAB_BACKUPS:-/var/lib/openhab2/backups}"
```
ecdye, line 45: Where does this come from, meaning `$OPENHAB_BACKUPS`

functions/menu.bash
```diff
@@ -25,12 +25,14 @@ Note backups are NOT active per default so remember to set them up right at the
...
 
 show_main_menu() {
-  choice=$(whiptail --title "Welcome to the openHABian Configuration Tool $(get_git_revision)" --menu "Setup Options" 18 116 11 --cancel-button Exit --ok-button Execute \
+  choice=$(whiptail --title "Welcome to the openHABian Configuration Tool $(get_git_revision)" --menu "Setup Options" 20 116 13 --cancel-button Exit --ok-button Execute \
   "00 | About openHABian"        "Information about the openHABian project and this tool" \
   "" "" \
   "01 | Select Branch"           "Select the openHABian config tool version (\"branch\") to run" \
   "02 | Upgrade System"          "Upgrade all installed software packages (incl. openHAB) to their latest version" \
   "03 | openHAB Stable"          "Install or upgrade to the latest stable release of openHAB 2" \
+  "04 | Upgrade to openHAB 3"    "Upgrade to openHAB 3 testing milestone (BETA !)" \
```
ecdye, line 34: Should we force switch to OH3 Branch when this option is selected?

functions/openhab.bash
```diff
@@ -182,13 +189,29 @@ multitail_openhab_scheme() {
   if cond_redirect sed -i -e 's|^# misc.*$|# openHAB logs\\ninclude:/etc/multitail.openhab.conf\\n#\\n# misc|g' /etc/multitail.conf; then echo "OK"; else echo "FAILED (include)"; return 1; fi
 }
 
+## Function to check if openHAB 2 is installed on the current system. Returns
+## 0 / true if openHAB is installed and 1 / false if not.
+##
+##    openhab2_is_installed()
+##
+openhab2_is_installed() {
+  if [[ $(dpkg -s 'openhab2' | grep Status | cut -d' ' -f2) == "install" ]]; then return 0; else return 1; fi
```
ecdye, line 198: use the `dpkg -s 'openhab2' &> /dev/null` style to avoid unwanted output on the user side.

includes/adminkey.pub
```diff
@@ -0,0 +1 @@
+ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAACAQDeEZqjLjbZmqFrhYL1inJM/U92m1/rZ+Lq/ZJvsEITk0I/7GBJqZ5Ze4Oyh/XKtbCC3l9QTwfhOyd/jlcjIaKaRPznY0Prml11lMY6OJHLPaBvKISaRoZb/J9xk0fwRAuXoVQNKMCE5DXCIBYNWEILex6Z2MzESRwbsEciDxdIuBSsJA9IPfCrFZhrgm/oBCtvEKflQCwxMcBI3kdfPUInFpqpafvvNJl0uIexsRMrKEyUMJ7+WDVXc/UQm4P/dodiLPXZkppwrPg5OzqIZWW5TZglCtlo5hMOMD4fTyKplTwWXliKtrGrb54lELCHJpSGDhx/y9aeX+mLXszXeC/QFWPg3DWaFpbaix6NGTR6cDDKpoUKau5hiu03AYz9EJgKDh3FZ/VgSZ5r4FLrU5zErdbyuipc1hcPzc4iEz5oKPxrPK0wvW7zXCibFsYruTxQkFWvF16nLDhbCQ97FjGFVRMP4e5v3V3W1hm4SOqXVfbDpukr6FXYWoeAtJZZpYNKlIkeeOdqFokisZLCpJPjo949n6pgXbyuaDtaArvgcvoIFtKKwfDOrY078vyyX4WsE4kDM3JSGHdkcNO52Ofe84rvFLRuIoRlZEuIfvhTRc6EMgQ89MDZJI4cMb6AzzS/okvKTSZoyualIZW+2fET1fRwSiH7XwLPgg3LpZkybQ== Markus Storm
```
ecdye, line 1: This should not be a change included in the master branch, right? Unless you want your key publicly available.

ecdye (changes requested):
Looks pretty good, just a few changes needed. When you do merge please be sure that the code is consistent throughout all branches.

ecdye:
> AFAIK there's a means for a reviewer in Github to suggest changes to a PR that the author can simply accept. A way safer method than a message like this one. Must admit though I have not figured out myself how that works but maybe you know or find out.

Hmm, I actually did not know that. I know that a reviewer can just make changes if the PR author selects the box that allows it but not a nice way like that. I'll look into it because that would be very convenient.

functions/backup.bash
mstormi: /etc/default/openhab[2]
but only if manually specified there which it isn't by default.

functions/menu.bash
mstormi: I was wondering, too.
I migrate when the user selects to switch to openHAB3 branch but don't when he switches from there elsewhere (back down), that's how the upgrade was meant to be triggered. But we had no downgrade path (as we cannot be sure if a specific branch is OH 2 or 3) so I introduced options 04 and 05
On 2nd thought yes we must switch branch or else future actions triggered from the tool will fail.

functions/openhab.bash
mstormi: ok

includes/adminkey.pub
mstormi: right, this was for testing only, will remove

mstormi:
Synced changes with openHAB3 again, included your changes and added switching branches on both up AND downgrades as the tool needs to work on the changed branch (so it needs to be replaced/restarted via branch switch).
Untested though. Merge like this ?

ecdye:
Is there a way to hide the menu options a little bit more so that we don't advertise potentially broken functionality?

mstormi:
> Is there a way to hide the menu options a little bit more so that we don't blatantly advertise potentially broken functionality?

I moved them to the 40 menu
Also rebased openHAB3 on master

Please have a final look before I merge

mstormi:
Noticed that Cancel on install and the check for an already installed OH didn't work so reordered some.
Seems to work now, hope I didn't overlook and break any of the other use cases.
Updated NEWS, too.
Will merge now and put up a small announcement.

PR #1222, opened 2020-11-02
mount /srv/* with rw forced

### LINKED ISSUE #1221: [srv_bind_mounts] srv_bin_mount are read-only on debian 10.6
## Issue information:
I am using Debian 10.6 in a virtual machine. openhabian ist a template to setup the system. So in facht my system is not an openhabian system, but it is based on openhabian.

I found that the se´rv_mount_binds under /srv/openhab-... are read only by default. I do not know if that is an intention or if that is an error. If the folders are rw than it is possible to use /srv/... as a central point to access all files from open hab und use them in smb-con instead of the "real" folders.

## Debug information:
Just try to create a new file under /srv/openhab-logs an what happen

## Possible solution:
Modify the mount-template from the include folder. Add a ",rw" in line Options=bind so that the new line is:

`Options=bind,rw`

This should fix the problem

mstormi:
Fixes #1221 

### DIFF
--- a/functions/system.bash
+++ b/functions/system.bash
@@ -235,7 +235,7 @@ create_mount() {
   source="$1"
 
   echo -n "$(timestamp) [openHABian] Creating mount $source in '/srv/openhab2-${destination}'... "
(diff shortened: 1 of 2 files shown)


ecdye (approved):
LGTM

PR #1259, opened 2020-11-19
take 1G off rootfs in image during resize boot run

mstormi:


mstormi:
Results in boot loop, could not figure out why so far.

Issue #1254, opened 2020-11-18
Do not have image fully expand on first boot


mstormi:
* take 1 GB off SD size on resize on first boot (argument to resize2fs line ~166 in build.bash)
  that way we're on the safe side when we want to dd-copy an SD later on to a new SD card that eventually is smaller

* cross-check/adjust size checker in mirror_SD to check partition #2 rather than the whole SD size (that will remain unchanged!)

see function set_autoexpand() in pishrink.sh (see #1253)

There must exist a similar script in Raspbian lite as on very first boot, it's expanding the root fs and reboots.
This scriptlet we should modify (take 1GB off the partition end).

preliminary code solution in build.bash **but this seems to result in a booting loop**
```
   echo_process "Cutting 1G sectors (0.5GB) off rootfs to make copies fit into smaller SD cards... "
   # shellcheck disable=SC2016
   sed -i '16 i \    size=$(( $(blockdev --getsize64 $ROOT_DEV) / 512 - 1024 * 1024 ))' "${buildFolder}/root/etc/init.d/resize2fs_once"
   # shellcheck disable=SC2016
   sed -i 's|resize2fs $ROOT_DEV|resize2fs $ROOT_DEV ${size}s|g' "${buildFolder}/root/etc/init.d/resize2fs_once"
   sed -i 's|bin/sh|bin/bash|g' "${buildFolder}/root/etc/init.d/resize2fs_once"
```

mstormi:
@holgerfriedrich are you alive ? could you take on this ?

mstormi:
gave up on this idea for now

PR #1261, opened 2020-11-20
take 0.5GB off root fs 

mstormi:
so when we need to copy the SD card it is more likely to fit into the destination even if that one has a number of bytes less.

Fixes: #1254 

Drafting because preliminary testing results in a reboot loop, reason t.b.d.


mstormi:
I spotted  some message saying "...LSB .... filling root partition" during boot.
That might be part of standard Raspbian and may interfere with the mechanism in this PR.
I could not find out so far.

mstormi:
Closing until there's news on those messages saying "Starting LSB ... Resize root to fill partition" during boot w.r.t. automatic resizing. If so, create a new PR against main.

Issue #1282, opened 2020-12-02
Migrate off Travis ?


mstormi:
Travis changed pricing: https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing
This is affecting our builds so we would need to move over to Github or some alternative.
Let's check if we can move to the (still free) 5 concurrent job plan first.

PR #1283, opened 2020-12-02
Migrate from Travis-CI to GitHub Actions

ecdye:
Travis-CI recently changed their policy in a manner that no longer
allows us to use their service in a way that will meet our needs, as
such this migrates our testing that previously took place on Travis-CI
to GitHub Actions.

Fixes: #1282


ecdye:
That was surprisingly less of a pain then I thought it would be. @mstormi Take a look at the tests and let me know what you think. They still need some polish and I had to make a few changes in order for them to run properly, but overall it worked pretty well.

Also any thoughts about the runtime? I had to use the cross build feature for the ARM tests as GitHub does not currently offer public ARM machines only private.

build-image/first-boot.bash
```diff
@@ -111,7 +111,7 @@ elif grep -qs "openHABian" /etc/wpa_supplicant/wpa_supplicant.conf && ! grep -qs
   else
     echo "OK"
   fi
-  if tryUntil "ping -c1 8.8.8.8 &> /dev/null || curl --silent --head http://www.openhab.org/docs |& grep -qs 'HTTP/1.1 200 OK'" 5 1; then
+  if tryUntil "curl --silent --head http://www.example.com/ |& grep -qs 'HTTP/1.1 200 OK'" 5 1; then
```
mstormi, line 114: You must not change this.
Actually with the early networking stages there can be a case where *any* URL with nothing but a hostname resolves to the *same* IP address and the curl command actually succeeds.
That's why I added a string to the URL used in this test

build-image/first-boot.bash
```diff
@@ -157,7 +157,7 @@ else
 fi
 
 echo -n "$(timestamp) [openHABian] Ensuring network connectivity... "
-if tryUntil "ping -c1 8.8.8.8 &> /dev/null || curl --silent --head http://www.openhab.org/docs |& grep -qs 'HTTP/1.1 200 OK'" 86400 1; then
+if tryUntil "curl --silent --head http://www.example.com/ |& grep -qs 'HTTP/1.1 200 OK'" 86400 1; then
```
mstormi, line 160: same here

docker-tests/entry.sh
```diff
@@ -30,35 +59,56 @@ function mount_dev()
 	mount -t debugfs nodev /sys/kernel/debug
 }
 
-function start_udev()
+function init_systemd()
```
mstormi, line 62: where did you get this from ?
Impossible to validate to me ...

functions/packages.bash
```diff
@@ -258,11 +258,14 @@ mqtt_setup() {
   echo -n "$(timestamp) [openHABian] Beginning the MQTT broker Eclipse Mosquitto installation... "
   if [[ -n $INTERACTIVE ]]; then
     if (whiptail --title "MQTT installation?" --yes-button "Continue" --no-button "Cancel" --yesno "$introText" 14 80); then echo "OK"; else echo "CANCELED"; return 0; fi
+  else
```
mstormi, line 261: Is this build related ?
If no please separate PRs

mstormi (changes requested):
cannot validate upfront if it works but I believe you have had a number of successful runs

so apart from the 3,4 little things LGTM

build-image/first-boot.bash
ecdye: Yeah, I was having issues with this during my initial testing and forgot to change it back, will do.

docker-tests/entry.sh
ecdye: This is the original file from balena's repo that I found.

functions/packages.bash
ecdye: Yes, actually. For some stupid reason without that check it was failing every time. Beats me.

build-image/first-boot.bash
mstormi: Make sure you don't get victim to that apparently working inet link like I got first time.

mstormi:
ready to merge? Tests are still failing....

ecdye:
> ready to merge? Tests are still failing....

Not quite, I need some time to look at them. I just rebased and pushed, but I haven't had time to see what is wrong.

PR #1300, opened 2020-12-14
set apt options to force installing maintainer version of OH3 config files

mstormi:


mstormi:
merge if good

ecdye (changes requested):
The change needs to be applied to both interactive and uninteractive installations, as the output will be hidden because of our `debugmode=off` setting.

mstormi:
wdy mean - where ? to line 95 ? It shouldn't be called unattended, should it ? If so it's a new install. Or what am I missing ?

ecdye:
I mean you only changed the offline install line. Make sure you change both.

mstormi:
ok now?

ecdye:
Yes looks good.

ecdye (approved):

mstormi:
good. let's get 1283,1301,1302 in then forward to stable so have a look at those PRs please

I'll locally clone master into stable then force-upload.
Or have you been able to come up with a nicer solution meanwhile ?

ecdye:
Not really, that is probably best

PR #1322, opened 2020-12-22
Fix download branch for installation instruction

Confectrician:
See discussion @ openhab/openhabian#1318

Fix here will be pulled into the docs repo with one of the next builds.
Please also note that the docs are always fetched on the master branch.
Pull Request target is done on purpose here.

Signed-off-by: Jerome Luckenbach <github@luckenba.ch>

### DIFF
--- a/docs/openhabian.md
+++ b/docs/openhabian.md
@@ -120,7 +120,7 @@ apt-get update
 apt-get install git
 
 # download, link and create config file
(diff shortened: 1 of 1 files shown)


mstormi:
Thanks. But I already change that (and more) so please take care to get this file revision uploaded into the official docs quickly.

PR #1342, opened 2020-12-28
Move network manager install to right before hotspot setup

mstormi:
to not deviate from standard OS setup more than necessary for the hotspot to work.

Yes I have tested using the hotspottest image:
1) w/ Ethernet
2) w/o Ethernet and WiFi SSID empty
3) w/o Ethernet and WiFi SSID fine
4) w/o Ethernet and WiFi SSID bad

The hotspot was only fired up in cases 2) and 4)


ecdye (approved):
LGTM

Issue #1341, opened 2020-12-28
Java 8 "forced downgrade"


mstormi:
There's some reports of Java being "downgraded" on OH3 installation

I just built a new image (hotspottest branch which is a copy of the current master branch) at it showed
`java_opt=Zulu8-32` in `openhabian.conf`

??? (not analyzed yet)

t09:
Hi,
choosing **02 | Upgrade System** reinstalls Java 8

Thanks for your work!

mstormi:
I think the issue is when java_opt does not get set in openhabian.conf
(that is only the case on oldish installation that did not have this in their config by the time the system was installed)

So the default taken from here applies:
java_install_or_update "${java_opt:-Zulu8-32}"

I'll fix in 2 ways:
First I'll set the default Java to Zulu11 also for OH2 (i.e. only master and stable branches)
Second I'll update openhabian.conf to contain all params as discussed in #1308

mstormi:
Just did an `apt upgrade` (menu 02) on my dev box and it resulted in openHABian trying to install Java 8.
It tried Java11-64 first but since I'm on a 32bit kernel it failed and the fallback apparently was Java 8.

@ecdye I think there's two issues here:
1) why did it try to go for the 64 bit version in the first place?
2) the last resort (hardcoded ?) is Java 8 and should be changed to Java11-32.

@ecdye as you've built that code could you please check ?
Note system_upgrade() was called when I selected menu 02 so problems could hide in that routine

On #1:
```
[12:52:14] root@devpi:/home/openhabian# grep java /etc/openhabian.conf
java_opt=Zulu11-64
```

I never entered that, it's possibly a result of my config-redo patch.
After I changed it to 11-32, upgrade (02) finished without trying to downgrade Java.




```
+ [[ 02 | Upgrade System == '' ]]
+ [[ 02 | Upgrade System == \0\0* ]]
+ [[ 02 | Upgrade System == \0\1* ]]
+ [[ 02 | Upgrade System == \0\2* ]]
+ wait_for_apt_to_finish_update
++ timestamp
++ date +%F_%T_%Z
+ echo -n '2020-12-31_12:42:04_CET [openHABian] Updating Linux package information... '
2020-12-31_12:42:04_CET [openHABian] Updating Linux package information... + [[ -z 19515 ]]
+ tail --pid=19515 -f /dev/null
+ echo OK
OK
+ system_upgrade
++ timestamp
++ date +%F_%T_%Z
[...]

Issue #1327, opened 2020-12-25
Running "upgrade system" from openhabian-config reinstalls java 8


SRGDamia1:
On openhabian on a pi 3b, choosing the "upgrade system" from the options on openhabian-config causes java zulu 8 to be installed, even after upgrading to oh3/zulu11.

Issue #1308, opened 2020-12-19
3.0 release


mstormi:
We need to switch to install OH3.0 by default.
Please help by double-checking so I don't forget anything of importance.

Plan is to
1) add last minute changes (release repo ?) to master
2) rebase openHAB3 branch on master
3) build a new image from the openHAB3 branch (to have clonebranch=openHAB3).
There was a report it works to install OH3 right away like this but we need to test this ourselves upfront
4) forward master to stable

Is that the best order ? Or should we dare to replace stable with openHAB3 ?

Ethan will you be available on Monday for OH3 release if there's a need for last minute changes?

ecdye:
I should be in the afternoon, but I'll be busy during the morning.

mstormi:
The final 3.0 package is now on bintray. It should work to install from either master or openHAB3 branches.
Please try yourself when you have a chance to.

I triggered another image build, will that pull in the latest "openhab" package? If not can you please build an image to include that.
I'll also forward master to stable now.

mstormi:
OH3 got announced. I've built another image (see actions) and will be putting it out as openHABian v1.6.2 release.
Please test when you have a chance to.

mstormi:
See https://community.openhab.org/t/openhabian-update-help-needed/110780/11

Not sure why it didn't upgrade Java. Possibly the existing java setting in openhabian.conf ?

mstormi:
Ethan could you please at times replace the images ? They still show the menu with "openHAB 2"

ecdye:
What do you mean?

mstormi:
Sorry. I mean the images in docs/openhabian.md that ultimately show up on www.openhab.org/docs

mstormi:
Regarding TODOs left on OH3:
#1273 imho is important as it seems to confuse quite some people so could you please tackle that?

I've fixed the duplicate repo warning in https://github.com/openhab/openhabian/commit/692eee77807d80859b5d33fe64ea3e7881b97c6d

What's left and most important is that apparently many people to upgrade to (or 1st time install?) OH3 do **not** end up in an active openHAB3 branch, then they complain to get affected by the differences in setup OH2/3, directory names for the most part. I'm pretty sure it's because they're in the wrong branch. Unfortunately there has not been a single useful report to date telling how they got there.

ecdye:
Perhaps we should make openHAB3 main (i can do it) and then set if as default and leave master for now as a legacy branch for any critical updates that may be needed.

[...]

PR #1347, opened 2020-12-29
update openhabian.conf on every start to have latest set of parameters

mstormi:
Related: #1308 

Fixes: #1327
Fixes: #1341

### DIFF
--- a/functions/openhabian.bash
+++ b/functions/openhabian.bash
@@ -82,13 +82,12 @@ update_openhabian_conf() {
   cp /dev/null $config
 
   while read -r line; do
-    #if line is a comment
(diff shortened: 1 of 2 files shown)


mstormi:
Ha this was fun to figure out. Even learnt some more bashery along the way.

Not 100% sure if we should call on every start but that's the safest way I can think of ATM.

ecdye (approved):
Not sure I completely understand that crazy bashery but LGTM.

mstormi:
> Not sure I completely understand that crazy bashery but LGTM.

That's why I put in the comments... ${!var} == var i.e. the _name_ of the variable. I think that one is key to understand.

PR #1349, opened 2020-12-29
small adjustments for zram handling

mues-lee:
- changed uncommented preselection to disable
- hide rPi4 8GB warning, when zram installation is disable in conf file

### DIFF
--- a/build-image/openhabian.conf
+++ b/build-image/openhabian.conf
@@ -65,7 +65,7 @@ debugmode=off
 java_opt=Zulu8-64
 
 # install ZRAM per default, set to "disable" to skip installation
-#zraminstall=enable
+#zraminstall=disable
 
 # start comitup hotspot if internet is not reachable
 hotspot=enable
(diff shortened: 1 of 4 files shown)


mstormi:
I don't see the benefit of your changes. If someone has a RPi4/8G I want him so see the warning no matter if he has set zraminstall.
Note that the default is to install ZRAM so the intended preset for zraminstall is enable.

mues-lee:
what are the benefits of showing someone a warning message even if they already actively, by changing the config file, decided to not use zram?

yes the default is to install zram, but since that line has a # , the users need to remove it and also change it into disable
so why not just change it into disable, so the users only needs to remove the hashtag if they want to disable zram install?

mstormi:
> what are the benefits of showing someone a warning message even if they already actively, by changing the config file, decided to not use zram?

they might have done it long ago and forgot they did ?

So no that does not convince me, I prefer to keep the status quo.

Thanks a lot for your contribution, however.

mues-lee:
ok, no problem
not all my ideas needs to be approved

Issue #1366, opened 2021-01-01
Installation of Node Red on OH3 fails


Tobster77:
Dear all,

I decided to test Node Red on my recent OpenHABian installation, v 1.6.2 using OH3.0 with all updates as of today, i.e. first release. I run a RPi 3B+.

While the installation using `openhabian-config` seems to run well, in the last steps I receive an error
`Adding an openHAB dashboard tile for 'nodered'... FAILED (openHAB or config folder missing)`

I assume that the installer complaints about the missing OH2 dashboard that provided all UIs in the past, but I have no deeper insight. The editor on port 1880 is running well!

Thanks for caring!

ecdye:
Thanks for the report. That is old code that needs to be removed.

mstormi:
@ecdye see https://github.com/openhab/openhabian/pull/1320#issuecomment-750019244
Please figure how easy or not the format change before you remove anything, maybe it's not so much work to keep it

Tobster77:
This is not resolved. I called openhabian-config today, re-installing Node-Red. Still the same error...

mstormi:
>  I was thinking that openhabian-config is updating itself on invoking, right?

It'll ask you for confirmation so you should know if you confirmed it.
Make sure to use the main branch for latest fixes.

Tobster77:
It asked. I confirmed. But I am not sure it did as the error remains.
The version I am on is openHABian Configuration Tool [master] patchday-20210119-1114(e0c0204)

I am on "master" branch. I applied the current system updates which pushed openHAB to 3.0.1, if this plays a role.

mstormi:
> I am on "master" branch. I applied the current system updates which pushed openHAB to 3.0.1, if this plays a role.

I said to use main, didn't I.

PR #1373, opened 2021-01-04
Update tiles for openHAB3

ecdye:
Fixes: #1366 


Tobster77:
Can I somehow get the Node Red tile that was not installed during installation one week ago? Will openHABian check for "lost" optional components?

ecdye:
If you use the function again it will add the tile at the end of the setup

Tobster77:
Thanks. Will this erase my flows?

Issue #1273, opened 2020-11-29
replace motd to match OH2 or OH3


mstormi:
on login to the box

make sure to change motd when migrating OH2<->OH3

ecdye:
I think I already fixed this? Did you test?

mstormi:
I just tried on my test box with current master and no it's still the 2.5 banner.
I think you said it would get deployed on a fresh install, that I haven't tried, but I would expect this to also change when you
up/downgrade OH through the menu.

(just to be sure we don't misunderstand: this is on ssh login - not in Karaf console).

ecdye:
Yes, perhaps I never merged it, I'll look into it.

ecdye:
I see what happened, it is a merge conflict and was overwritten on the OH3 branch. I'm fixing it now.

ecdye:
Actually, I was wrong, the code is in place the problem is that the branch update occurs after the migration and it must because it forces a restart of the tool to load the changes. I will look into a workaround.

mstormi:
Ran a downgrade to OH2 and ended up with

+ echo 'The tool will now restart to load the updates... OK'
The tool will now restart to load the updates... OK
+ exec '/opt/openhabian/openhabian-setup.sh migration'
/opt/openhabian/functions/openhabian.bash: line 183: /opt/openhabian/openhabian-setup.sh migration: No such file or directory
[21:08:21] root@openHABianDevice:/home/openhabian# ls -l /opt/openhabian/openhabian-setup.sh
-rwxr-xr-x 1 root root 3258 Dec  1 21:08 /opt/openhabian/openhabian-setup.sh
[21:08:45] root@openHABianDevice:/home/openhabian#

mstormi:
it was bad quotes, I fixed it

mstormi:
I got this on a fresh install (which has openhab2 installed by default)


sed: /var/lib/openhab/etc/version.properties kann nicht gelesen werden: Datei od                                                  er Verzeichnis nicht gefunden
sed: /var/lib/openhab/etc/version.properties kann nicht gelesen werden: Datei od                                                  er Verzeichnis nicht gefunden

                          _   _     _     ____   _
  ___   ___   ___   ___  | | | |   / \   | __ ) (_)  ____   ___
 / _ \ / _ \ / _ \ / _ \ | |_| |  / _ \  |  _ \ | | / _  \ / _ \
[...]

PR #1392, opened 2021-01-07
Dynamically generate the login screen

mstormi:
Fixes: #1273

### DIFF
--- a/includes/bash_profile
+++ b/includes/bash_profile
@@ -1,25 +1,22 @@
 # ~/.bash_profile: executed by bash(1) for interactive shells.
 
+source /opt/openhabian/functions/openhab.bash
+
 if [ -f "/opt/openHABian-install-failed" ]; then
-  echo ""
-  echo -e -n "\e[31;01mAttention! \e[39;49;00m"
+  echo -e -n "\\n\e[31;01mAttention! \e[39;49;00m"
   echo "The openHABian setup process seems to have failed on your system."
   echo "Sorry, this shouldn't happen! Please restart the installation process. Chances"
-  echo "are high setup will succeed on the second try."
-  echo ""
+  echo -e "are high setup will succeed on the second try.\\n"
   echo "In order to find the cause of the problem, have a look at the installation log:"
-  echo -e "   \e[90;01msudo cat /boot/first-boot.log\e[39;49;00m"
-  echo ""
+  echo -e "   \e[90;01msudo cat /boot/first-boot.log\e[39;49;00m\\n"
   echo "If the error persists, please read the debug guide at:"
-  echo -e "\e[94;04mhttps://github.com/openhab/openhabian/blob/master/docs/openhabian-DEBUG.md\e[39;49;00m"
-  echo ""
+  echo -e "\e[94;04mhttps://github.com/openhab/openhabian/blob/master/docs/openhabian-DEBUG.md\e[39;49;00m\\n"
   return 0
 elif [ -f "/opt/openHABian-install-inprogress" ]; then
-  echo ""
-  echo -e -n "\e[36;01mAttention! \e[39;49;00m"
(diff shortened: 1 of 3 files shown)


includes/bash_profile
```diff
@@ -1,25 +1,22 @@
...
+
 if [ -f "/opt/openHABian-install-failed" ]; then
-  echo ""
-  echo -e -n "\e[31;01mAttention! \e[39;49;00m"
+  echo -e -n "\\n\e[31;01mAttention! \e[39;49;00m"
   echo "The openHABian setup process seems to have failed on your system."
   echo "Sorry, this shouldn't happen! Please restart the installation process. Chances"
-  echo "are high setup will succeed on the second try."
-  echo ""
+  echo -e "are high setup will succeed on the second try.\\n"
```
mstormi, line 9: Didn't know about this text so far so not strictly related to your changes but we should fix this nonetheless.
Reword to explain *how* exactly to restart install. A beginner will not know and AFAIK that isn't even explained anywhere.
I think it should go into the debug guide.
Even I myself am not sure what to write, my current best best would be
`rm -f /opt/openHABian*; reboot`

includes/bash_profile
```diff
@@ -1,25 +1,22 @@
...
   echo "If the error persists, please read the debug guide at:"
-  echo -e "\e[94;04mhttps://github.com/openhab/openhabian/blob/master/docs/openhabian-DEBUG.md\e[39;49;00m"
-  echo ""
+  echo -e "\e[94;04mhttps://github.com/openhab/openhabian/blob/master/docs/openhabian-DEBUG.md\e[39;49;00m\\n"
   return 0
 elif [ -f "/opt/openHABian-install-inprogress" ]; then
-  echo ""
-  echo -e -n "\e[36;01mAttention! \e[39;49;00m"
+  echo -e -n "\\n\e[36;01mAttention! \e[39;49;00m"
   echo "The openHABian setup process is not finished yet."
```
mstormi, line 17: _has_ not ? You're the native speaker.

mstormi (commented):
LGTM
however things I'm unaware of are
* today there already is something done to replace(?) .bash_profile on OH install/up-/downgrade. Would you need to remove anything of that ?
* how and when is .bash_profile created ?
  can or should we do anything about it w.r.t. existing installations ? put into openhabian-config startup maybe ?
  Not sure but be aware this is sensible and easily confuses people whenerver there's a change.

includes/bash_profile
ecdye: I didn't look at the text, I just simplified the output code. I'll fix.

> LGTM
> however things I'm unaware of are
> 
> * today there already is something done to replace(?) .bash_profile on OH install/up-/downgrade. Would you need to remove anything of that ?

As of right now, I am inclined to leave it so that hopefully it gets replaced again with the new fixed version and then remove it in a couple of weeks.

> * how and when is .bash_profile created ?

On original installation, and for now on branch upgrade/switch.

>   can or should we do anything about it w.r.t. existing installations ? put a check into openhabian-config startup maybe ?

see above.

>   Not sure but be aware this is sensible and easily confuses people whenever there's a change.
>   Ideally we would not even _overwrite_ .bash_profile but allow the user to edit it.

After this whole mess is overwith, that should no longer be an issue as it will no longer be ovewritten because once I remove the code to replace it every time it will only be modified on original installation.


Any reason you are creating a duplicate PR?

mstormi: ok. The PR is your code just not against master as I wanted to test it upfront (but got interrupted)

ecdye: I can just push my code to a new branch if you would like. Do you ant it based off of OH3 or OH2

mstormi: Oh dear I though I was reviewing your PR only noticed now that it was actually my _copy_ of your PR I was reviewing.
So please take this as the input, fix the text according to my 2 comments above here then wen can merge it.

PR #1395, opened 2021-01-08
Use proper format to warn on not using a 8GB RPi

Hilbrand:
This will display the message in a nicely formatted box.

Signed-off-by: Hilbrand Bouwkamp <hilbrand@h72.nl>


ecdye (approved):
LGTM

Issue #1409, opened 2021-01-13
dhcpcd failing on fresh install


mstormi:
credentials for wlan0 were provided in `openhabian.conf`

change wifi.bash to write into /etc/network/interfaces.d/wlan0-some-filename.cfg instead of /etc/network/interfaces ??
```
[17:53:16] root@mysite:/home/openhabian# systemctl start dhcpcd.service
Job for dhcpcd.service failed because the control process exited with error code.
See "systemctl status dhcpcd.service" and "journalctl -xe" for details.
[17:53:18] root@mysite:/home/openhabian# journalctl -xu dhcpcd.service
-- Logs begin at Tue 2021-01-12 17:50:52 CET, end at Wed 2021-01-13 17:48:48 CET. --
Jan 13 17:48:48 mysite systemd[1]: Starting dhcpcd on all interfaces...
-- Subject: A start job for unit dhcpcd.service has begun execution
-- Defined-By: systemd
-- Support: https://www.debian.org/support
--
-- A start job for unit dhcpcd.service has begun execution.
--
-- The job identifier is 31075.
Jan 13 17:48:48 mysite dhcpcd[32206]: Not running dhcpcd because /etc/network/interfaces
Jan 13 17:48:48 mysite dhcpcd[32206]: defines some interfaces that will use a
Jan 13 17:48:48 mysite dhcpcd[32206]: DHCP client or static address
Jan 13 17:48:48 mysite systemd[1]: dhcpcd.service: Control process exited, code=exited, status=6/NOTCONFIGURED
-- Subject: Unit process exited
-- Defined-By: systemd
-- Support: https://www.debian.org/support
--
-- An ExecStart= process belonging to unit dhcpcd.service has exited.
--
-- The process' exit code is 'exited' and its exit status is 6.
Jan 13 17:48:48 mysite systemd[1]: dhcpcd.service: Failed with result 'exit-code'.
-- Subject: Unit failed
-- Defined-By: systemd
-- Support: https://www.debian.org/support
--
-- The unit dhcpcd.service has entered the 'failed' state with result 'exit-code'.
Jan 13 17:48:48 mysite systemd[1]: Failed to start dhcpcd on all interfaces.
-- Subject: A start job for unit dhcpcd.service has failed
-- Defined-By: systemd
-- Support: https://www.debian.org/support
--
-- A start job for unit dhcpcd.service has finished with a failure.
--
-- The job identifier is 31075 and the job result is failed.
[...]

PR #1414, opened 2021-01-14
Fix dhcpcd

mstormi:
Fixes: #1409 

Signed-off-by: Markus storm <markus.storm@gmx.net>


mstormi:
Well in the end I believe dhcpcd failing was not due to an improper network config so this patch would not fix it.
We can reopen should we need to have another need to change the hotspot code.

PR #1449, opened 2021-02-03
Update frontail-theme.css

openPhiL:
website looks bad if openhab is offline (internet-outtage etc.). I have no idea why having this css available locally is a bad idea :)
I used the github-change-option, so I cannot add the bootstra.min.css file to the repo myself.

### DIFF
--- a/includes/frontail-theme.css
+++ b/includes/frontail-theme.css
@@ -1,4 +1,4 @@
-@import url('//stackpath.bootstrapcdn.com/bootstrap/3.4.1/css/bootstrap.min.css');
+@import url('bootstrap.min.css');
 
 body {
   padding-top: 7em;


ecdye:
Thanks for your contribution, but we cannot incorporate this change.

The problem is that `bootstrap.min.css` is not provided by frontail, and manually downloading it would cause `npm` to complain on upgrades. As a result, this cannot work, that is why we source from the online link in the first place.

openPhiL:
ok, I already thought that there was a reason, wanted to try out that merge thing anyway and.. to be honest, still don't get it 100%. It is a openhab-specific css, so frontail-updates will not mess with it, and it is a openhab-specific installation (openhabian could place the bootstrap.css somewhere once). 
On the other hand, it really doesn't matter for 99,9% of users anyway :D

ecdye:
> ok, I already thought that there was a reason, wanted to try out that merge thing anyway and.. to be honest, still don't get it 100%. It is a openhab-specific css, so frontail-updates will not mess with it, and it is a openhab-specific installation (openhabian could place the bootstrap.css somewhere once).

Its not the openhab specific css that is the issue, rather the bootstrap css that is the issue. For some reason on some machines the bootstrap css will not be imported when stored locally.

PR #1451, opened 2021-02-04
Update a few links which lacked the `docs` path

brianwarner:
Just a few very minor tweaks to the links in the README.

Signed-off-by: Brian Warner <brian@bdwarner.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ We'll be happy to include that in openHABian so you can use your box with openHA
 However, that does not make your box a "supported" one as we don't have it available for our further development and testing.
 So there remains a risk that future openHABian releases will fail to work on your SBC because we changed a thing that broke support for your HW - unintentionally so however inevitable.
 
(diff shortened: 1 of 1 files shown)


ecdye (approved):
LGTM

mstormi:
I ommited that from the README which should essentially be a copy of the respective docs/openhabian.md section.
In that file however links must not start with docs/ because that file is uploaded to the official openHAB docs and leading "docs/" break links as it is currently the case. I'll fix that.

ecdye:
Hmm, why does that break it? Are the maintainers of openHAB-docs aware of this bug?

mstormi:
it is not a bug. It's a relative link relative to the perspective of the location of openhabian.md which is in the same docs/ subdir as itself (obviously) or the other .md files with READM.md being the sole exception. But README is not being linked to while openhabian.md is.
That's what it is no matter if in our repo or the openhabian docs location our docs/ dir is mirrored to.

ecdye:
Ahh, ok I misinterpreted.

Issue #1469, opened 2021-02-25
mirror mirror who is the fairest ...


mstormi:
Following up on complaints on the forum a user cannot boot from a mirror SD he made of a mirror, it drops into emergency mode.
Duh.

Not ultimately clarified yet but educated guessing it's possibly because it cannot mount /dev/sdX3 as /storage.
Note that partition a) is created only if on running setup_mirror_SD, the external card is larger than the internal one and b) unlike partitions 1 & 2, it is not mirrored so does not exist on the mirror's mirror.

Not fully sure yet how to handle that best so writing it down in the first place. @ecdye let me know your opinion.
If someone has equal sized cards, it won't create the storage partition on mirror setup because the flashed image expands /dev/mmcblk0p2 to all available space but I didn't manage to avoid that (this is what #1254 is really about).
But if it's a mirror inserted to boot from it could exist there so we could check for existence of /dev/mmcblk0p3 in setup_mirror_SD() and dd-copy it, too. I believe it won't exist on flashed images but eventually on mirror cards.
And/or should we remove or at least disable the storage in `/etc/systemd/system/storage.mount` ?
[how does disabling work if you didn't boot from it but mounted "/" as "/mnt" or other? option to sysctl? Manually rm the link?]

ecdye:
Honestly, after the first mirror for our sake and sanity I would say anything beyond that should be unsupported as the mirror is already a pretty good backup.

Also, nice title.

mstormi:
Hmm. Thought that in the beginning, too, but what if someone really needs to resort to using the mirror, how would he need to proceed then to re-activate another mirror? Buy a half-the-size SD and amrestore, then switch to boot from there ? Awkward for newbies.
I'd guess he would expect it should work to re-select the SD mirror menu. Simple. But it won't, will it.

PR #1470, opened 2021-02-25
Mirror 3rd partition if it exists

mstormi:
Fixes: #1469 

### DIFF
--- a/functions/backup.bash
+++ b/functions/backup.bash
@@ -365,7 +365,7 @@ mirror_SD() {
 
   if [[ $1 == "raw" ]]; then
     for i in 1 2 3; do
-      sfdisk -d ${src}|grep -q "^${src}p${i}" || continue
+      sfdisk -d ${src} | grep -q "^${src}p${i}" || continue
       srcSize="$(blockdev --getsize64 "$src"p${i})"
       destSize="$(blockdev --getsize64 "$dest"${i})"
       if [[ "$destSize" -lt "$srcSize" ]]; then
--- a/functions/backup.bash
+++ b/functions/backup.bash
@@ -1,5 +1,4 @@
 #!/usr/bin/env bash
-# shellcheck disable=SC2012
 
 ## Create a backup of the current openHAB configuration using openHAB's builtin tool
 ##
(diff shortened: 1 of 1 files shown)


ecdye (approved):
LGTM, DNT.

mstormi:
spend some thought to validate the logic please (it gave me a hard time so wanna be sure I don't miss anything)

On selecting mirroring via menu,
*if*a 3rd partition exists on the source SD then the partition table is completely copied
if not then if there is enough remaining space (4+ GB) we make a /storage partition out of it.

Finally in mirror_SD() partitions 1 & 2 are always copied, 3 also but only if it exists

functions/backup.bash
```diff
@@ -446,7 +446,7 @@ setup_mirror_SD() {
   local dest
   local srcSize
   local destSize
-  local minBackupSize
+  local minStorageSize=4294967296                   # 4 GB
```
ecdye, line 449: I don't mean to give you a headache, but because of the nasty people in charge of marketing there is a difference between GiB and GB:

Screen Shot 2021-02-26 at 10 18 45 AM

Not sure that it matters in this case but worth a thought.

functions/backup.bash
```diff
@@ -507,19 +507,24 @@ setup_mirror_SD() {
...
-    (sfdisk -d /dev/mmcblk0; echo "/dev/mmcblk0p3 : start=${start},size=${destSize}, type=83") | sfdisk --force "$dest"
-    partprobe
-    cond_redirect mke2fs -F -t ext4 "${dest}3"
-    if ! sed -e "s|%DEVICE|${dest}3|g" -e "s|%STORAGE|${storageDir}|g" "${BASEDIR:-/opt/openhabian}"/includes/storage.mount-template > "${serviceTargetDir}/${mountUnit}"; then echo "FAILED (create storage mount)"; fi
-    if ! cond_redirect chmod 644 "${serviceTargetDir}/${mountUnit}"; then echo "FAILED (permissions)"; return 1; fi
-    if ! cond_redirect systemctl -q daemon-reload &> /dev/null; then echo "FAILED (daemon-reload)"; return 1; fi
-    if ! cond_redirect systemctl enable --now "$mountUnit"; then echo "FAILED (enable storage mount)"; return 1; fi
+  mountUnit="$(basename "${storageDir}").mount"
+  systemctl stop "${mountUnit}"
+  if sfdisk -d ${src}$ | grep -q "^${src}p3"; then
```
ecdye, line 512:
```suggestion
  if sfdisk -d ${src} | grep -q "^${src}p3"; then
```

ecdye (changes requested):
Logic looks good, just a couple of things that need some more thought.

functions/backup.bash
mstormi: ok changed

ecdye: Really, did you forget to push?

functions/backup.bash
mstormi: yes it was a copy'n paste typo that went unspotted

functions/backup.bash
mstormi: no not pushed yet
Still looking how to deal with that "suggestion" commit

mstormi: ah ok I can apply it, pull from my box then push but need to set upstream then

ecdye (approved):
LGTM, DNT.

PR #1472, opened 2021-02-26
Exclude some directories from Amanda backup

### LINKED ISSUE #1471: do not backup cache, tmp
We've got a definition problem in Amanda disklist.
/var/lib/openhab is backed up INCLUDING cache and tmp. That we mustn't. They didn't use to be backed up.
I thought it happened yesterday when you removed persistence but /var/lib/openhab was already there so it must have happened somewhere along the way.
So either we define disklist as it used to be (explicitly name those /var/lib/openhab/xxx we want to be backed up)
or we could try to make use of amgtar. Does it have an exclude option/property we could set like @ecdye you recently did?

ecdye:
Fixes #1471


mstormi (changes requested):
so where's the exclude file ?

ecdye:
Darn, I hate it when I forget to add new files in git. Adding it now.

mstormi (approved):

PR #1478, opened 2021-03-07
add Tailscale default key of mstormi

mstormi:


ecdye:
I'm no sure that you want that key to be public right?

mstormi:
Just disregard pls it's a test for private use

PR #1484, opened 2021-03-11
Make hotspotpw parameter configurable as GitHub Secret

mstormi:
Github env name HOTSPOTPW

### DIFF
--- a/.github/workflows/build-action.yml
+++ b/.github/workflows/build-action.yml
@@ -46,9 +46,9 @@ jobs:
        - name: Build openHABian 32bit image
          id: build
          run: |
-           sed -i -e "s|^userpw=.*$|userpw=\"${{secrets.USERPW}}\"|g" ./openhabian.conf.dist
-           #sed -i -e "s|ap_password:.*$|ap_password: ${{secrets.USERPW}}|g" ./includes/comitup.conf
            sudo -E ./ci-setup.bash github .pi-raspios32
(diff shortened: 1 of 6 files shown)


ecdye (approved):
LGTM, DNT

ecdye:
By default, the secrets are empty right now. Do we want to set them to the default `openhabian` password we use elsewhere?

mstormi:
I had set them as Github secrets but they disappear if you do.
Is that a Github bug?

ecdye:
> I had set them as Github secrets but they disappear if you do.

What do you mean? You set them to the default?

> Is that a Github bug?

I'm not sure, I've never played with GitHub Secrets before. Did you look in the documentation?

mstormi:
> What do you mean? You set them to the default?

Yes to "openhabian" (you can under 'Settings' - Secrets)

> I'm not sure, I've never played with GitHub Secrets before. Did you look in the documentation?

Me neither. Did not find anything mentioned.

ecdye:
Hmm, that's odd. Is it blank in the resulting image as well?

mstormi:
apparently so at least with my forked repo this was the case.

ecdye:
Hmm, maybe just a GH bug. Perhaps we should wait to merge until sorted out.

ecdye:
I've looked into this, and I think that being unable to view the current secret is a security feature. Feel free to merge if you would like.

mstormi:
did you test? As I wrote I believe the password was gone when I build an image last time but I'm not sure any more

mstormi:
just tested, it kept my Github-provided password so will merge

PR #1489, opened 2021-03-15
Update extension from "influxdb" to "persistence-influxdb"

EliasGabrielsson:


mstormi (commented):
Not using InfluxDB so don't know what that does or fixes.
Does it mean it didn't work to use InfluxDB as persistence so far ?
Could you give any pointer please ?

EliasGabrielsson:
The script will currently not help the user with installation of the "influxdb addon" which it originally did.
The namespace scheme of addons  in openhab have been changes since I originally scripted this installation.
This patch corrects the script to use the new scheme.

Issue #1487, opened 2021-03-12
add support for waveshare UPS


mstormi:
See https://www.waveshare.com/wiki/UPS_HAT 

Needed:
python3-smbus package
Enablement of I2C bus:
add `dtparam=i2c_arm=on` to `/boot/config.txt`

PR #1490, opened 2021-03-16
Add support for Waveshare Pi UPS (enables I2C bus on RPi)

mstormi:
Fixes: #1487 

### DIFF
--- a/functions/system.bash
+++ b/functions/system.bash
@@ -59,7 +59,7 @@ needed_packages() {
   then echo "OK"; else echo "FAILED"; return 1; fi
 
   if is_pizerow || is_pithree || is_pithreeplus || is_pifour && [[ -z $PREOFFLINE ]]; then
-    if cond_redirect apt-get install --yes python3-smbus; then echo "OK"; else echo "FAILED"; return 1; fi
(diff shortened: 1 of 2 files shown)


ecdye (commented):
Does enabling this by default cause any issues with other parts of the system, say if someone is using the GPIO pins for something else? Also can you explain what exactly this python code does?

mstormi:
> Does enabling this by default cause any issues with other parts of the system, say if someone is using the GPIO pins for something else?

Obviously cannot tell for sure but if you're really into HW hacking, enabling I2C is a popular option definitely more popular than to use GPIO binding so that is rather "the" default and I don't see any issue with that.
The python code creates simple status output

```
openhabian@openhab:~ $ python3 INA219.py
Load Voltage:   8.280 V
Current:       -0.034900 A
Power:          0.288 W
Percent:       95.0%

Load Voltage:   8.284 V
Current:        0.001100 A
Power:          0.082 W
Percent:       95.2%

```

ecdye (approved):
Seems fine to me, DNT. Feel free to merge if you feel confident that it won't break anything

Issue #1503, opened 2021-05-03
Replace bintray repos with artifactory


mstormi:
See #1502, https://community.openhab.org/t/openhabian-frontail-log-doesnt-work-21-leads-to-error/121788 and probably others

Should we replace ALL bintray repos by their artifactory equivalents on openhabian-config startup ?
If we only do per-menu option this'll keep us busy telling people over and over.

Probably we should create a popup informing the user so he knows where to look should something go wrong.

ecdye:
Sounds like a good idea to me, have we finished migrating all of the repos to artifactory yet?

mstormi:
Would we need to backport our fix to the stable branch, too ?

marcelser:
I definitly would opt for backporting. Cause anyone using openhab2 or OpenHAB3 (stable) branch will have this problems all of a sudden and like me at first have no clue how to resolve it.

PR #1504, opened 2021-05-03
Migrate openHAB APT repository to Artifactory

mstormi:
not sure if that's all we need to replace. @BClark09  ?

Fixes: #1503

### DIFF
--- a/functions/openhab.bash
+++ b/functions/openhab.bash
@@ -305,7 +305,9 @@ dashboard_add_tile() {
 ##    fix_openhab_repo
 ##
 fix_openhab_repo() {
-  local warningText="Due to bintray shutdown of services, there is an immediate need to change the openhab stable repository configuration.\\nBintray repo in /etc/apt/sourceslist.d/openhab.list needs to be replaced by Artifactory.\\n\\nDo you want this to be replaced now ?"
+  local warningText="Due to bintray shutdown of services, there is an immediate need to change the openhab stable repository configuration.\\nBintray repo in /etc/apt/sourceslist.d/openhab.list needs to be replaced by Artifactory.\\n\\nDo you want this to be replaced now?"
+
+  if ! grep -qs 'https://dl.bintray.com/openhab/apt-repo2' /etc/apt/sources.list.d/openhab.list; then return; fi
 
   echo -n "$(timestamp) [openHABian] Offering to replace outdated bintray repository with artifactory repo... "
   if ! (whiptail --title "Repository replacement" --yes-button "Replace" --no-button "Skip" --yesno "$warningText" 10 80); then echo "SKIPPED"; return 0; fi
--- a/functions/openhab.bash
+++ b/functions/openhab.bash
@@ -307,8 +307,9 @@ dashboard_add_tile() {
 fix_openhab_repo() {
   local warningText="Due to bintray shutdown of services, there is an immediate need to change the openhab stable repository configuration.\\nBintray repo in /etc/apt/sourceslist.d/openhab.list needs to be replaced by Artifactory.\\n\\nDo you want this to be replaced now ?"
 
-  if ! (whiptail --title "Repository replacement" --yes-button "Continue" --no-button "Cancel" --yesno "$warningText" 10 80); then echo "CANCELED"; return 0; fi
-  rm /etc/apt/sources.list.d/openhab2.list
+  echo -n "$(timestamp) [openHABian] Offering to replace outdated bintray repository with artifactory repo... "
+  if ! (whiptail --title "Repository replacement" --yes-button "Replace" --no-button "Skip" --yesno "$warningText" 10 80); then echo "SKIPPED"; return 0; fi
+  cond_redirect rm -f /etc/apt/sources.list.d/openhab2.list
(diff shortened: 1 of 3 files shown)


BClark09:
:+1: Is there a chance that `/etc/apt/sources.list.d/openhab2.list` still exists? If so we should remove it here.

mstormi:
> 
> 
> 👍 Is there a chance that `/etc/apt/sources.list.d/openhab2.list` still exists? If so we should remove it here.

Read the source, Luke ;-)

BClark09 (dismissed):

BClark09:
Doh! I'll put my glasses on next time ;)

openhabian-setup.sh
```diff
@@ -125,6 +125,7 @@ else
   load_create_config
   openhabian_console_check
   openhabian_update_check
+  fix_openhab_repo
```
ecdye, line 128: Do we want to add a check to see if the function has already been run so that it does not pop up every time even if it has already been performed?

functions/openhab.bash
```diff
@@ -299,3 +299,17 @@ dashboard_add_tile() {
 
   if echo -e "\\norg.openhab.core.ui.tiles:${application}-link-name=${tileDesc}\\norg.openhab.core.ui.tiles:${application}-link-url=${tileURL}\\norg.openhab.core.ui.tiles:${application}-link-imageurl=${tileImg}" >> "$dashboardConfig"; then echo "OK"; else echo "FAILED"; return 1; fi
 }
+
+## Replace outdated bintray repo
+##
+##    fix_openhab_repo
+##
+fix_openhab_repo() {
+  local warningText="Due to bintray shutdown of services, there is an immediate need to change the openhab stable repository configuration.\\nBintray repo in /etc/apt/sourceslist.d/openhab.list needs to be replaced by Artifactory.\\n\\nDo you want this to be replaced now ?"
```
ecdye, line 308: Proper English says that the '?' goes without a space after the end of a sentence.

ecdye (changes requested):
Overall, it looks fine. Just a few QOL concerns.

openhabian-setup.sh
mstormi: here you are

functions/openhab.bash
mstormi: well, then

ecdye (approved):
LGTM, DNT.

mstormi:
@ecdye on a sidenote
The most recent tag's name is displayed in openhabian-config window title, most users understandably think this is the current version.
Please *tag* relevant changes so we can see from the title if a patch is in a specific install.

Now with this patch in, shouldn't we forward main to openHAB3 ? openHAB3 is still named to be used in install docs.
If you agree please do (I'm always a little unsure how nest to do it. Interactively use Github web to deleting the branch & recreate it from main, that easiest. But I'm still looking for a way that creates a changelog to contain the descriptions of the patches that went in.).

ecdye:
Yes, I agree with you in that one. Tagging relevant changes is a very good idea. I'll work on forwarding main to openHAB 3 later today.

ecdye:
@mstormi on a second thought the tagline also shows the commit hash so that should be enough for us to be able to tell. It also saves us the work of creating more tags.

mstormi:
People won't name the commit hash so no that's not the same. You must not tag every commit just important ones.

ecdye:
No, what I mean is we just ask them to give us the entire tagline that it displays and then it can easily be determined, or perhaps we create a menu option that displays debug information for them to post when we ask for it.

mstormi:
I don't want to have to ask to help anyone in the first place, preferrably I want to be able answer them right away based on what they provide as info without being asked for, and that will not contain the hash.

ecdye:
I don't think you see what I mean, I mean that we create a debug menu for the issues where it is not obvious what is going on, perhaps it could even change the debug verbosity and whatnot. I see one problem with your though most people don't include the tag on the first post, and they often think they are on the latest version when the are not.

I just don't think that make certain commits worthy of tags just because we know it might break something. That is what the whole two branches idea was supposed to help mitigate in the first place. To make myself clear, I think that this is a waste of effort for us as developers because it does not help us enough, when people follow directions on the forum for how to help us help them we should be able to tell what is wrong in most cases without having to look at the commits, or revision.

mstormi:
I agree a debug menu might make sense but then again there's so much about debugging that you cannot automate without major investment (recording and parsing of logs for example) so I'm actually refraining from that.

You are right people don't always give the version in the first place but we can still make use of it, for example if you include the date like I did on the currentmost tag, you know when people upgraded last, and they know, too. You can tell or ask them "is your version older than <date>, you are affected by error XYZ".
It was not meant to be a big thingy just as a quick win so why not take it? As I said you don't have to tag everything just stuff
that makes a relevant difference.

ecdye:
I wonder if there might be a way to show the date of the commit that they are on in the tag line to make our lives easier and not have to deal with tags.

mstormi:
yep good idea. Replacing latestTag in first routine in openhabian.bash should do.

ecdye:
Will do

PR #1518, opened 2021-05-10
Simplify `tests/Dockerfile.*`

ecdye:


.github/workflows/installation-action.yml
```diff
@@ -50,15 +50,15 @@ jobs:
...
-           context: ./
+           context: .
            file: ./tests/Dockerfile.amd64-installation
            builder: ${{ steps.buildx.outputs.name }}
            tags: localhost:5000/openhabian/amd64-openhabian:latest
            push: true
        - name: openHABian installation test with 64 bit Azul Java 11 on a native amd64
          run: |
            docker run --name "openhabian-amd64" -d localhost:5000/openhabian/amd64-openhabian
-           docker exec -i "openhabian-amd64" bash -c './build.bash local-test && mv ~/.profile ~/.bash_profile && /boot/first-boot.bash'
```
mstormi, line 61: did you test to install from such an image? I don't think either but wouldn't know for sure that these settings are needed during unattended install

mstormi (approved):
LGTM but DNT so I'd expect you to build & install an image before you merge

.github/workflows/installation-action.yml
ecdye: Yep, I sure did. The new settings are what is actually required now.

Issue #1486, opened 2021-03-11
Add Zigbee controller firmware flash tool


mstormi:
https://github.com/Elelabs/elelabs-zigbee-ezsp-utility

mstormi:
on 2nd thought anyone can download for him/herself, and wrapping that in menus is too risky.

mstormi:
Deploy https://github.com/walthowd/husbzb-firmware instead ? It's a docker image I have not found any alternative for yet but it can do all Ember chipset ZigBee controllers.

mstormi:
Ember firmware to work with the elelabs tool:
https://github.com/xsp1989/zigbeeFirmware/tree/master/firmware/Zigbee3.0_Dongle

PR #1520, opened 2021-05-11
Add Elelabs ZigBee controller chipset flash tool

mstormi:
Fixes: #1486 

### DIFF
--- a/functions/system.bash
+++ b/functions/system.bash
@@ -564,17 +564,18 @@ prepare_serial_port() {
 ##
 ezspUtility_setup() {
   local repo="https://github.com/Elelabs/elelabs-zigbee-ezsp-utility"
+  local emberfwrepo="https://github.com/xsp1989/zigbeeFirmware.git"     # for iTead (Sonoff) sticks
   local destdir="/usr/share/openhab"
   local target="Elelabs_EzspFwUtility.py"
   local successText="The ELElabs python tool to flash Ember chipset ZigBee controllers has been successfully installed as ${destdir}/${target}.\\n\\nSee https://github.com/Elelabs/elelabs-zigbee-ezsp-utility#how-to how to use it."
+  local temp
 
 
   temp="$(mktemp -d "${TMPDIR:-/tmp}"/openhabian.XXXXX)"
 
   echo -n "$(timestamp) [openHABian] Installing ELElabs firmware flash tool ... "
   cond_redirect git clone "$repo" "$temp"
-  git clone https://github.com/xsp1989/zigbeeFirmware.git /tmp/zbfw
-  if ! cond_redirect pip3 install -r ${temp}/requirements.txt; then echo "FAILED (install .py)"; return 1; fi
+  if ! cond_redirect pip3 install -r ${temp}/requirements.txt; then echo "FAILED (install python requirements)"; return 1; fi
   if cond_redirect install -o openhabian -g openhabian -m 755 $target $destdir; then echo "OK" else echo "FAILED (install .py)"; return 1; fi
 
   if [[ -n $INTERACTIVE ]]; then
--- a/functions/system.bash
+++ b/functions/system.bash
@@ -563,9 +563,9 @@ prepare_serial_port() {
 ##
 ##
 ezspUtility_setup() {
-  local repo=https://github.com/Elelabs/elelabs-zigbee-ezsp-utility
-  local destdir=/usr/share/openhab
-  local target=Elelabs_EzspFwUtility.py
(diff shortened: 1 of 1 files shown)


mstormi:
I'm a bit undecided where to put it so for the time being it's just the routine to install it as I don't want to lose work so here it is.

functions/system.bash
```diff
@@ -554,3 +554,31 @@ prepare_serial_port() {
...
+## Function for installing the ELElabs EZSP firmware update utility
+## Useful to flash various ZigBee controllers that have an Ember chipset.
+##
+##    ezspUtility_setup
+##
+##
+ezspUtility_setup() {
+  local repo=https://github.com/Elelabs/elelabs-zigbee-ezsp-utility
+  local destdir=/usr/share/openhab
+  local target=Elelabs_EzspFwUtility.py
```
ecdye, line 568: Remember to double quote variables, we established that as a coding guideline a while back.

functions/system.bash
```diff
@@ -554,3 +554,31 @@ prepare_serial_port() {
...
+  local successText="The ELElabs python tool to flash Ember chipset ZigBee controllers has been successfully installed as ${destdir}/${target}.\\n\\nSee https://github.com/Elelabs/elelabs-zigbee-ezsp-utility#how-to how to use it."
+
+
+  temp="$(mktemp -d "${TMPDIR:-/tmp}"/openhabian.XXXXX)"
```
ecdye, line 572: Make a localized declaration above.

functions/system.bash
```diff
@@ -554,3 +554,31 @@ prepare_serial_port() {
...
+
+  echo -n "$(timestamp) [openHABian] Installing ELElabs firmware flash tool ... "
+  cond_redirect git clone "$repo" "$temp"
+  git clone https://github.com/xsp1989/zigbeeFirmware.git /tmp/zbfw
+  if ! cond_redirect pip3 install -r ${temp}/requirements.txt; then echo "FAILED (install .py)"; return 1; fi
```
ecdye, line 577:
```suggestion
  if ! cond_redirect pip3 install -r ${temp}/requirements.txt; then echo "FAILED (install python requirements)"; return 1; fi
```

functions/system.bash
```diff
@@ -554,3 +554,31 @@ prepare_serial_port() {
...
+  local destdir=/usr/share/openhab
+  local target=Elelabs_EzspFwUtility.py
+  local successText="The ELElabs python tool to flash Ember chipset ZigBee controllers has been successfully installed as ${destdir}/${target}.\\n\\nSee https://github.com/Elelabs/elelabs-zigbee-ezsp-utility#how-to how to use it."
+
+
+  temp="$(mktemp -d "${TMPDIR:-/tmp}"/openhabian.XXXXX)"
+
+  echo -n "$(timestamp) [openHABian] Installing ELElabs firmware flash tool ... "
+  cond_redirect git clone "$repo" "$temp"
+  git clone https://github.com/xsp1989/zigbeeFirmware.git /tmp/zbfw
```
ecdye, line 576: Why clone it into `/tmp` why not just clone both to `/opt/elelabs` and then on the second run update the repos. See the FireMotD install for an example.

ecdye (changes requested):

functions/system.bash

functions/system.bash
mstormi: ok

functions/system.bash
mstormi: ok
(btw I'd take the suggestion, however I don't know what happens when I batch-fix all of yout raised comments in my local sources...  I believe accepting a suggestion will patch my online repo. But that means that I would need to merge (pull ?) those changes into my local sources before I can proceed there.)  But I'm not sure what'll happen so that's why I didn't dare to accept suggestions so far.

functions/system.bash
mstormi: I think I got interrupted or some such. I don't want to get the firmware from that 2nd repo in all cases so I just added the URL only (no cloning) now just to remember

mstormi:
do you know what these "dependabot" PRs are ?

ecdye:
> do you know what these "dependabot" PRs are ?

Yep that's to keep the actions used in our CI up to date. I just merged them, as long as the checks related to the changes pass they should be fine to merge.

functions/system.bash
ecdye: I do the same, GitHub needs to work on that part.

ecdye (approved):
LGTM, DNT

PR #1521, opened 2021-05-12
Fix announcements to not display unless changed

ecdye:
If the users asks announcements not to be displayed this will make it so
they will not be displayed again until the maintainers have actually
changed the announcements, whereas before it would only last until the
next git update.


mstormi (approved):
LGTM DNT

PR #1536, opened 2021-05-31
Enable Karaf console and password by default

mstormi:

### DIFF
--- a/functions/openhab.bash
+++ b/functions/openhab.bash
@@ -152,9 +152,6 @@ openhab_shell_interfaces() {
       fi
     done
   fi
-  if [[ -z $sshPass ]]; then
-    sshPass="habopen"
-  fi
 
   if ! cond_redirect sed -i -e 's|^sshHost = 127.0.0.1.*$|sshHost = 0.0.0.0|g' /var/lib/openhab/etc/org.apache.karaf.shell.cfg; then echo "FAILED (sshHost)"; return 1; fi
   if cond_redirect sed -i -e 's|openhab = .*,|openhab = '"${sshPass}"',|g' /var/lib/openhab/etc/users.properties; then echo "OK"; else echo "FAILED (sshPass)"; return 1; fi
--- a/functions/openhab.bash
+++ b/functions/openhab.bash
@@ -133,7 +133,7 @@ openhab_setup() {
 openhab_shell_interfaces() {
   local introText
   local sshPass
-  local sshPass1
+  local sshPass1="habopen"
   local sshPass2
   local successText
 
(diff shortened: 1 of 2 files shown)


functions/openhab.bash
```diff
@@ -143,12 +143,12 @@ openhab_shell_interfaces() {
...
       if ! sshPass2="$(whiptail --title "Authentication setup" --passwordbox "\\nPlease confirm the password:" 9 80 3>&1 1>&2 2>&3)"; then echo "CANCELED"; return 0; fi
-      if [[ $sshPass1 == "$sshPass2" ]] && [[ ${#sshPass1} -ge 8 ]] && [[ ${#sshPass2} -ge 8 ]]; then
+      if [[ $sshPass1 == "$sshPass2" ]] && [[ ${#sshPass1} -ge 7 ]] && [[ ${#sshPass2} -ge 7 ]]; then
         sshPass="$sshPass1"
       else
-        whiptail --title "Authentication setup" --msgbox "Password mismatched, blank, or less than 8 characters... Please try again!" 7 80
+        whiptail --title "Authentication setup" --msgbox "Password mismatched, or less than 7 characters... Please try again!" 7 80
       fi
     done
   fi
```
ecdye, line 154: This will not work, it still requires interactive setup with the current code.

functions/openhab.bash
```diff
@@ -158,7 +158,6 @@ openhab_shell_interfaces() {
 
   if ! cond_redirect sed -i -e 's|^sshHost = 127.0.0.1.*$|sshHost = 0.0.0.0|g' /var/lib/openhab/etc/org.apache.karaf.shell.cfg; then echo "FAILED (sshHost)"; return 1; fi
   if cond_redirect sed -i -e 's|openhab = .*,|openhab = '"${sshPass}"',|g' /var/lib/openhab/etc/users.properties; then echo "OK"; else echo "FAILED (sshPass)"; return 1; fi
-  cond_redirect systemctl restart openhab.service
```
ecdye, line 161: Are you sure it is OK to remove this?

ecdye (changes requested):
Is this really a good idea to enable this by default? It would allow malicious actors take advantage of careless users who do not think to change the default password. Because this will make the Karaf console available on external interfaces this problem will be much worse than in the current implementation where a malicious actor needs direct access to the box to do any harm.

functions/openhab.bash
mstormi: yes you can change those files with OH up

functions/openhab.bash
mstormi: if you want to define the password yes this requires interaction (just an instant idea: we could maybe reuse userpw)
but the more important part is the editing of .cfg and that should be done unattended during install

mstormi: What about this: no password set after unattended install but if run interactively it'll deploy the default password when user does not enter anything. To manually edit sshHost is a lot more hassle to users and for remote admins it's awkward without a console.

openhabian-setup.sh
```diff
@@ -102,6 +102,7 @@ if [[ -n "$UNATTENDED" ]]; then
   vimrc_copy
   install_tailscale "install" && setup_tailscale
   misc_system_settings
+  openhab_shell_interfaces
```
ecdye, line 105: Shouldn't this go after openHAB is installed?

ecdye (changes requested):

openhabian-setup.sh
mstormi: indeed, here you are

ecdye (approved):
LGTM, DNT

Issue #1538, opened 2021-06-16
Mosquitto install not working ?


mstormi:
@ecdye see https://community.openhab.org/t/openhabian-mosquitto-mqtt-failed-to-start/102951/86

I was able to reproduce by flashing an image, installing mosquitto from the menu and then reboot.
See below for status after reboot. So the reason is `/var/log/mosquitto` is gone.
Manually purged + installed mosquitto again via menu: `/var/log/mosquitto` is back.
Rebooted again, `/var/log/mosquitto` is gone again.

There's a report this happens with openhab and homegear, too (although I don't trust that statement as it obviously isn't true for openhab) but it's sign of a common zram issue.

Note unlike some other packages we create the systemd config for, the mosquitto package comes with its own mosquitto.service so we would need to sed-edit or replace that OR use the override config as in my [post #41](https://community.openhab.org/t/openhabian-mosquitto-mqtt-failed-to-start/102951/41) see there for my fix. That is not in our code though.

```
[14:43:31] root@smarthouse:/home/admin# systemctl start mosquitto
Job for mosquitto.service failed because the control process exited with error code.
See "systemctl status mosquitto.service" and "journalctl -xe" for details.
[13:13:10] root@smarthouse:/home/admin# zramctl
NAME       ALGORITHM DISKSIZE  DATA COMPR TOTAL STREAMS MOUNTPOINT
/dev/zram3 zstd          450M 27.4M  1.2M  1.6M       4 /opt/zram/zram3
/dev/zram2 zstd          350M 16.3M  6.1K  100K       4 /opt/zram/zram2
/dev/zram1 zstd          350M 16.3M  3.7K   68K       4 /opt/zram/zram1
/dev/zram0 lzo-rle       450M    4K   86B   12K       4 [SWAP]

[13:13:01] root@smarthouse:/home/admin# cat /lib/systemd/system/mosquitto.service
[Unit]
Description=Mosquitto MQTT v3.1/v3.1.1 Broker
Documentation=man:mosquitto.conf(5) man:mosquitto(8)
After=network-online.target
Wants=network-online.target

[Service]
Type=notify
NotifyAccess=main
ExecStart=/usr/sbin/mosquitto -c /etc/mosquitto/mosquitto.conf
Restart=on-failure

[Install]
WantedBy=multi-user.target

[13:13:19] root@smarthouse:/home/admin# /usr/sbin/mosquitto -c /etc/mosquitto/mosquitto.conf
[...]

PR #1541, opened 2021-06-17
Fix services startup when logdir on zram

mstormi:
Fixes: #1538 

### DIFF
--- a/functions/influxdb+grafana.bash
+++ b/functions/influxdb+grafana.bash
@@ -215,9 +215,10 @@ influxdb_install() {
   # Disable authentication, to allow changes in existing installations
   if ! cond_redirect sed -i -e 's|auth-enabled = true|# auth-enabled = false|g' /etc/influxdb/influxdb.conf; then echo "FAILED (disable authentication)"; return 1; fi
   if ! zram_dependency install influxdb; then return 1; fi
-  if [[ -f /etc/ztab ]] && ! grep -qs "/influxdb.bind" /etc/ztab; then
+  if [[ -s /etc/ztab ]] && ! grep -qs "/influxdb.bind" /etc/ztab; then
     echo -n "$(timestamp) [openHABian] Adding InfluxDB to zram... "
-    if ! cond_redirect sed -i '/^.*persistence.bind$/a dir	zstd		150M		350M		/var/lib/influxdb		/influxdb.bind' /etc/ztab; then echo "FAILED (sed)"; return 1; fi
+    if ! mkdir -p /opt/zram/influxdb.bind/influxdb /var/lib/influxdb && chown influxdb /opt/zram/influxdb.bind/influxdb /var/lib/influxdb; then echo "FAILED (create zram database directory)"; return 1; fi
+    if ! cond_redirect sed -i '/^.*persistence.bind$/a dir	zstd		150M		350M		/var/lib/influxdb		/influxdb.bind' /etc/ztab; then echo "FAILED (editing ztab)"; return 1; fi
     if ! cond_redirect zram-config "start"; then echo "FAILED (start temporary configuration)"; return 1; fi
   fi
   if cond_redirect systemctl enable --now influxdb.service; then echo "OK"; else echo "FAILED (enable service)"; return 1; fi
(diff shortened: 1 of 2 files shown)


functions/influxdb+grafana.bash
```diff
@@ -291,7 +292,7 @@ grafana_install(){
 
   cond_echo "\\nRestarting Grafana... "
   if ! cond_redirect systemctl restart grafana-server.service; then echo "FAILED (restart service)"; return 1; fi
-  if (curl -4 --retry 6 --retry-connrefused --silent --head http://localhost:3000); then echo "OK"; else echo "FAILED (wait for Grafana to start)"; return 1; fi
+  if (curl -4 -m 120 --retry 20 --retry-connrefused --silent --head http://localhost:3000); then echo "OK"; else echo "FAILED (wait for Grafana to start)"; return 1; fi
```
ecdye, line 295: This seems like a lot more time, what changed from before to cause it to no longer work in a timely fashion?

functions/packages.bash
```diff
@@ -284,9 +285,8 @@ mqtt_setup() {
     if ! cond_redirect sed -i -e '/allow_anonymous/d' /etc/mosquitto/mosquitto.conf; then echo "FAILED"; return 1; fi
     if cond_redirect rm -f /etc/mosquitto/passwd; then echo "OK"; else echo "FAILED"; return 1; fi
   fi
-  cond_redirect mkdir -p /var/log/mosquitto
-  cond_redirect chown mosquitto /var/log/mosquitto
   grep -qs "/log.bind" /etc/ztab && if ! zram_dependency install mosquitto; then return 1; fi
+  if ! mkdir -p /opt/zram/log.bind/mosquitto /var/log/mosquitto && chown mosquitto /opt/zram/log.bind/mosquitto /var/log/mosquitto; then echo "FAILED (create zram logdir)"; return 1; fi
```
ecdye, line 289: My one concern is that I have Mosquitto installed and running on my personal machine and never had an issue with it not starting, What could be making it not work now? Same for all of the other instances of this as well, why have we never observed this behavior before now, especially because I have personally seen them work.

functions/influxdb+grafana.bash
```diff
@@ -265,6 +265,7 @@ grafana_install(){
   if ! cond_redirect chmod -R 0750 /var/run/grafana; then echo "FAILED (chmod)"; return 1; fi
   if ! cond_redirect chown -R grafana:grafana /var/run/grafana; then echo "FAILED (chown)"; return 1; fi
   if ! zram_dependency install grafana; then return 1; fi
+  if ! mkdir -p /opt/zram/log.bind/grafana /var/log/grafana && chown grafana /opt/zram/log.bind/grafana /var/log/grafana; then echo "FAILED (create zram logdir)"; return 1; fi
```
ecdye, line 268: Might want to check if zram is installed before running as if it is not this is unnecessary. Please apply this feedback to all other instances.

ecdye (changes requested):

functions/packages.bash
mstormi: See https://github.com/openhab/openhabian/issues/1538#issuecomment-863225019
Me neither but I and probably you as well we have manually installed mosquitto. Once the logdir on the lower FS is created it'll work from there on. But that doesn't happen in a fresh image based install.

functions/influxdb+grafana.bash
mstormi: frankly I don't know

functions/influxdb+grafana.bash
mstormi: hmm, this will complicate the code do you insist ?

functions/packages.bash
ecdye: My question is why is this a bug that keeps recurring, my only guess is that somehow the overlayfs-tools are not properly handling the directory creation in the log dir, but I'm not sure why. For now it should work as a solution but I am still wanting to find the cause of the problem.

functions/influxdb+grafana.bash
mstormi: Let me know how you think we should proceed. Merge as-is? If no, which missing changes do you consider mandatory ?

ecdye: I don't think that it will be an issue as the bug has to do with directories with special permissions not being correctly transferred to the lower directory. In a regular installation it won't hurt to skip the zram stuff if it is not enabled. I would go ahead and add a check that skips the permissions if zram is not installed. After that I think that is my only issue.

Once again, sorry for the delay on this I apparently missed your first comment on that one and never responded.

mstormi:
like so ? If ok you can merge.

functions/influxdb+grafana.bash
mstormi: Actually the same code was in there for InfluxDB, too, and the BATS test for InfluxDB failed in our CI.
After I applied the same fix it now works. Fix for both Influx and Grafana are part of the PR.

ecdye (approved):
LGTM

Issue #1097, opened 2020-08-22
HM-MOD-RPI-PCB GPIO access not working for Homegear


harabeck:
## Issue information:
Hardware wise BidCos RF Module (HM-MOD-RPI-PCB) is simply connected to the UART (RX - GPIO15, TX - GPIO14) which is provided by dtoverlay=miniuart-bt in /boot/config.txt
and the ReSeT is connected to GPIO18 on the Raspberry PI3 and probably 4 as well . The GPIO18 causes all the trouble.

## Debug information:
https://community.openhab.org/t/hm-mod-rpi-pcb-raspberry-pi-4/84576
https://forum.homegear.eu/t/openhabian-homegear-fehler-could-not-connect-to-socket-permission-denied/3609/30

## System information:
RaspBerry Pi 3 + HM-MOD-RPI-PCB
originally everything worked fine with Openhab installation 2.5.4 until the update to 2.5.7
I tried the Strech and Buster version.
openhabianpi-raspbian-201804031720-gitdba76f6-crc9e93c3eb.img.xz
openhabian-pi-raspbian-201908050414-gitca0976f-crc6a66b5a1.img.xz

xasz:
I tried setting up the permissions as you told, and the problem troubles more users with the HM-MOD-RPI-PCB going to a newer openhabian version. For now i did not find a solution.

fmeies:
After updating to openHABian Buster, I had the same issues with my HM-MOD-RPI-PCB. From my point of view, this is not an ownership or permission issue since I was perfectly able to export gpio18 as user homegear manually. The problem was the missing initialization of gpio18. The [homegear docu advises](https://doc.homegear.eu/homegear/installation.html) to add this to `/etc/systemd/system/homegear.service`: 
```
ExecStartPre=/usr/bin/gpio export 18 out
ExecStop=/usr/bin/gpio unexport 18
```
Maybe _Fix permissions_ should add these lines to `homegear.service` automatically?

mstormi:
I could add it to homegear.service.

However, even with the package from homegear install docs installed, /usr/bin/gpio does not exist on Raspbian.
Can you do `dpkg -S /usr/bin/gpio` ?

fmeies:
[...]

PR #1550, opened 2021-07-09
Reset GPIO 18 on Homegear start to support HM-MOD-RPI-PCB controller

mstormi:
Related: #1097 


functions/packages.bash
```diff
@@ -223,7 +223,7 @@ homegear_setup() {
   if ! cond_redirect install -m 644 "${BASEDIR:-/opt/openhabian}"/includes/homegear-management.service /etc/systemd/system/homegear-management.service; then echo "FAILED (copy service)"; return 1; fi
   if ! cond_redirect rm -f /lib/systemd/system/homegear*; then echo "FAILED (clean default service)"; return 1; fi
   if running_in_docker; then sed -i '/RuntimeDirectory/d' /etc/systemd/system/homegear*; fi
-  if ! zram_dependency install homegear homegear-management; then return 1; fi
+  if ! zram_dependency install homegear homegear-nodes-core homegear-management wiringpi; then return 1; fi
```
ecdye, line 226: I know why you added the `wiringpi` package but why did you add the `homegear-nodes-core` package?

ecdye (changes requested):
Please fix tests as they currently do not all pass.

functions/packages.bash
mstormi: because that's part of the homegear install instructions

mstormi:
> 
> 
> Please fix tests as they currently do not all pass.

I have no idea why they fail. Can't imagine it's this change though.
Frankly I don't know how to debug this for the time being. Do you have any idea ?

ecdye:
Well the test that failed IIRC was part of the homegear package which was changed by this PR. Maybe because the emulation does not know about GPIO pins and therefore fails. Fair warning that is just a wild guess.

mstormi:
I've really forgotten how to debug that, could you give me a hint please ?


For a test to approach the cause, I've removed everything but the mod to systemd config which is the core of the PR, and BATS still fails so this seems to be the cause. That now is something like `ExecStartPre=/usr/bin/gpio export 18 out` so it accesses HW which very likely fails (can't really see the command's output) on our virtual HW.
Any idea how to handle that other than to disable the test?

functions/packages.bash
mstormi: but you're right, I have not encountered a true need so removed it again

ecdye:
It was the `if [ "$status" -ne 0 ]; then echo "$output" >&3; fi` command we just placed extras while debugging to find errors or we would just echo the output anyway to see what was happening.

mstormi:
Thanks. Yes it was the access to the GPIO HW that didn't work (for obvious reasons).
See how I solved that now and merge if you are ok with that.

ecdye (approved):
Hmmm, I feel like it might be a little unnecessary to use that solution but if you think it is best go ahead and merge. An alternative could be to use the `running_in_docker` command from the helpers file.

mstormi:
> An alternative could be to use the `running_in_docker` command from the helpers file.

Trouble is, it's initiated from the systemd config so we would need to dynamically generate that.
And that would still not be enough, compatibility IMHO should always be validated on startup not just at install time.
Think what if someone changes HW or packages after install or uses non-RPi HW (possibly incompatible with wiringpi etc).

PR #1552, opened 2021-07-14
Add or reset tags on Tailscale install

mstormi:
to allow for tagging of clients (use in ACLs etc)

### DIFF
--- a/functions/vpn.bash
+++ b/functions/vpn.bash
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+# shellcheck disable=SC2154
+
 ## Install wireguard from unstable Debian as long as it is not in the Raspbian repo
 ## Valid arguments: "install" or "remove"
 ##
@@ -184,7 +186,6 @@ install_tailscale() {
   local sudoersPath="/etc/sudoers.d"
 
   if [[ -n "$UNATTENDED" ]]; then
-    # shellcheck disable=SC2154
     if [[ ! -v preauthkey ]]; then echo "$(timestamp) [openHABian] tailscale VPN installation... SKIPPED (no preauthkey defined)"; return 1; fi
   fi
 
@@ -227,7 +228,6 @@ install_tailscale() {
 ##   setup_tailscale(String action)
 ##
 setup_tailscale() {
-  # shellcheck disable=SC2154
   local preAuthKey=${preauthkey}
   local tags="${tstags}"
   local consoleProperties="${OPENHAB_USERDATA:-/var/lib/openhab}/etc/org.apache.karaf.shell.cfg"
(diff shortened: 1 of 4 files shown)


functions/vpn.bash
```diff
@@ -228,6 +228,8 @@ install_tailscale() {
 ##
 setup_tailscale() {
   local preAuthKey=${preauthkey}
+  # shellcheck disable=SC2154
+  local tags=${tstags}
```
ecdye, line 232: We should be double quoting variable assignments whenever possible.

functions/vpn.bash
```diff
@@ -228,6 +228,8 @@ install_tailscale() {
 ##
 setup_tailscale() {
   local preAuthKey=${preauthkey}
+  # shellcheck disable=SC2154
```
ecdye, line 231: I would move this to the top of the function because it applies to multiple areas in the function.

functions/vpn.bash
```diff
@@ -239,7 +241,8 @@ setup_tailscale() {
     if ! preAuthKey="$(whiptail --title "Enter pre auth key" --inputbox "\\nIf you have not received / created the tailscale pre auth key at this stage, please do so now or tell your administrator to. This can be done on the admin console. There's a menu option on the tailscale Windows client to lead you there.\\n\\nPlease enter the tailscale pre auth key for this system:" 11 80 "$preAuthKey" 3>&1 1>&2 2>&3)"; then echo "CANCELED"; return 0; fi
   fi
 
-  if ! tailscale up --authkey "${preAuthKey}"; then echo "FAILED (join tailscale VPN)"; return 1; fi
+  # if $tstags is empty, this will reset existing tags
```
ecdye, line 244: Clarification if `${tags}/${tstags}` if empty would be a better comment.

ecdye (changes requested):
TBH I don't know exactly why this is important but the code looks pretty good, just a couple of thoughts.

functions/vpn.bash
mstormi: not needed here but ok

functions/vpn.bash
mstormi: ok

functions/vpn.bash
mstormi: ok

mstormi:
> I don't know exactly why this is important

Well I guess you haven't used tailscale yet, I myself just started discovering a number of features it provides.
Actually for security reasons I was looking for a means to have a asynchronous VPN where a master can log into/provision slave sites but not the other way round.
You can configure ACLs with JSON style input, and they're based on tags so tags need to be applied to devices which is what we now can do with this PR.

functions/vpn.bash
ecdye: Actually, this is used ~5-6 times throughout this file why not make it global for the file by placing it at the top.

mstormi: hmm, we didn't in other files. But ok here

ecdye (approved):
LGTM, DNT

PR #1553, opened 2021-07-16
Migrate openHABian to OH3 on menu 03

mstormi:
dunno why this slipped before
See e.g. https://community.openhab.org/t/openhab-3-1-successful-but-openhabian-putty-always-starts-in-2-5-12-1/124518/11

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,13 @@ Hit tab to unselect buttons and scroll through the text using UP/DOWN or
 PGUP/PGDN. All announcements are stored in `/opt/openhabian/docs/CHANGELOG.md`
 for you to lookup.
 
+## Install openHAB function changed ## July 23, 2021
+Menu option 03 "Install or upgrade to openHAB release 3" was changed to now
+actually do what most users have been expecting it to: it now also upgrades
+the environment to work with openHAB3 and installs openHAB3 "stable" version.
+Previously, you had to additionally use menu option 42 to migrate the
+environment.
+
 ## Telldus Core service removed ## May 20, 2021
 The Telldus Core service has now been removed from openHABian, and will no
 longer receive active support from the openHABian developers. Existing
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -1,3 +1,10 @@
+## Install openHAB function changed ## July 23, 2021
+Menu option 03 "Install or upgrade to openHAB release 3" was changed to now
+actually do what most users have been expecting it to: it now also upgrades
+the environment to work with openHAB3 and installs openHAB3 "stable" version.
+Previously, you had to additionally use menu option 42 to migrate the
+environment.
+
 ## Telldus Core service removed ## May 20, 2021
 The Telldus Core service has now been removed from openHABian, and will no
 longer receive active support from the openHABian developers. Existing
--- a/functions/menu.bash
+++ b/functions/menu.bash
@@ -63,7 +63,7 @@ show_main_menu() {
 
   elif [[ "$choice" == "03"* ]]; then
     wait_for_apt_to_finish_update
-    openhab_setup openHAB3 "stable"
+    migrate_installation "openHAB3"
 
   elif [[ "$choice" == "10"* ]]; then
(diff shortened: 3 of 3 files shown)


ecdye (changes requested):
This doesn't work for existing openHAB3 or openHAB2 installations to upgrade them. Personally I think that we should not worry about trying to upgrade from menu option 3 as we have the option in a menu for a reason. 

As a very hypothetical situation, what if a user only wanted the latest release of OH2 and used this option figuring that it would not upgrade to OH3 but then was upset when it did.

The user on the forum started completely wrong anyway, as they should have used the migrate option on the stable branch to start the upgrade properly. Perhaps a better solution is simply to make a new announcement on the stable branch that never goes away (ie displays every time openhabian-config is opened) that states that OH2 is no longer supported and explains what menu option to use to automatically upgrade to OH3.

mstormi:
Hmm. We need an option to work for the majority of people, more than are successful with the current code.
Sure we can't handle all cases but we should try to at least handle the case "user selects 03 without reading the docs" because while that's naive, that's what the official docs tell users to do.

I agree this proposal here might be insufficient but I'm unsure why you think it wouldn't work in the cases you named.
(originally it was migrate && (original openhab_setup) but turned out that's called during migration).
Please help with developing a more foolproof solution: it's better to check some prerequisites and fail if they're not met rather than to execute and fail after.

ecdye:
The `migrate_installation` code will automatically stop if it sees that the user is already on OH3 that is why I said it will not work with the way it is now. If you insist I is probably ok to include migrate in menu option 3 but this is not a functional solution.

mstormi:
> The `migrate_installation` code will automatically stop if it sees that the user is already on OH3

But that is the correct, expected/promised behavior, isn't it ?
Again: 03 promises to "Install or upgrade to OH3" (it does not say "upgrade OH3", that's what option 02 is for).
Ok strictly speaking the "install" use case does not exist as some OH version is *always* installed (unless a user manually removed it himself outside of openhabian-config, but even in that case 03 would do what it promises to, wouldn't it ?)

ecdye:
Hmm, I guess if you put it like that yes that works. If you go through with this then we need to have Kai change the release notes to reflect the new behavior.

My only concern then is that in a way this is a very language specific thing and perhaps in the failed message we should redirect them to menu option 2.

mstormi:
> If you go through with this then we need to have Kai change the release notes to reflect the new behavior.

Not necessarily so. They apply to the case that the user has 2.X installed in which case this PR should deliver what it's supposed to (sure it would be a good idea nonetheless to point that out in the rel notes but needed it isn't).
Maybe something in the 00 text ? Or a popup on both, 02 and 03 ?

ecdye:
I don't know, I feel like we should just have the release notes changed in the future and possibly just note the behavior change in our announcements. 

I don't think that much more than that is necessary. We may want to change the behavior in the stable branch too so that it's consistent across openHABian.

mstormi:
So do I understand it right, you want:
* add migration call (i.e. merge current PR)
* add a note to the OH rel notes that in order to *upgrade* you must not use 03 but 02
  (regarding 03, it's not really a *change* in behavior but merely a *fix* - it now does what it always has been promising it would do
* add this to our own OHian rel notes

ecdye:
Yes, although I disagree it is a change in behavior because something that worked before no longer does the way it used to, even if it is a change to make it do what it says it will.

mstormi:
fyi, I wanted to add a small note the README on dedicating the OH box vs. installing other software.
It might have gotten a little larger and more emotional than intended. Please check.

ecdye:
> fyi, I wanted to add a small note the README on dedicating the OH box vs. installing other software.
> It might have gotten a little larger and more emotional than intended. Please check.

I changed it slightly but mostly just in line with our formatting standards, it seemed fine to me. I think it was a good inclusion that helped clarify the things we are willing and able to support.

mstormi:
Another one bites the dust
https://community.openhab.org/t/clear-items-after-updating-oh2-via-openhabian/123093/12

So changes are ok ?

ecdye:
An you make the announcement for our repo as a part of this PR? Other than that yes the changes are fine.

mstormi:
here you are

ecdye (approved):

mstormi:
FWIW it was already in stable just not in openHAB3
https://github.com/openhab/openhabian/blob/stable/functions/menu.bash#L66

PR #1558, opened 2021-07-29
Forward Docker base CI image from buster to bullseye

mstormi:
Mostly LGTM

Only thing I noticed is https://github.com/mstormi/openhabian/runs/3190479326?check_suite_focus=true#step:8:1389

```
W: Skipping acquire of configured file 'ui/binary-armhf/Packages' as repository 'http://archive.raspberrypi.org/debian bullseye InRelease' doesn't have the component 'ui' (component misspelt in sources.list?)
```

### DIFF
--- a/functions/influxdb+grafana.bash
+++ b/functions/influxdb+grafana.bash
@@ -196,10 +196,10 @@ influxdb_install() {
   else
     myOS="$(lsb_release -si)"
   fi
-  #myRelease="$(lsb_release -sc)"
-  #if [[ "$myRelease" == "n/a" ]]; then
-  myRelease=${osrelease:-buster}        # no InfluxDB bullseye repo available
-  #fi
+  myRelease="$(lsb_release -sc)"
+  if [[ "$myRelease" == "n/a" ]]; then
+    myRelease=${osrelease:-buster}
+  fi
 
   if ! dpkg -s 'influxdb' &> /dev/null; then
     if ! add_keys "https://repos.influxdata.com/influxdb.key"; then return 1; fi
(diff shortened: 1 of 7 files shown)


ecdye (approved):
LGTM, but we should probably wait to merge until the 14th of next month at the earliest.

ecdye:
Also, Grafana doesn't seem to be installing correctly with Bullseye according to the BATS tests

mstormi:
Nah, we should upgrade CI asap. We will need CI to work with both, buster AND bullseye as we will have to keep supporting users with both of these - most will not upgrade their buster OS. I've revoked the rpi2 test to keep running based on a buster image.

Grafana, I believe the test fails because InfluxDB installation dynamically creates the repo name but there is no InfluxDB repo for bullseye yet so for the time being, I've changed that to to be static the buster repo. Let's see if that proceeds.

ecdye:
> Nah, we should upgrade CI asap. We will need CI to work with both, buster AND bullseye as we will have to keep supporting users with both of these - most will not upgrade their buster OS.

Fair enough, do you want to work on putting together a draft of a guide on how to upgrade to Bullseye from an existing system?

> Grafana, I believe the test fails because InfluxDB installation dynamically creates the repo name but there is no InfluxDB repo for bullseye yet so for the time being, I've changed that to to be static the buster repo. Let's see if that proceeds.

Doesn't seem to have worked this time

mstormi:
> Fair enough, do you want to work on putting together a draft of a guide on how to upgrade to Bullseye from an existing system?

No. Anyone who wants to can follow Debian guides Anyone else we recommend to reinstall.

> Doesn't seem to have worked this time

Pls have a closer look.
I think my InfluxDB change worked fine (so will merge this if you don't mind).
Something's wrong with the Grafana repo, though. Not sure if we can do anything about that.

mstormi:
I fixed Grafana install (it was slightly bad repo also for buster) but I have been unable to fix InfluxDB and Homegear so far.
Please cross-check and eventually merge this as a correct but clearly not final patch please

ecdye:
I'm going to call this good enough for now. 

FYI, I'll be going out of town tomorrow for a family trip. I won't be able to be online very much during that trip, I should be back by Friday evening of next week.

mstormi:
thanks for the info. Wish you a great time!

Issue #1563, opened 2021-08-03
Default hostname in disklist although hostname changed in openhabian.conf


mstormi:
See https://community.openhab.org/t/openhabian-unattended-amanda-not-installed/125040/15

He changed hostname in openhabian.conf then installed Amanda unattended, this resulted in the dynamically generated disklist
to contain the default hostname (openhabian) instead of the new one.
Not sure how that happened to sneak in,only occurences I see are in build-image/webserver.bash:6 or 27
=> replace HOSTNAME by hostname in backup.bash and eventually elsewhere

PR #1566, opened 2021-08-05
Prefer `$hostname` to allow for proper first setup

mstormi:
not 100% sure it won't have side effects so please have a closer look

Fixes: #1563

### DIFF
--- a/build-image/webserver.bash
+++ b/build-image/webserver.bash
@@ -3,7 +3,7 @@
 ## start:             Starts a minimalistic web server that shows the status of the
 ##                    openHABian installation.
 ## reinsure_running:  Checks if webserver is running.
-## inst_done:         Create finish message and link to http://${HOSTNAME:-openhabian}:8080.
+## inst_done:         Create finish message and link to http://${hostname:-openhabian}:8080.
 ## cleanup:           Stops the webserver and removes all no longer needed files.
 
 port="80"
@@ -24,7 +24,7 @@ fi
 
 if [[ $1 == "inst_done" ]]; then
   mkdir -p "${TMPDIR:-/tmp}"/webserver
-  sed 's|%HOSTNAME|'"${HOSTNAME:-openhabian}"'|g' /opt/openhabian/includes/webserver/install-complete.html > "${TMPDIR:-/tmp}"/webserver/index.html
+  sed 's|%HOSTNAME|'"${hostname:-openhabian}"'|g' /opt/openhabian/includes/webserver/install-complete.html > "${TMPDIR:-/tmp}"/webserver/index.html
 fi
 
 if [[ $1 == "cleanup" ]]; then
--- a/functions/backup.bash
+++ b/functions/backup.bash
@@ -161,7 +161,7 @@ create_amanda_config() {
   if ! cond_redirect mkdir -p "$configDir" "$databaseDir" "$logDir" "$indexDir"; then echo "FAILED (create directories)"; return 1; fi
   if ! cond_redirect touch "${configDir}/tapelist"; then echo "FAILED (touch tapelist)"; return 1; fi
   # shellcheck disable=SC2154
-  ip=$(dig +short "$hostname")
+  ip=$(dig +short "$HOSTNAME")
   revip=$(host "${ip}" | cut -d' ' -f1)
   if [[ -n "$revip" ]]; then (echo -e "${ip} ${backupUser} amdump\\n${revip} ${backupUser} amdump" > "$amandaHosts"); fi
   if ! (echo -e "${hostname} ${backupUser} amdump\\n${hostname} root amindexd amidxtaped\\nlocalhost ${backupUser}\\nlocalhost root amindexd amidxtaped" >> "$amandaHosts"); then echo "FAILED (Amanda hosts)"; return 1; fi
--- a/functions/backup.bash
+++ b/functions/backup.bash
@@ -160,10 +160,10 @@ create_amanda_config() {
   echo -n "$(timestamp) [openHABian] Creating Amanda filesystem... "
   if ! cond_redirect mkdir -p "$configDir" "$databaseDir" "$logDir" "$indexDir"; then echo "FAILED (create directories)"; return 1; fi
   if ! cond_redirect touch "${configDir}/tapelist"; then echo "FAILED (touch tapelist)"; return 1; fi
-  # shellcheck disable=SC2154
   ip=$(dig +short "$HOSTNAME")
   revip=$(host "${ip}" | cut -d' ' -f1)
   if [[ -n "$revip" ]]; then (echo -e "${ip} ${backupUser} amdump\\n${revip} ${backupUser} amdump" > "$amandaHosts"); fi
+  # shellcheck disable=SC2154
   if ! (echo -e "${hostname} ${backupUser} amdump\\n${hostname} root amindexd amidxtaped\\nlocalhost ${backupUser}\\nlocalhost root amindexd amidxtaped" >> "$amandaHosts"); then echo "FAILED (Amanda hosts)"; return 1; fi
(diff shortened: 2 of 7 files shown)


build-image/webserver.bash
```diff
@@ -24,7 +24,7 @@ fi
 
 if [[ $1 == "inst_done" ]]; then
   mkdir -p "${TMPDIR:-/tmp}"/webserver
-  sed 's|%HOSTNAME|'"${HOSTNAME:-openhabian}"'|g' /opt/openhabian/includes/webserver/install-complete.html > "${TMPDIR:-/tmp}"/webserver/index.html
+  sed 's|%HOSTNAME|'"${hostname:-openhabian}"'|g' /opt/openhabian/includes/webserver/install-complete.html > "${TMPDIR:-/tmp}"/webserver/index.html
```
ecdye, line 27: I believe this should stay as it was before because it will otherwise try to broadcast to an hostname that is not currently its own. this only run during the first boot setup anyway.

functions/backup.bash
```diff
@@ -160,10 +160,11 @@ create_amanda_config() {
   echo -n "$(timestamp) [openHABian] Creating Amanda filesystem... "
   if ! cond_redirect mkdir -p "$configDir" "$databaseDir" "$logDir" "$indexDir"; then echo "FAILED (create directories)"; return 1; fi
   if ! cond_redirect touch "${configDir}/tapelist"; then echo "FAILED (touch tapelist)"; return 1; fi
-  ip=$(dig +short "$HOSTNAME")
+  # shellcheck disable=SC2154
+  ip=$(dig +short "$hostname")
```
ecdye, line 164: Will this work if the hostname that the machine currently has and its hostname that the user wants to set are different?

functions/find.bash
```diff
@@ -28,7 +30,8 @@ find3_setup() {
...
+  local successText="FIND3 setup was successful.\\n\\nSettings can be configured in '/etc/default/find3server'. Be sure to restart the service after.\\n\\nYou must manually activate 'find3-cli-scanner', for more information see:\\nhttps://www.internalpositioning.com/doc/cli-scanner.md\\n\\nThere is also an Android app for collecting fingerprints. There is no iOS app for fingerprint collection, for details on why see:\\nhttps://www.internalpositioning.com/doc/faq.md#iphone\\n\\nCheck out your
+  FIND3 server's dashboard at: http://${hostname}:8003\\n\\nFor more information see:\\nhttps://www.internalpositioning.com/doc/"
```
ecdye, line 34: Get rid of the extra blank line here that you added

functions/openhab.bash
```diff
@@ -138,7 +140,7 @@ openhab_shell_interfaces() {
   local successText
 
   introText="\\nThe openHAB remote console is a powerful tool for every openHAB user. For details see https://www.openhab.org/docs/administration/console.html\\n\\nThis menu option will make the console available on all interfaces of your system. Please provide a secure password for this connection. Blank input will get you the default password \"habopen\"."
-  successText="The openHAB remote console was successfully activated on all interfaces and openHAB has been restarted. You can reach the console using\\n\\nssh openhab@$HOSTNAME -p 8101\\n\\nBe aware that the first connection attempt may take a few minutes or may result in a timeout due to key generation.\\nThe default password is \"habopen\"."
+  successText="The openHAB remote console was successfully activated on all interfaces and openHAB has been restarted. You can reach the console using\\n\\nssh openhab@$hostname -p 8101\\n\\nBe aware that the first connection attempt may take a few minutes or may result in a timeout due to key generation.\\nThe default password is \"habopen\"."
```
ecdye, line 143: We should probably add the brackets around hostname here for consistency.

```suggestion
  successText="The openHAB remote console was successfully activated on all interfaces and openHAB has been restarted. You can reach the console using\\n\\nssh openhab@${hostname} -p 8101\\n\\nBe aware that the first connection attempt may take a few minutes or may result in a timeout due to key generation.\\nThe default password is \"habopen\"."
```

functions/packages.bash
```diff
@@ -117,8 +118,8 @@ exim_setup() {
   fi
   if cond_redirect install_dnsutils; then echo "OK"; else echo "FAILED"; return 1; fi
 
-  interfaces="$(dig +short "$HOSTNAME" | tr '\n' ';')127.0.0.1;::1"
-  relaynets="$(dig +short "$HOSTNAME" | cut -d'.' -f1-3).0/24"
+  interfaces="$(dig +short "$hostname" | tr '\n' ';')127.0.0.1;::1"
+  relaynets="$(dig +short "$hostname" | cut -d'.' -f1-3).0/24"
```
ecdye, line 122: Same question as earlier, does this work if the hostnames are not the same?

ecdye (changes requested):

functions/packages.bash
mstormi: this didn't change so assuming it worked before it still will

functions/find.bash
mstormi: ok reverted

functions/backup.bash
mstormi: $hostname is taken (early on start) from openhabian.conf so I think yes

ecdye: But does it take effect before a reboot? If it did then this entire PR probably wouldn't be necessary.

build-image/webserver.bash
mstormi: $hostname is taken from openhabian.conf, sourced at the beginning of first-boot.bash so it should be available here.

Not sure but isn't it the other way round that *if* user changed it (remember: not interactively but in openhabian.conf) ?
Then at this stage the old code ($HOSTNAME) has been wrong ever, hasn't it ? So that was a bug that has gone unspotted so far.

functions/backup.bash
mstormi: The point is whatever is in $hostname is e.g. written to disklist. 
It does not matter when it takes effect.

ecdye: But it might in this case because wouldn't dig be expecting the current hostname when it was looking for an IP not the hostname that the user has set but that hasn't taken effect. Do you get what I mean?

mstormi: I get it. Frankly I'm not sure when in the install process it takes effect but amanda install is always *after* a reboot so I hoped there it did for sure. But we should test.

mstormi: Then again HOSTNAME has worked before and there's no need to change it. So I just reverted this change.

build-image/webserver.bash
ecdye: But it does work, I know it does because I have used it on every install I have performed. Maybe this just boils down to me being cautious about changing things that work now.

functions/packages.bash
ecdye: It did change though you changed it from $HOSTNAME to $hostname.

build-image/webserver.bash
mstormi: > I have used it on every install I have performed

you mean you changed hostname via openhabian.conf *and* watched the webserver output on each of these ?
That's the minimum conditions for the bug to show (if there was/is, as I said not sure).

mstormi: how to proceed, is this the only requested change left? Saw my response?

ecdye: No, I have changed the hostname and the output worked correctly on the original default hostname of openhabian.

mstormi: Err I don't understand that sentence. What code do you refer to, old or new? And what does that have to do with the default hostname ? Do you mean you tried changing it but it rather used the default ("openhabian") ?

ecdye: Nvm, I think I misread. It should be fine. I don't think it will break anything and if it does we can always revert and try again.

ecdye (approved):
LGTM, DNT

PR #1567, opened 2021-08-10
Minor change in knxd info text

udo1toni:
knxd config is done via /etc/knxd.conf (at least when using systemd)


ecdye:
Thanks for your contribution @udo1toni! Can you fix the DCO by following the instructions that are given when you hit details on the failed check.

udo1toni:
Done (hope so) :)

PR #1579, opened 2021-09-08
Mount logs under /srv/

spacemanspiff2007:
Signed-off-by: Sebastian Gerber git@sgerber.de


mstormi (changes requested):
Err ... no. Sharing log would create issues. It's intentionally absent.

spacemanspiff2007:
Could you elaborate what kind of issues so it's properly documented?
I've been running an installation like that and it worked without problems.

ecdye:
It tends to mess with zram because the log directory for openhab is then being mounted in multiple places and often has been observed to be a cause of data loss in logs. It doesn't always break but often does especially when random chance is acting on a very large group of users.

PR #1584, opened 2021-09-25
Add option to en/disable framebuffer on RPi

mstormi:

### DIFF
--- a/functions/menu.bash
+++ b/functions/menu.bash
@@ -136,13 +136,15 @@ show_main_menu() {
     "   | Disable NTP"            "Disable time synchronization via systemd-timesyncd to NTP servers" \
     "34 | Change passwords"       "Change passwords for Samba, openHAB Console or the system user" \
     "35 | Serial port"            "Prepare serial ports for peripherals like Razberry, SCC, Pine64 ZWave, ..." \
-    "36 | WiFi setup"             "Configure wireless network connection" \
+    "36 | Disable framebuffer"    "Disable framebuffer on RPi to minimize memory usage" \
+    "   | Enable framebuffer"     "Enable framebuffer (standard setting)" \
+    "37 | WiFi setup"             "Configure wireless network connection" \
     "   | Disable WiFi"           "Disable wireless network connection" \
-    "37 | Move root to USB"       "Move the system root from the SD card to a USB device (SSD or stick)" \
     "38 | Use zram"               "Use compressed RAM/disk sync for active directories to avoid SD card corruption" \
     "   | Uninstall zram"         "Don't use compressed memory (back to standard Raspberry Pi OS filesystem layout)" \
-    "39 | Setup Exim Mail Relay"  "Install Exim4 to relay mails via public email provider" \
-    "3A | Setup Tailscale VPN"    "Establish or join a WireGuard based VPN using the Tailscale service" \
+    "39 | Move root to USB"       "Move the system root from the SD card to a USB device (SSD or stick)" \
+    "3A | Setup Exim Mail Relay"  "Install Exim4 to relay mails via public email provider" \
+    "3B | Setup Tailscale VPN"    "Establish or join a WireGuard based VPN using the Tailscale service" \
     "   | Remove Tailscale VPN"   "Remove the Tailscale VPN service" \
     "   | Install WireGuard"      "Setup WireGuard to enable secure remote access to this openHABian system" \
     "   | Remove WireGuard"       "Remove WireGuard VPN from this system" \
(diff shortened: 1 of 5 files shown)


functions/menu.bash
```diff
@@ -136,13 +136,15 @@ show_main_menu() {
     "   | Disable NTP"            "Disable time synchronization via systemd-timesyncd to NTP servers" \
     "34 | Change passwords"       "Change passwords for Samba, openHAB Console or the system user" \
     "35 | Serial port"            "Prepare serial ports for peripherals like Razberry, SCC, Pine64 ZWave, ..." \
-    "36 | WiFi setup"             "Configure wireless network connection" \
+    "36 | Disable framebuffer"    "Disable framebuffer on RPi to minimize memory usage" \
```
ecdye, line 139: You need to make sure to expand the menu size to accommodate the new menu options.

functions/menu.bash
```diff
@@ -136,13 +136,15 @@ show_main_menu() {
...
+    "   | Enable framebuffer"     "Enable framebuffer (standard setting)" \
+    "37 | WiFi setup"             "Configure wireless network connection" \
     "   | Disable WiFi"           "Disable wireless network connection" \
-    "37 | Move root to USB"       "Move the system root from the SD card to a USB device (SSD or stick)" \
     "38 | Use zram"               "Use compressed RAM/disk sync for active directories to avoid SD card corruption" \
     "   | Uninstall zram"         "Don't use compressed memory (back to standard Raspberry Pi OS filesystem layout)" \
-    "39 | Setup Exim Mail Relay"  "Install Exim4 to relay mails via public email provider" \
-    "3A | Setup Tailscale VPN"    "Establish or join a WireGuard based VPN using the Tailscale service" \
+    "39 | Move root to USB"       "Move the system root from the SD card to a USB device (SSD or stick)" \
```
ecdye, line 145: Not really related, but do we really still want to support this?

functions/menu.bash
```diff
@@ -157,13 +159,15 @@ show_main_menu() {
       *Disable\ NTP) setup_ntp "disable" ;;
       34\ *) change_password ;;
       35\ *) prepare_serial_port ;;
-      36\ *) configure_wifi setup;;
+      36\ *) use_framebuffer disable;;
+      *Enable\ framebuffer) use_framebuffer enable ;;
```
ecdye, line 163: double quote for safety

functions/system.bash
```diff
@@ -471,6 +471,26 @@ memory_split() {
   else
     if echo "gpu_mem=16" >> /boot/config.txt; then echo "OK"; else echo "FAILED"; return 1; fi
   fi
+
+  sed -i '/^dtoverlay=vc4-fkms-v3d/d' /boot/config.txt
```
ecdye, line 475: What exactly is this line doing, and why is it now needed?

functions/system.bash
```diff
@@ -471,6 +471,26 @@ memory_split() {
...
+}
+
+## disable or enable framebuffer to provide the maximum amount of memory for Linux operations.
+##
+##    use_framebuffer()
+##
+use_framebuffer() {
+  if ! is_pi; then return 0; fi
+  # TODO
```
ecdye, line 484: Remove todo

functions/system.bash
```diff
@@ -471,6 +471,26 @@ memory_split() {
...
+
+  sed -i '/^dtoverlay=vc4-fkms-v3d/d' /boot/config.txt
+}
+
+## disable or enable framebuffer to provide the maximum amount of memory for Linux operations.
+##
+##    use_framebuffer()
+##
+use_framebuffer() {
+  if ! is_pi; then return 0; fi
```
ecdye, line 483: Probably should make a clean exit statement if interactive on a non RPi system so the user knows why they just got booted back to the menu instead of where they expected

functions/system.bash
```diff
@@ -471,6 +471,26 @@ memory_split() {
...
+  # TODO
+  # menu option
+  sed -i '^[[:space:]]*max_framebuffers' /boot/config.txt
```
ecdye, line 486: What is this sed statement doing? all it looks like is it is finding the line, not performing any operations on it right? Don't you want to delete this line instead so that you don't end up with a whole bunch of duplicated on and off switches of this option?

ecdye (changes requested):
Looks pretty good, not sure exactly what this is doing though. Is it just allowing a user to use the RPi to access the console directly from HDMI instead of ssh?

functions/menu.bash
mstormi: done

mstormi:
> Is it just allowing a user to use the RPi to access the console directly from HDMI instead of ssh?

Yes it switches the HDMI console on or off. It's on per default which is costing a little memory so users can disable it now to make use of that mem.
I kept the default to HDMI on though as it's clearly more important to get the box installed in the first place, and if that requires a console for whatever reason it should not be disabled.

functions/menu.bash
mstormi: I just renumbered options but wanted to keep zram 38 so I don't have to change docs that may reference this.

functions/system.bash
mstormi: It's some graphics option for Pi4 only that appeared as part of Raspi OS, but it gets deleted now ("/d") because it's *not* needed .

functions/menu.bash
mstormi: wanna compete against shellcheck ? ;-)

functions/system.bash

functions/system.bash
mstormi: ok

functions/system.bash
mstormi: Need this line to delete any occurences. The next lines then re-add it.
well spotted, fixed

functions/menu.bash
ecdye: Nope, just an addition to shellcheck :)

ecdye (approved):
LGTM,DNT

PR #1585, opened 2021-09-25
Partial config restore option

mstormi:

### DIFF
--- a/functions/backup.bash
+++ b/functions/backup.bash
@@ -48,6 +48,7 @@ restore_openhab_config() {
   local backupPath="${OPENHAB_BACKUPS:-/var/lib/openhab/backups}"
   local filePath
   local fileSelect
+  local storageText="Only text input files *.things *.items *.rules will be restored from the backup zipfile.\\nConfig elements that were defined using the GUI will not be restored."
 
   echo -n "$(timestamp) [openHABian] Choosing openHAB backup zipfile to restore from... "
   if [[ -n "$INTERACTIVE" ]]; then
@@ -68,7 +69,10 @@ restore_openhab_config() {
 
   echo -n "$(timestamp) [openHABian] Restoring openHAB backup... "
   if [[ "$1" == "textonly" ]]; then
-    if ! (yes | cond_redirect "${OPENHAB_RUNTIME:-/usr/share/openhab/runtime}/bin/restore --textonly $filePath"); then echo "FAILED (restore)"; return 1; fi
+    if [[ -n "$INTERACTIVE" ]]; then
+      if ! (whiptail --title "Restore text config only" --yes-button "Continue" --no-button "Cancel" --yesno "$storageText" 10 80); then echo "CANCELED"; return 0; fi
+    fi
+    if ! (yes | cond_redirect "${OPENHAB_RUNTIME:-/usr/share/openhab/runtime}"/bin/restore --textonly "$filePath"); then echo "FAILED (restore)"; return 1; fi
   else
     if ! cond_redirect systemctl stop openhab.service; then echo "FAILED (stop openHAB)"; return 1; fi
     if ! (yes | cond_redirect openhab-cli restore "$filePath"); then echo "FAILED (restore)"; return 1; fi
--- a/functions/backup.bash
+++ b/functions/backup.bash
@@ -50,7 +50,7 @@ restore_openhab_config() {
   local fileSelect
   local introText="This will restore a backup of your openHAB configuration using openHAB's builtin backup tool.\\n\\nWould you like to continue?"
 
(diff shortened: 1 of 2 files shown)


functions/backup.bash
```diff
@@ -48,9 +48,8 @@ restore_openhab_config() {
   local backupPath="${OPENHAB_BACKUPS:-/var/lib/openhab/backups}"
   local filePath
   local fileSelect
-  local introText="This will restore a backup of your openHAB configuration using openHAB's builtin backup tool.\\n\\nWould you like to continue?"
```
ecdye, line 51: Why remove the intro text and first confirmation question, I think it is a good thing to have as if a user is not careful they could overwrite their config unintentionally.

functions/menu.bash
```diff
@@ -217,6 +218,7 @@ show_main_menu() {
     case "$choice2" in
       50\ *) backup_openhab_config ;;
       51\ *) restore_openhab_config ;;
+      *Restore\ text\ only*) restore_openhab_config "--textonly" ;;
```
ecdye, line 221: Why use GNU long style options here? Its not how we usually do it so I would prefer if we only kept the actual flag when we call the command.

functions/menu.bash
```diff
@@ -206,7 +206,8 @@ show_main_menu() {
   elif [[ "$choice" == "50"* ]]; then
     choice2=$(whiptail --title "openHABian Configuration Tool — $(get_git_revision)" --menu "Backup/Restore" 14 116 7 --cancel-button Back --ok-button Execute \
     "50 | Backup openHAB config"      "Backup the current active openHAB configuration" \
-    "51 | Restore an openHAB config"  "Restore a previous openHAB configuration from backup" \
+    "51 | Restore an openHAB config"  "Restore an openHAB configuration from backup zipfile" \
+    "   | Restore text only config"   "Restore text only configuration without restarting" \
```
ecdye, line 210: Perhaps we should add a prompt to more fully explain what a text only restore really is for users as for a new user they may not understand what that means and it may not be what they want. Also does openhab support creating text only backups, and if so do we want to as well?

functions/backup.bash
```diff
@@ -69,9 +67,13 @@ restore_openhab_config() {
   fi
 
   echo -n "$(timestamp) [openHABian] Restoring openHAB backup... "
-  if ! cond_redirect systemctl stop openhab.service; then echo "FAILED (stop openHAB)"; return 1; fi
-  if ! (yes | cond_redirect openhab-cli restore "$filePath"); then echo "FAILED (restore)"; return 1; fi
-  if cond_redirect systemctl restart openhab.service; then echo "OK"; else echo "FAILED (restart openHAB)"; return 1; fi
+  if [[ "$1" == "--textonly" ]]; then
+    if ! (yes | cond_redirect "${OPENHAB_RUNTIME:-/usr/share/openhab/runtime}/bin/restore --textonly $filePath"); then echo "FAILED (restore)"; return 1; fi
```
ecdye, line 71: Can this not be called from the openhab-cli? If it can I would much prefer to use that then a hard link.

Also the quoting on this line is a little hard to read might be better to go with:
```suggestion
    if ! (yes | cond_redirect "${OPENHAB_RUNTIME:-/usr/share/openhab/runtime}"/bin/restore --textonly "$filePath"); then echo "FAILED (restore)"; return 1; fi
```

ecdye (changes requested):
Mostly some questions and a couple of suggestions on this one but it looks good.

functions/backup.bash
mstormi: They need to select the file to restore from anyway (can cancel there of course, too) so it was a quite unnecessary extra input.

functions/menu.bash
mstormi: it was because that's what gets passed to the binary.
But you're right in terms of clean coding that's a fail so

functions/backup.bash
mstormi: openhab-cli would enforce openHAB being taken down but that's a major point here: importing text only allows for keeping OH running.

functions/backup.bash
ecdye: OK, I'm ok with it.

functions/backup.bash
ecdye: Ahh, that makes sense. thanks for clarifying.

functions/menu.bash
mstormi: ok I added a prompt.
It actually only extracts the text config files no matter if there's other files in the zipfile as well so backup creation is unaffected.

ecdye (approved):
LGTM, DNT

Issue #1561, opened 2021-08-03
Refactor permissions_corrections()


mstormi:
We should intro is_installed_XXX() functions for all 3rd party options we offer as part of openHABian.

Then refactor permissions_corrections() in system.bash using those functions to order / structure the code.
It gets more and more convoluted, increasing likelihood to introduce logical mistakes.

Also add permissions_corrections() to all 3rd party BATS tests else packages might be mis-installed hence mis-tested

- [x] intro is_installed_XXX() for all 3rd party
- [x] refactor permissions_corrections to only use these new functions
- [ ] add permissions_corrections() to all 3rd party BATS tests

mstormi:
@ecdye Renamed it just for you to enjoy. Does that motivate you to work this out ?

ecdye:
I can certainly work on it, I have just been busy lately I will start on this, I will probably do a PR for each part just to simplify the process.

mstormi:
Thanks, I'm currently hitting many todos - small ones I fix right away but also some that require more work - so happy for any load you can take off me. Will eventually document some more of these issues.

mstormi:
On a sidenote I wanted to make you aware that I've just forwarded main to openHAB3 branch.
Hope I didn't break anything this way.

That has sort of been overdue anyway and should also fix [this](https://community.openhab.org/t/openhabian-hassle-free-openhab-setup/13379/1494) interesting issue. I believe installs have been failing on non RPi.
Eventually the "wrong architecture java binary" issues I've seen occasionally might also have been caused by this.

```
openhabian@devpi:~/openhabian $ git checkout main
Zu Branch 'main' gewechselt
Ihr Branch ist auf demselben Stand wie 'origin/main'.
openhabian@devpi:~/openhabian $ git remote -v
myrepo  https://github.com/mstormi/openhabian.git (fetch)
myrepo  https://github.com/mstormi/openhabian.git (push)
origin  https://github.com/openhab/openhabian.git (fetch)
origin  https://github.com/openhab/openhabian.git (push)
openhabian@devpi:~/openhabian $ git push origin openHAB3
Gesamt 0 (Delta 0), Wiederverwendet 0 (Delta 0), Pack wiederverwendet 0
remote:
remote: Heads up! The branch 'openHAB3' that you pushed to was renamed to 'backup.openHAB3'.
remote:
To https://github.com/openhab/openhabian.git
   75187cd..3cdb213  openHAB3 -> openHAB3
[...]

PR #1586, opened 2021-09-25
Introduce and transition to `XXX_is_installed`

ecdye:
Related #1561

### DIFF
--- a/functions/helpers.bash
+++ b/functions/helpers.bash
@@ -558,11 +558,148 @@ set_cpu_speed() {
 }
 
 
-## check if InfluxDB is installed
+## Functions to check it a given supported program is installed should be placed below here
+
+## Check if amanda is installed
 ##
-##    iinfluxdb_is_installed
+##    amanda_is_installed
+##
+amanda_is_installed() {
+  if dpkg -s 'amanda-common' 'amanda-server' 'amanda-client' &> /dev/null; then return 0; fi
+  return 1
+}
+
+## Check if InfluxDB is installed
+##
+##    influxdb_is_installed
 ##
 influxdb_is_installed() {
-  if [[ -d /var/lib/influxdb ]]; then return 0; fi
+  if dpkg -s 'influxdb' &> /dev/null && [[ -d /var/lib/influxdb ]]; then return 0; fi
+  return 1
+}
+
+## Check if Grafana is installed
+##
+##    grafana_is_installed
+##
+grafana_is_installed() {
+  if dpkg -s 'grafana' &> /dev/null && [[ -s /etc/grafana/grafana.ini ]]; then return 0; fi
+  return 1
+}
+
+## Check if node is installed
+##
+##    node_is_installed
+##
+node_is_installed() {
+  if [[ -x $(command -v npm) ]] || [[ $(node --version) == "v14"* ]]; then return 0; fi
+  return 1
+}
+
+## Check if samba is installed
+##
+##    samba_is_installed
+##
+samba_is_installed() {
+  if dpkg -s 'samba' &> /dev/null && [[ -s /etc/samba/smb.conf ]]; then return 0; fi
+  return 1
+}
+
+## Check if firemotd is installed
+##
+##    firemotd_is_installed
+##
+firemotd_is_installed() {
+  if [[ -d /opt/firemotd ]] && [[ -x $(command -v FireMotD) ]]; then return 0; fi
+  return 1
+}
+
+## Check if exim is installed
+##
+##    exim_is_installed
+##
+exim_is_installed() {
+  if dpkg -s 'mailutils' 'exim4' &> /dev/null; then return 0; fi
+  return 1
+}
+
+## Check if homegear is installed
+##
+##    homegear_is_installed
+##
+homegear_is_installed() {
+  if dpkg -s 'homegear' &> /dev/null && [[ -d /var/lib/homegear ]]; then return 0; fi
+  return 1
+}
+
+## Check if mqtt is installed
+##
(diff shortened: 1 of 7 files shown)


ecdye:
This should contain the important ones, let me know if you want any added, also this should complete tasks 1&2 of #1561

functions/helpers.bash
```diff
@@ -558,11 +558,148 @@ set_cpu_speed() {
...
+homegear_is_installed() {
+  if dpkg -s 'homegear' &> /dev/null && [[ -d /var/lib/homegear ]]; then return 0; fi
+  return 1
+}
+
+## Check if mqtt is installed
+##
+##    mqtt_is_installed
+##
+mqtt_is_installed() {
```
mstormi, line 639: shouldn't that be mosquitto rather than mqtt

functions/helpers.bash
```diff
@@ -558,11 +558,148 @@ set_cpu_speed() {
...
+  if dpkg -s 'grafana' &> /dev/null && [[ -s /etc/grafana/grafana.ini ]]; then return 0; fi
+  return 1
+}
+
+## Check if node is installed
+##
+##    node_is_installed
+##
+node_is_installed() {
+  if [[ -x $(command -v npm) ]] || [[ $(node --version) == "v14"* ]]; then return 0; fi
```
mstormi, line 595: is it still correct to test for v14 ? I recall there were issues without but I guess we should at least query for a version equal *or greater* 14

functions/packages.bash
```diff
@@ -277,7 +277,7 @@ mqtt_setup() {
     echo "OK"
   fi
 
-  if ! dpkg -s 'mosquitto' 'mosquitto-clients' &> /dev/null; then
+  if ! mqtt_is_installed; then
```
mstormi, line 280: mqtt->mosquitto

functions/packages.bash
```diff
@@ -301,7 +301,7 @@ mqtt_setup() {
     if cond_redirect rm -f /etc/mosquitto/passwd; then echo "OK"; else echo "FAILED"; return 1; fi
   fi
   if ! zram_dependency install mosquitto; then return 1; fi
-  if [[ -s /etc/ztab ]] && ! mkdir -p /opt/zram/log.bind/mosquitto /var/log/mosquitto && chown mosquitto /opt/zram/log.bind/mosquitto /var/log/mosquitto; then echo "FAILED (create zram logdir)"; return 1; fi
+  if zram_is_installed && ! mkdir -p /opt/zram/log.bind/mosquitto /var/log/mosquitto && chown mosquitto /opt/zram/log.bind/mosquitto /var/log/mosquitto; then echo "FAILED (create zram logdir)"; return 1; fi
   echo -n "$(timestamp) [openHABian] Setting up MQTT service... "
```
mstormi, line 305: "mosquitto MQTT broker service" or the like

functions/system.bash
```diff
@@ -324,37 +324,36 @@ permissions_corrections() {
...
-      if ! cond_redirect fix_permissions /opt/zram/log.bind/mosquitto "mosquitto:${username:-openhabian}" 644 755; then echo "FAILED (mosquitto log permissions on zram)"; retval=1; fi
-    fi
-  fi
   if ! cond_redirect setfacl --recursive --remove-all "${openhabFolders[@]}"; then echo "FAILED (reset file access lists)"; retval=1; fi
 
   if ! cond_redirect fix_permissions /var/log/unattended-upgrades root:root 644 755; then echo "FAILED (unattended upgrades logdir)"; retval=1; fi
   if ! cond_redirect fix_permissions /var/log/samba root:root 640 750; then echo "FAILED (samba logdir)"; retval=1; fi
   if ! cond_redirect fix_permissions /var/log/openhab "openhab:${username:-openhabian}" 664 775; then echo "FAILED (openhab log)"; retval=1; fi
 
+  if mqtt_is_installed; then
```
mstormi, line 333: mqtt->mosquitto

mstormi (changes requested):
essentially LGTM just some minor naming

functions/helpers.bash
ecdye: I was lazy and didn't want to type it all the way out, plus the install function uses mqtt as well, I can change it if you want

functions/helpers.bash
ecdye: Yes it still works, we currently use node 14 as our supposed baseline, I moved it from 12 about a year ago. I don't think we want to do greater than or equal to 14 because there are breaking changes with some of the packages we use between major revisions of the node version we use. It also makes sure that we don't screw someones version up by checking for nom being installed, because if it is we don't even look at the node version.

That whole string was actually taken from the install function to begin with so I think it should be fine.

functions/packages.bash
ecdye: I'm confused what your referring to here, do you want me to change the text on line 305? if so that should probably be a separate PR as it is very unrelated.

mstormi: yes line 305
you can make it another PR but in here it's fine, too

functions/helpers.bash
mstormi: yes please

ecdye: fixed

functions/packages.bash
ecdye: fixed

functions/packages.bash
ecdye: fixed

functions/system.bash
ecdye: fixed

mstormi (approved):
LGTM

Issue #1590, opened 2021-09-28
Add forward proxy and proxyconfig


mstormi:
If installed, nginx in openhabian can also provide a *forward* proxy.
This can be used to access devices in a remote location via openHABian box serving as a proxy in that location.
Users would need to manually configure this a the proxy in his browser.
As an enhancement, we maybe could build a browser auto proxy config that
considers the remote network (to be extracted from the network interface config?) as destinations it would enforce proxying for while anything else will be accessed directly.
Need to place proxy autoconfig as a file `/etc/openhab/html/proxyconf.pac` which is served as `http://\<openhabian>/static/proxyconf`. Take care it needs to be served with a specific mime type `application/x-ns-proxy-autoconfig`.

- [x] nginx config enhancement see #1591
- [ ] proxy autoconfig file

mstormi:
On 2nd thought the proxyconfig is a little over the top.
Will hence close this for now, maybe some day when we're out of work left

PR #1591, opened 2021-09-28
Add nginx forward proxy

mstormi:
Related: #1590

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,12 @@ Hit tab to unselect buttons and scroll through the text using UP/DOWN or
 PGUP/PGDN. All announcements are stored in `/opt/openhabian/docs/CHANGELOG.md`
 for you to lookup.
 
+## Forward web proxy ## September 28, 2021
+Installing nginx (menu option 44) will now provide a *forward* proxy on port
+8888 in addition to the reverse proxy setup it has been providing.
+Configure this as a manual proxy in your browser to access devices in remote
+(VPN) locations by their IP.
+
 ## Install openHAB function changed ## July 23, 2021
 Menu option 03 "Install or upgrade to openHAB 3" has been changed to now
 actually do what it claims it will do: only install or upgrade to openHAB 3, it
--- a/NEWS.md
+++ b/NEWS.md
@@ -23,9 +23,3 @@ difficult to maintain as a result of it requiring packages that are no longer
 provided by the Debian distribution used in openHABian. If you would like to
 install it on your own please see [this](https://community.openhab.org/t/89856)
 forum thread for some guidance.
-
-## deCONZ / Phoscon companion app added ## May 10, 2021
-There's a new menu option to install the deCONZ software / Phoscon companion app
-to support the popular Dresden Elektronik Conbee and Raspbee ZigBee controllers.
-Note you will need to use the [deconz binding](https://www.openhab.org/addons/bindings/deconz/)
-and pair your devices using the Phoscon web interface running on port 8081.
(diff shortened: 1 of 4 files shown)


NEWS.md
```diff
@@ -2,6 +2,12 @@ Hit tab to unselect buttons and scroll through the text using UP/DOWN or
 PGUP/PGDN. All announcements are stored in `/opt/openhabian/docs/CHANGELOG.md`
 for you to lookup.
 
+## Forward web proxy ## September 28, 2021
```
ecdye, line 5: Make sure to only keep the 3 most recent announcements in here. We keep all announcements in `doc/CHANGELOG.md` so make sure to add it there.

functions/menu.bash
```diff
@@ -183,7 +183,7 @@ show_main_menu() {
     "42 | Upgrade to openHAB 3"   "Upgrade OS environment to openHAB 3 release" \
     "   | Downgrade to openHAB 2" "Downgrade OS environment from openHAB 3 back to openHAB 2 stable (DANGEROUS)" \
     "43 | Remote Console"         "Bind the openHAB SSH console to all external interfaces" \
-    "44 | Reverse Proxy"          "Setup Nginx with password authentication and/or HTTPS access" \
+    "44 | Nginx Proxy"            "Setup reverse and forward web proxy" \
```
ecdye, line 186: I actually really like this change in wording I never could figure out a good way to word it but I think this works well.

includes/nginx.conf
```diff
@@ -2,6 +2,21 @@
 # openHABian NGINX Configuration #
 ##################################
 
+
+## Forward proxy
+## Configure <server>:8888 as a manual (forward) proxy in your browser
+## Access any local device in a *remote* (Tailscale VPN) location by target IP
+
```
ecdye, line 9: Is this the only change we need to make other than what you mentioned in task 2 on the issue? Also please tell me this doesn't allow remote access to any device on the RPi's LAN as that would be a big security issue.

ecdye (changes requested):
Looks good just move around the announcement and also, do we want to announce it yet as don't we have to do task 2 in the original issue to finish up.

includes/nginx.conf
mstormi: Yes it's the only change.
Proxy is on port 8888 which of course noone should publish/open to the outside.
Yes it allows for access to any device - but that is just like you can access these when you're in the LAN.
So I don't see the security issue in here ?
Maybe we should restrict it to *web* interface, but thing is it would need to be ports 80 AND 443 and I don't know how to configure that right now other than to allow on all ports.  Also, nginx is not installed by default.

functions/menu.bash
mstormi: thanks. yeah I have been thinking the same in a long time.

NEWS.md
mstormi: ok

mstormi:
> do we want to announce it yet as don't we have to do task 2 in the original issue to finish up

Yes I think so. I don't know if we ever make it to include task 2, too, and even if so we can change the announcement.

includes/nginx.conf
ecdye: Fair enough, I just see the potential for malicious actors to find a way into a LAN if given the chance and from there they could potentially do some nasty stuff.

ecdye (changes requested):
Just add the new announcement to `docs/CHANGELOG.md` and we'll be good to go.

ecdye (approved):

PR #1592, opened 2021-09-29
Import/download config from menu

mstormi:

### DIFF
--- a/functions/backup.bash
+++ b/functions/backup.bash
@@ -53,7 +53,11 @@ restore_openhab_config() {
   echo -n "$(timestamp) [openHABian] Choosing openHAB backup zipfile to restore from... "
   if [[ -n "$INTERACTIVE" ]]; then
     if [[ $# -gt 0 ]]; then
-      if ! [[ -s "$1" ]]; then echo "FAILED (restore config $1)"; return 1; fi
+      if ! [[ -s "$1" ]]; then
+        echo "FAILED (restore config $1)"
+        whiptail --title "Restore failed" --msgbox "Restoration of selected openHAB configuration failed." 7 80
+        return 1
+      fi
       filePath="$1"
       echo "OK"
     else
--- a/functions/backup.bash
+++ b/functions/backup.bash
@@ -52,23 +52,25 @@ restore_openhab_config() {
 
   echo -n "$(timestamp) [openHABian] Choosing openHAB backup zipfile to restore from... "
   if [[ -n "$INTERACTIVE" ]]; then
-    # shellcheck disable=SC2012
-    readarray -t backupList < <(ls -alh "${backupPath}"/openhab*-backup-* 2> /dev/null | head -20 | awk -F ' ' '{ print $9 " " $5 }' | xargs -d '\n' -L1 basename | awk -F ' ' '{ print $1 "\n" $1 " " $2 }')
-    if [[ -z "${backupList[*]}" ]]; then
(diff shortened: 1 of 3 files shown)


functions/menu.bash
```diff
@@ -222,7 +226,7 @@ show_main_menu() {
     case "$choice2" in
       50\ *) backup_openhab_config ;;
       51\ *) restore_openhab_config ;;
-      *Restore\ text\ only*) restore_openhab_config "textonly" ;;
+      *Restore\ text\ only*) restore_openhab_config "${initialConfig:-/boot/initial.zip}" "textonly" ;;
```
ecdye, line 229: Can this file be changed by the user? Don't we want them to be able to choose a place to restore a text only config from if its not on the first boot/just after initial setup.

ecdye (changes requested):
This seems off somehow. I feel like you're missing something on the text config import, but I can't figure out what. Can you lend some brain power and double check that the text only import logic is really the way we want it/a user would expect it to be. IMHO the logic between the two should be exactly the same just one only imports text based backups and the other does a full restore.

functions/backup.bash
```diff
@@ -52,23 +52,25 @@ restore_openhab_config() {
...
   echo -n "$(timestamp) [openHABian] Choosing openHAB backup zipfile to restore from... "
   if [[ -n "$INTERACTIVE" ]]; then
-    # shellcheck disable=SC2012
-    readarray -t backupList < <(ls -alh "${backupPath}"/openhab*-backup-* 2> /dev/null | head -20 | awk -F ' ' '{ print $9 " " $5 }' | xargs -d '\n' -L1 basename | awk -F ' ' '{ print $1 "\n" $1 " " $2 }')
-    if [[ -z "${backupList[*]}" ]]; then
-      whiptail --title "Could not find backup!" --msgbox "We could not find any configuration backup file in the storage directory $backupPath" 8 80
-      echo "CANCELED"
-      return 0
+    if [[ $# -gt 0 ]]; then
+      if ! [[ -s "$1" ]]; then echo "FAILED (restore config $1)"; return 1; fi
```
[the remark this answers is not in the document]
mstormi: added whiptail feedback
Not sure about your feedback regarding textonly: this PR did not change anything about it?
The intended logic is that *if* a user knows what he is to import is text only then he can choose the secondary (recently added) menu option that calls the restore script with --textonly because that option will NOT restart OH.
User can also select the primary option then it gets restarted nonetheless (as is needed when some non-text config is like jsondb and UI's json are imported). Note this is independent of what is contained in the zipfile to restore.
I'd hope this logic is what I implemented, please validate.

functions/menu.bash
mstormi: Yes user can change the file but if the zipfile he wants to restore is elsewhere he can instead change the initialconfig setting in `openhabian.conf` to point there.
Plus if interactive this is just the proposed default the restore routine proposes when it makes a whiptail query.

ecdye: But if I read the logic correctly, for the textonly config we would never give the user a chance to select the file to restore from as a whiptail query or am I just reading the code wrong?

mstormi: So I hope but could be wrong as well.
textonly is now $2 ($1 is a new argument), in import_openhab_config() $1 is evaluated first (passed as proposal to whiptail)

ecdye (approved):
LGTM, DNT

PR #1596, opened 2021-10-04
Fix openHAB 2 install from new images

### LINKED ISSUE #1594: Default install OH2
After getting word that installing with clonebranch=stable does NOT install OH2 but 3, I'd think this might be because the image is based on the main branch and `/opt/openhabian/openhabian-setup.sh` actually is being executed BEFORE it is being overwritten with the `stable` code.
Fix this by evaluating if to install OH2 or OH3 then pass this as $1 of the `openhab_setup` call in openhabian-setup.sh

ecdye:
Fixes #1594

### DIFF
--- a/.github/workflows/build-action.yml
+++ b/.github/workflows/build-action.yml
@@ -6,12 +6,14 @@ on:
     paths:
       - 'build.bash'
       - 'build-image/openhabian-installer.service'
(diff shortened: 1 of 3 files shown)


mstormi (dismissed):
good find

mstormi (approved):

Issue #1597, opened 2021-10-05
Frontail install failing


mstormi:
Some reports on the forum.

Reproduceable with a fresh install:

```
+ cd /opt/frontail
+ cond_redirect npm install --force -g
+ [[ -n '' ]]
+ echo -e '\n\033[90;01m$ npm install --force -g \033[39;49;00m'

$ npm install --force -g
+ npm install --force -g
/opt/openhabian/functions/helpers.bash: Zeile 31: npm: Kommando nicht gefunden.
+ return 127
+ echo 'FAILED (install)'
FAILED (install)
+ return 1
+ '[' 1 -ne 0 ']'
+ whiptail --msgbox 'There was an error or interruption during the execution of:\n  "20 | Optional Components"\n\nPlease try again. If the error persists, please read /opt/openhabian/docs/openhabian-DEBUG.md or https://github.com/openhab/openhabian/blob/main/docs/openhabian-DEBUG.md how to proceed.' 14 80
+ return 0
```

mstormi:
Frankly though I don't understand why npm all of a sudden got missing.
Maybe some debian package dependency rework ?

mstormi:
@ecdye fyi https://community.openhab.org/t/openhabian-frontail-install-error/39228/6
then again I guess that reinstalls npm just we do.

PR #1598, opened 2021-10-05
Install npm to fix frontail install

mstormi:
Fixes: #1597 

### DIFF
--- a/functions/nodejs-apps.bash
+++ b/functions/nodejs-apps.bash
@@ -6,10 +6,7 @@
 ##    nodejs_setup()
 ##
 nodejs_setup() {
-  if node_is_installed ; then return 0; fi
-  if ! is_armv6l; then
-    if ! cond_redirect apt-get install --yes npm; then echo "FAILED (npm install)"; return 1; fi
-  fi
+  if node_is_installed || ! is_armv6l; then return 0; fi
 
   local keyName="nodejs"
   local link="https://unofficial-builds.nodejs.org/download/release/v14.17.5/node-v14.17.5-linux-armv6l.tar.xz"
@@ -44,9 +41,9 @@ nodejs_setup() {
 
     echo -n "$(timestamp) [openHABian] Installing NodeJS... "
     if [[ -n $PREOFFLINE ]]; then
-      if cond_redirect apt-get --quiet install --download-only --yes nodejs; then echo "OK"; else echo "FAILED"; return 1; fi
+      if cond_redirect apt-get --quiet install --download-only --yes nodejs npm; then echo "OK"; else echo "FAILED"; return 1; fi
     else
-      if cond_redirect apt-get install --yes nodejs; then echo "OK"; else echo "FAILED"; return 1; fi
+      if cond_redirect apt-get install --yes nodejs npm; then echo "OK"; else echo "FAILED"; return 1; fi
     fi
   fi
 }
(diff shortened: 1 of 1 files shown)


ecdye (changes requested):
Not sure why this broke all of the sudden but I would rather just install `npm` at the same time as `nodejs` so why dot you move this to lines 47 & 49.

mstormi:
> 
> 
> Not sure why this broke all of the sudden but I would rather just install `npm` at the same time as `nodejs` so why dot you move this to lines 47 & 49.

See original code, it wasn't executed when either NodeJS was already installed (which it was in Raspi OS)
OR when it was NOT a RPi0. So effectively it never was.
I don't wanna break that logic by moving this down.

if node_is_installed || ! is_armv6l; then return 0; fi

ecdye:
> > Not sure why this broke all of the sudden but I would rather just install `npm` at the same time as `nodejs` so why dot you move this to lines 47 & 49.
> 
> See original code, it wasn't executed when either NodeJS was already installed (which it was in Raspi OS) OR when it was NOT a RPi0. So effectively it never was. I don't wanna break that logic by moving this down.
> 
> if node_is_installed || ! is_armv6l; then return 0; fi

But this new code has the same problem does it not? If NodeJS is installed it won't execute and it nodeJS is not installed it and we are not on a RPi0 we will install npm before we install NodeJS, right? That is what I don't want if anything we should install both at the same time.

mstormi:
Yes it has the same behavior. But that has not been a problem so far so why do you call it a problem? And then why change?
And I don't get why you insist on installing both at the same time. Why isn't the current proposal code (doing npm first nodejs later) fine ?

I don't mind changing but I think I didn't get your point so as you can edit, I suggest you do to suit your idea.

ecdye:
> Yes it has the same behavior. But that has not been a problem so far so why do you call it a problem? And then why change? And I don't get why you insist on installing both at the same time. Why isn't the current proposal code (doing npm first nodejs later) fine ?
> 
> I don't mind changing but I think I didn't get your point so as you can edit, I suggest you do to suit your idea.

I just think that it is much cleaner and more effective to install them both at the same time rather than doing one then the other. It also reduces the amount of code overall.

ecdye (approved):

mstormi:
did you test it works ? nodejs_is_installed at the beginning checks for node OR npm so if node is there but npm is not it still won't install, will it?
That's why I took the apparently longer route I did.

ecdye:
Yes it appears to work for me, but I can change it to ensure that if one or the other is not there to just install both again.

mstormi:
I'm still unsure we talk about the right point so again, to be explicit about the important point: did you test with a fresh image? 
if at the early stage in automated installation where we have nothing but Raspi OS, the first line in nodejs_setup() calls node_is_installed() and returns(!) when node_is_installed() returns true which is when EITHER node OR npm is installed.
So if node is there but npm is not, it won't get installed, leading to the frontail error later. Or am I wrong here ?

(and note Raspi OS can also change at any time in the future, too, to have either of these packages installed or not, and I want a permanent, future proof solution to work independent of what's in the image or not)

ecdye:
Yes I did test and it worked on a fresh image. I'm going to make the change so that both npm and nodeJS must be installed to let the check pass. if one or the other is missing it will reinstall once I make the change.

ecdye:
OK so really what needs to happen is when npm or nodejs get mixed up and are not both installed just reinstall nodejs let me issue the fix.

ecdye:
@WolfgangSn try again now.

PR #1603, opened 2021-10-10
Add support for Java 17

### LINKED ISSUE #1602: Java 17
As Java 17 is now out as the current long term support version and likely the one openHAB will move to soon(ish), we should include this in openHABian asap.
Open for ideas how to do it best, probably as another alternative (like Java 8 and Adopt are today) with the default still being Java 11. @ecdye Ethan I hope it's okay I assign that to you.

ecdye:
See commit descriptions for details on some of the decisions made in this PR.

Fixes #1602 

### DIFF
--- a/.github/workflows/installation-action.yml
+++ b/.github/workflows/installation-action.yml
@@ -174,7 +174,7 @@ jobs:
   #          push: true
   #      - name: openHABian installation test with 32bit Azul Java 17 on a virtual RPi3
   #        run: |
-  #          docker run --name "openhabian-rpi3" -d localhost:5000/openhabian/rpi3-openhabian
+  #          docker run --privileged --rm --name "openhabian-rpi3" -d localhost:5000/openhabian/rpi3-openhabian
   #          docker exec -i "openhabian-rpi3" bash -c './build.bash local-test && /boot/first-boot.bash'
   #        shell: bash
   rpi3-zulu17-64:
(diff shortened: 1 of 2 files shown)


ecdye:
@mstormi This is more or less ready now in terms of Azul. The problem is that Azul and as far as I can tell, no other providers are building Java 17 with arm32 support. We can move forward and merge this but it will only work on 64bit installations. 

As for AdoptOpenJDK they are undergoing a transition to being owned by eclipse and have not created the apt repos under their new brand name (Adoptium) so for the moment I will leave that code as is until they have made an apt package repo under their new name.

Thoughts?

mstormi:
Hmm ok. I think we should offer it, with a popup maybe upfront to have a statement that Java 17 is not (yet) being supported by openHAB but that we provide this feature to get feedback as early as possible.
On 64 bit, am I right that it would run on 32 bit Raspi OS just with the drawback of additional mem usage ?
Then we should include that in the popup statement.
As for Adopt yeah let's wait for them to settle before you put work in there.
Thanks for pushing this.

ecdye:
No, 64bit will not run on 32bit system at all, which may actually push openHAB's development decisions to refrain from upgrading past Java 11. @openhab/architecture-council What are your thoughts on the Java 17 decision to only support 64 bit officially (meaning no-one in their right mind is building/supporting 32bit Java 17).

wborn:
Similar to how we went to Java 11, the first step is to support running openHAB with both Java 11 and Java 17. After that we can decide to also compile everything for Java 17. It's probably also possible to use Java 17 in source code but compile to Java 11.

> What are your thoughts on the Java 17 decision to only support 64 bit 

Do you have a link to this decision or is it just that they haven't bothered setting up and fixing 32-bit builds? As you may have noticed, there is a 32-bit JDK 17 for Windows: https://www.azul.com/downloads/?version=java-17-lts&architecture=x86-32-bit&package=jdk

Maybe creating builds for these older 32-bit Linux platforms is still in their backlog and doesn't have as a high priority. Most applications still need to be updated (like openHAB) and then you start with testing/running them with Java 17 on your 64-bit laptop/desktop. :wink:

PR #1617, opened 2021-11-11
Improve CI-Pipeline and other related changes

ecdye:
This primarily removes the need for the systemctl replacement in docker
by fully utilizing the initsystem provided by Balena in their docker
images. This system does not work for BATS tests in GitHub Actions
currently. However, the initsystem does work for BATS on my machine in
Docker but I cannot get it to work in the online GitHub Actions
environment. 

This also includes a bunch of bug fixes that were found along the way.


.github/workflows/bats-action.yml
```diff
@@ -64,14 +64,17 @@ jobs:
...
-           docker run --name "openhabian-rpi4" -d localhost:5000/openhabian/rpi4-openhabian
+           docker run --rm --name "openhabian-rpi4" -d localhost:5000/openhabian/rpi4-openhabian
            docker exec -i "openhabian-rpi4" bash -c 'bats --tap --recursive --filter "development-." .'
            docker exec -i "openhabian-rpi4" bash -c 'bats --tap --recursive --filter "unit-." .'
            docker exec -i "openhabian-rpi4" bash -c 'bats --tap --recursive --filter "installation-." .'
            docker exec -i "openhabian-rpi4" bash -c 'bats --tap --recursive --filter "destructive-." .'
+           docker stop "openhabian-rpi4"
          shell: bash
   amd64-bats:
     runs-on: ubuntu-latest
```
mstormi, line 80: So if I correctly understood you replace the Ubuntu CI run by a generic Debian bullseye one.
Would you also need to replace the "ubuntu" in "runs-on" ?

build.bash
```diff
@@ -258,11 +258,11 @@ elif [ "$1" == "local-test" ]; then
   # Use local filesystem's version of openHABian
   # shellcheck disable=SC2016
   if ! running_in_docker; then
-    sed -i 's|$(eval "$(openhabian_update &> /dev/null)") -eq 0|true|' /boot/first-boot.bash
+    sed -i 's|$(eval "$(openhabian_update "${clonebranch:-openHAB3}" &> /dev/null)") -eq 0|true|' /boot/first-boot.bash
```
mstormi, line 261: Should this not have been in already?

But what does this have to do with CI changes?

functions/find.bash
```diff
@@ -90,7 +90,7 @@ find3_setup() {
   if ! cond_redirect cd $find3Dir/server/ai; then echo "FAILED (cd)"; return 1; fi
   if cond_redirect python3 -m pip install -r requirements.txt; then echo "OK"; else echo "FAILED (pip)"; return 1; fi
 
-  if ! zram_dependency install find3server find3ai; then return 1; fi
+  if ! cond_redirect zram_dependency install find3server find3ai; then return 1; fi
```
mstormi, line 93: same here please don't mix everything into this CI PR

functions/influxdb+grafana.bash
```diff
@@ -212,7 +212,7 @@ influxdb_install() {
   echo -n "$(timestamp) [openHABian] Setting up InfluxDB service... "
   # Disable authentication, to allow changes in existing installations
   if ! cond_redirect sed -i -e 's|auth-enabled = true|# auth-enabled = false|g' /etc/influxdb/influxdb.conf; then echo "FAILED (disable authentication)"; return 1; fi
-  if ! zram_dependency install influxdb; then return 1; fi
+  if ! cond_redirect zram_dependency install influxdb; then return 1; fi
```
mstormi, line 215: more mixes that don't belong here

functions/influxdb+grafana.bash
```diff
@@ -258,18 +258,17 @@ grafana_install(){
...
 
   echo -n "$(timestamp) [openHABian] Setting up Grafana service... "
-  # Workaround for strange behavior in CI
-  if ! cond_redirect mkdir -p /run/grafana; then echo "FAILED (mkdir)"; return 1; fi
-  if ! cond_redirect chmod -R 0750 /run/grafana; then echo "FAILED (chmod)"; return 1; fi
-  if ! cond_redirect chown -R grafana:grafana /run/grafana; then echo "FAILED (chown)"; return 1; fi
-  if ! zram_dependency install grafana-server; then return 1; fi
-  if zram_is_installed && ! mkdir -p /opt/zram/log.bind/grafana /var/log/grafana && chown grafana /opt/zram/log.bind/grafana /var/log/grafana; then echo "FAILED (create zram logdir)"; return 1; fi
-  if ! cond_redirect systemctl -q daemon-reload &> /dev/null; then echo "FAILED (daemon-reload)"; return 1; fi
+  if ! cond_redirect zram_dependency install grafana-server; then return 1; fi
```
mstormi, line 261: more mixes

functions/influxdb+grafana.bash
```diff
@@ -258,18 +258,17 @@ grafana_install(){
...
+    if ! cond_redirect mkdir -p /opt/zram/log.bind/grafana /var/log/grafana; then echo "FAILED (create zram logdir)"; return 1; fi
+    if ! cond_redirect chown grafana /opt/zram/log.bind/grafana /var/log/grafana; then echo "FAILED (create zram logdir)"; return 1; fi
+  fi
+  if ! cond_redirect systemctl -q daemon-reload; then echo "FAILED (daemon-reload)"; return 1; fi
   if cond_redirect systemctl enable --now grafana-server.service; then echo "OK"; else echo "FAILED (enable service)"; return 1; fi
 
   echo -n "$(timestamp) [openHABian] Setting up Grafana... "
   cond_echo "\\nWait for Grafana to start... "
-  if ! (curl -4 --retry 6 --retry-connrefused --silent --head http://localhost:3000); then echo "FAILED (wait for Grafana to start)"; return 1; fi
+  if ! cond_redirect curl -4 --retry 6 --retry-connrefused --insecure --head http://localhost:3000; then echo "FAILED (wait for Grafana to start)"; return 1; fi
```
mstormi, line 271: have not seen before

but also more mixes

functions/influxdb+grafana.bash
```diff
@@ -289,8 +288,8 @@ grafana_install(){
   fi
 
   cond_echo "\\nRestarting Grafana... "
-  if ! cond_redirect systemctl restart grafana-server.service; then echo "FAILED (restart service)"; return 1; fi
-  if (curl -4 -m 120 --retry 20 --retry-connrefused --silent --head http://localhost:3000); then echo "OK"; else echo "FAILED (wait for Grafana to start)"; return 1; fi
+  if ! cond_redirect systemctl restart grafana-server.service; then echo "FAILED (restart service)"; cond_redirect journalctl -xe; return 1; fi
+  if cond_redirect curl -4 --retry 6 --retry-connrefused --insecure --head http://localhost:3000; then echo "OK"; else echo "FAILED (wait for Grafana to start)"; return 1; fi
 }
```
mstormi, line 293: again

functions/java-jre.bash
```diff
@@ -279,7 +279,7 @@ java_zulu_enterprise_apt() {
   if ! add_keys "https://www.azul.com/files/0xB1998361219BD9C9.txt" "$keyName"; then return 1; fi
 
   echo -n "$(timestamp) [openHABian] Adding Zulu repository to apt... "
-  if ! echo "deb [signed-by=/usr/share/keyrings/${keyName}.gpg] http://repos.azulsystems.com/debian stable main" > /etc/apt/sources.list.d/zulu-enterprise.list; then echo "FAILED"; return 1; fi
```
mstormi, line 282: Java changes ? Please unmix

functions/openhab.bash
```diff
@@ -59,25 +59,25 @@ openhab_setup() {
   fi
 
   if [[ $2 == "unstable" ]]; then
```
mstormi, line 61: I recall this was changed before, why is this back ?? Did you maybe diff against some outdated git repo ?

functions/openhab.bash
```diff
@@ -108,8 +108,8 @@ openhab_setup() {
   cond_redirect usermod -aG "$gid" "$username" &> /dev/null
 
   echo -n "$(timestamp) [openHABian] Setting up openHAB service... "
-  if ! zram_dependency install ${ohPkgName}; then return 1; fi
-  if ! cond_redirect systemctl -q daemon-reload &> /dev/null; then echo "FAILED (daemon-reload)"; return 1; fi
+  if ! cond_redirect zram_dependency install ${ohPkgName}; then return 1; fi
```
mstormi, line 111: mixes again

functions/openhab.bash
```diff
@@ -108,8 +108,8 @@ openhab_setup() {
...
+  if ! cond_redirect systemctl -q daemon-reload; then echo "FAILED (daemon-reload)"; return 1; fi
```
mstormi, line 112: didn't check but doesn't zram_dependency do that ?

functions/system.bash
```diff
@@ -38,7 +38,7 @@ basic_packages() {
 
   if cond_redirect apt-get install --yes screen vim nano mc vfu bash-completion coreutils \
     htop curl wget multitail git util-linux bzip2 zip unzip xz-utils cpufrequtils lsb-release \
-    software-properties-common man-db whiptail acl usbutils dirmngr arping; \
+    software-properties-common man-db whiptail acl usbutils dirmngr arping apt-utils; \
```
mstormi, line 41: why? hasn't that been in ?

functions/zram.bats
```diff
@@ -82,22 +82,20 @@ check_zram_removal() {
   if [ "$status" -ne 0 ]; then echo "$output" >&3; fi
   [ "$status" -eq 0 ]
   echo -e "# ${COL_GREEN}$(timestamp) [openHABian] Availability of zram mounts verified.${COL_DEF}" >&3
-  run init_zram_mounts "uninstall" 3>&-
```
mstormi, line 85: this is from your zram patch (which I've only noticed by coincidence as you mentioned it in NEWS.md)
and like above this should not be in this PR

tests/Dockerfile.rpi4-BATS
```diff
@@ -1,43 +1,25 @@
...
-    systemd-remount-fs.service \
-    getty.target \
-    graphical.target \
-    kmod-static-nodes.service
-
-ENTRYPOINT ["./tests/entry.sh"]
-
 RUN git clone https://github.com/gdraheim/docker-systemctl-replacement && \
-    cp docker-systemctl-replacement/files/docker/systemctl3.py /bin/systemctl
+    cp docker-systemctl-replacement/files/docker/systemctl3.py /usr/bin/systemctl
```
mstormi, line 19: Err I'm not getting this.
You said you removed my 3rd party systemctl so why this ?

tests/entry.sh
```diff
@@ -47,28 +56,50 @@ mount_dev() {
...
+	printf '%q ' "$@" >> /etc/balenaApp.sh
+	chmod +x /etc/balenaApp.sh
+
+	mkdir -p /etc/systemd/system/balena.service.d
+	cat <<-EOF > /etc/systemd/system/balena.service.d/override.conf
+		[Service]
+		WorkingDirectory=$(pwd)
+	EOF
+
+	sleep infinity &
```
mstormi, line 77: do not comprehend all of this so cannot comment on whether it works or has side effects etc

Either way a 'sleep infinity' sounds like a strange thing.

mstormi (changes requested):
you mixed many old non-CI related changes in (wrong rebase ?), please remove this

The CI changes I do not comprehend (which is ok) but they don't really or fully seem to do what you said they would (*remove* the Python systemctl replacement)

.github/workflows/bats-action.yml
ecdye: No, that is just the system that GH Actions is running on. Docker is running on a different system.

build.bash
ecdye: I noticed the bug while working on ci changes and this particular function is called in CI so I fixed it.

functions/find.bash
ecdye: I mixed these in because I discovered that the output was funky without the `cond_redirect`

functions/influxdb+grafana.bash
ecdye: That had everything to do with CI, I spent quite a bit of time trying to get Grafana to work with the new Initsystem in BAT, and it did on my system, but for whatever reason it would not work properly in the GH Actions env online. That is why the initsystem is only being used on the installation tests and the BATS tests still use the systemctl replacement.

functions/influxdb+grafana.bash
ecdye: See above

functions/influxdb+grafana.bash
ecdye: see above.

functions/java-jre.bash
ecdye: That I discovered as an error it the CI testing but I can easily unmit that.

functions/openhab.bash
ecdye: No, it outputs wrong the way it is, say your on my branch then it will output whatever the branch name is instead of openHAB (version 2 or three doesn't matter).

functions/openhab.bash
ecdye: Do what?

functions/system.bash
ecdye: No, I noticed that we were getining delayed configuration because it wasn't in so I added it because it should have already been in.

functions/zram.bats
ecdye: That is mostly true but in this case it is related to CI, I know this is a jumbled patch but it was essentially a bunch of cumulated patches that I discovered while doing CI upgrades.

tests/Dockerfile.rpi4-BATS
ecdye: Only in the non-BATS systems, see my comment on the Grafana for more info.

tests/entry.sh
ecdye: It keeps the initsystem running so that the docker container does not close right away when you run it.

functions/openhab.bash
ecdye: Figured it out, yes actually so we can remove that legacy code. I'll do that as a separate patch

ecdye:
@mstormi I have added a bunch of the smaller unrelated patches to the codebase and removed them from this PR, please review again.

mstormi (approved):
Well ok then I don't mind fixing bugs along the way it just seemed to be a little to way off the core PR.
Thanks
DNT but as it is only CI we'll see.

PR #1638, opened 2021-12-26
Fix nginx authentication setup prompt

No3x:
Prompt for username and password did not show up because `$auth` was checked for `"yes"` instead of `"true"` ([set 2 lines before](https://github.com/No3x/openhabian/commit/d39383f173c8e8bff587a16de454a4096cc3f45e#diff-6db4f5084cdcf651cf5b9f8eb2265f3d5b46e1140997e4c83c588aa6df338f27L497)).

### DIFF
--- a/functions/packages.bash
+++ b/functions/packages.bash
@@ -496,7 +496,7 @@ nginx_setup() {
   if openhab3_is_installed || (whiptail --title "Authentication setup" --yesno "Would you like to secure your openHAB interface with username and password?" 7 80); then
     auth="true"
   fi
-  if [[ "$auth" == "yes" ]]; then
+  if [[ "$auth" == "true" ]]; then
     if nginxUsername="$(whiptail --title "Authentication setup" --inputbox "\\nEnter a username to sign into openHAB:" 9 80 openhab 3>&1 1>&2 2>&3)"; then
       while [[ -z $nginxPass ]]; do
         if ! nginxPass1="$(whiptail --title "Authentication setup" --passwordbox "\\nEnter a password for ${nginxUsername}:" 9 80 3>&1 1>&2 2>&3)"; then echo "CANCELED"; return 0; fi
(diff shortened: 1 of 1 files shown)


ecdye:
The change looks good, can I have you follow the instructions on the DCO test that failed to fix it so we have a proper sign off?

No3x:
I just noticed that this is not sufficient to fix basic auth setup. Later there is a openhab2_installed check where the credentials are used which is not executed on oh3 install. Changed this too assuming that it's fine for oh2 and oh3 as well. Please check the change.

functions/packages.bash
```diff
@@ -572,7 +572,7 @@ nginx_setup() {
   if cond_redirect sed -i -e 's|DOMAINNAME|'"${domain}"'|g' /etc/nginx/sites-enabled/openhab; then echo "OK"; else echo "FAILED (set domain name)"; return 1; fi
 
   if [[ $auth == "true" ]]; then
-    if openhab2_is_installed; then
+    if openhab_is_installed; then
```
ecdye, line 575: This is actually not needed, if you look closely this is actually just provided for backwards  compatibility with OH2 the routine already accounts for OH3 on line 582.

ecdye (changes requested):
Sign-off looks good. Please revert the second change as it is not actually necessary

functions/packages.bash
No3x: You are right I have overseen this line. I just understood that for OH3 another method (`Header Set-Cookie`) is used. The nginx config does not use the credentials anyway (only in OH2 case). See https://github.com/openhab/openhabian/blob/a1800e5583e2cc7edb298da8d175f070168ac5bf/functions/packages.bash#L575-L579
But then it does not makes sense to show the auth prompt anyway for OH3? At this point my little change reveals a lot of questions to me and I'm not sure what the intended results for OH2 and OH3 cases are. Now that you are aware of the issue someone may reconsider the setup for this step.

ecdye: OK, I see what's happened here, when we were making the upgrade to OH3 we made a mistake in the coding for this function. I will go ahead and fix it is a commit and incorporate this change as well if you don't have any issue with it.

PR #1647, opened 2022-01-08
Dynamically adjust zram memory constraints

### LINKED ISSUE #1646: Zram size based on ram size ?
See e.g. https://community.openhab.org/t/openhabian-restore-doesnt-work/131433/5)

I myself have come close to those limits, too,. and now I am also using larger sizes since I moved from a RPi 3 to a 4 w/ 2GB.

I think we should make zram sizes in ztab dependent on hardware (RAM size).
@ecdye wdyt? maybe double it for RPi4 ?

ecdye:
Fixes #1646

### DIFF
--- a/functions/helpers.bash
+++ b/functions/helpers.bash
@@ -397,6 +397,20 @@ has_lowmem() {
   if [[ $totalMemory -lt 900000 ]]; then return 0; else return 1; fi
 }
 
+## Returns 0 / true if device has more than 1500MB of total memory
+## Returns 1 / false if device has less than 1500MB of total memory
+##
+##    has_hasmem()
+##
+has_highmem() {
+  local totalMemory
+
+  totalMemory="$(awk '/MemTotal/ {print $2}' /proc/meminfo)"
+
+  if [[ -z $totalMemory ]]; then return 1; fi # assume that device does not have high memory
+  if [[ $totalMemory -gt 1500000 ]]; then return 0; else return 1; fi
+}
+
 ## Attempt to update apt package lists 10 times
 ## unless 'apt-get update' evaulates to 0.
 ## Sleeps for 1 second between each attempt.
--- a/functions/openhab.bash
+++ b/functions/openhab.bash
@@ -209,6 +209,8 @@ openhab_misc() {
   echo -n "$(timestamp) [openHABian] Optimizing openHAB to run on low memory single board computers... "
   if has_lowmem; then
     if cond_redirect sed -i -e 's|^EXTRA_JAVA_OPTS=.*$|EXTRA_JAVA_OPTS="-Xms16m -Xmx256m"|g' /etc/default/openhab; then echo "OK"; else echo "FAILED"; return 1; fi
+  elif has_highmem; then
+    if cond_redirect sed -i -e '/^[^#]/ s/\(^.*EXTRA_JAVA_OPTS=.*$\)/#\ \1/' /etc/default/openhab; then echo "OK"; else echo "FAILED"; return 1; fi
   else
     if cond_redirect sed -i -e 's|^EXTRA_JAVA_OPTS=.*$|EXTRA_JAVA_OPTS="-Xms192m -Xmx320m"|g' /etc/default/openhab; then echo "OK"; else echo "FAILED"; return 1; fi
   fi
--- a/functions/zram.bash
+++ b/functions/zram.bash
@@ -41,7 +41,7 @@ init_zram_mounts() {
       fi
     fi
 
-    if ! dpkg -s 'make' 'libattr1-dev' &> /dev/null; then
+    if ! dpkg -s 'make' &> /dev/null; then
       echo -n "$(timestamp) [openHABian] Installing zram required package (make)... "
       if cond_redirect apt-get install --yes make; then echo "OK"; else echo "FAILED"; return 1; fi
     fi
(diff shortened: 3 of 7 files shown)


mstormi (changes requested):
I'd ask to increase disk_size even further. mem_limit is critically affecting RAM and therefore important to system stability but disk_size isn't really but it's beneficial in situations like the restore in the forum post. Worth to increase for standard ztab as well ?

ecdye:
> I'd ask to increase disk_size even further. mem_limit is critically affecting RAM and therefore important to system stability but disk_size isn't really but it's beneficial in situations like the restore in the forum post. 

I don't think so. As it states in the zram documentation the disk size should only be a max of 400% more than your mem_limit otherwise it is not beneficial and will add increased overhead. Additionally, the 400% value is only really applicable in cases where the data is super compressible compressible, and I would venture to guess that openHAB's data is only moderately compressible on average. The values I used are a little conservative, however I believe that in the vast majority of cases they will be more than sufficient without causing too much impact on system resources on the targeted systems. Keep in mind we are targeting a large user base and the vast majority of which don't need super large amounts of memory for their persistence data. Keep in mind the only data actually stored in RAM is the data accumulated since the last sync to disk (ie reboot). 

> Worth to increase for standard ztab as well ?

Once again no, both for the reasons listed above, and for the simple fact that up to this point there has been no need to and many users have had great results. I prefer to operate under the don't fix what ain't broke rule. 


In the end I believe that the values I have chosen will offer the greatest benefit to the most users. Much more beyond what I've done here really enters custom territory better left to the enterprising user who is comfortable editing the config file to obtain the optimal result for their setup.

mstormi:
> As it states in the zram documentation the disk size should only be a max of 400% more than your mem_limit otherwise it is not beneficial and will add increased overhead.

Increased overhead does not mean not beneficial, just a decrease in storage efficiency. And that all by itself is also no reason not to go there when the (application level) benefits outweigh that. I'd also think that the increase in overhead won't be much really.

> Additionally, the 400% value is only really applicable in cases where the data is super compressible compressible
> and I would venture to guess that openHAB's data is only moderately compressible on average.

I think we must distinguish logging from persistence.
I'm using openHAB default rrd4j persistence. Compression ratio is ~5:1.
And watch that log compression ratio!

[12:33:52] root@mysmarthouse:/var/log# zramctl
NAME       ALGORITHM DISKSIZE   DATA COMPR TOTAL STREAMS MOUNTPOINT
/dev/zram2 zstd          600M 407,9M 77,6M  200M       4 /opt/zram/zram2
/dev/zram1 zstd         1000M 381,9M  6,7M  186M       4 /opt/zram/zram1
/dev/zram0 lzo-rle       600M 182,3M 42,8M 76,2M       4 [SWAP]

> Keep in mind the only data actually stored in RAM is the data accumulated since the last sync to disk (ie reboot).

Ok for logging. Logging is also pretty well limited due to logrotate/log4j config.
But my main concern is persistence and for persistence your statement is not really true. Files get written to on every item update. From zram point of view the whole file is stored in RAM from there on as zram cannot do sub-file level diffs.

ecdye:
I don't completely agree, as this is the first time we have ever really seen this issue I would prefer to be more conservative and adjust later. Perhaps something we could do in the future is add a menu option to enable an even more aggressive zram profile.

mstormi:
I still dislike the persistence disk_limit.
Doubling it like you did is not enough IMHO, I went for 1G on persistence and 600M on the others myself and suggest you do here, too.
As I said there I'm an average user in terms of persistence size, plus things will worsen the longer have your system in operations and accumulate data.

But what's worse, it applies to old small mem systems as well. As we should not double mem_size there we should at least quadruple disk_size.

mstormi:
A note on conservatism w.r.t ztab values:
Remember it only applies to new installs. With issues to eventually rise because of this but only on new installs, we must not be conservative in changing or we won't be seeing effect nor getting feedback.
So if we think disk_size should be say 1G then we should deploy 1G from now on and not anything less.
Conservatism is good where it applies to running systems but on new ones it annoyingly slows down the process of determining a well working value.

ecdye:
Because you have expressed so much concern over this issue I'm going to do some more research and testing to try and find the optimal values for my 1G system and then scale up from there. As a handy reference that might interest you to give more details about why I believe what I do see <https://unix.stackexchange.com/questions/594817/why-does-zram-occupy-much-more-memory-compared-to-its-compressed-value>.

ecdye:
Additionally, the COMPR value that you keep referencing is a bit misleading. Try running `zramctl --output-all` and pay attention to the MEM_USED value as that is that amount of RAM that the zram device is actually using including all of the other overhead, and it matches my research of about 100%-150% of the mem_limit as the disksize value.

mstormi:
I would not think we should increase swap size as that potentially increases need for RAM during standard operations.
Maybe ok on 2+G but risky and not beneficial on 1G.
Swapping is a tradeoff memory vs CPU. We don't have CPU shortage so no benefit in trading in RAM to lower CPU.
Current swap is large enough not to make openHABian/Java run into out of virtual memory (for the whole OS that is), but if there's more swap available, OS will proactively put copies of RAM pages there it might never need to remove from process space RAM but this will occupy swap hence RAM because swap is in zram.

ecdye:
@mstormi Unrelated to our unresolved issue with devices size. Something in the GHA environment changed that no longer allows our BATS zram tests to run. I am working on implementing a very complex testing procedure to allow me to run tests on the main repo for zram. As such, and because I have not been able to find a simple or good fix for our tests I believe that at the moment the best option is to remove them from our testing suite. Thoughts?

mstormi:
> Thoughts?

No immediate ones. I'm ok with removing zram bats tests for the time being.

mstormi (approved):
while I'd still like to see more for persistence such as those 400M/1G you spent for logging,
let's merge and see

ecdye:
I'm actually about to push some breaking changes to zram-config proper and will need to update the code here for installation so this is perfect timing. I'll merge and then push those other unrelated changes once I have merged them in zram-config.

PR #1655, opened 2022-02-19
Fix TIP box presentation

Confectrician:
The missing newline probably makes the rendering fail for the link on the website.
GitHub seems to work, because it does not know anything about the tip syntax.

Fixes openhab/openhab-docs#1760


ecdye (approved):
LGTM

PR #1664, opened 2022-03-31
Support Raspberry Pi OS 64bit

ecdye:

### DIFF
--- a/.github/workflows/build-action.yml
+++ b/.github/workflows/build-action.yml
@@ -55,12 +55,12 @@ jobs:
          run: |
            sudo -E ./tests/ci-setup.bash github ${{ github.ref_name }}
            sed -i -e "s|^userpw=.*$|userpw=\"${{secrets.USERPW}}\"|g" build-image/openhabian.pi-raspios32.conf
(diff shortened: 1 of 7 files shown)


ecdye:
@mstormi I'm ready and confident enough to support it. I've been using it for the past couple of months with no real issues so I think we are good to go.

I'm debating if we should bump to v1.8 for such a major support change or not WDYT?

mstormi:
To speak frankly, to me there's still no point in running in 64 and only people who want to be on the bleeding edge will want to.
I'm not keen on supporting this in a sense that I would need to spend any of my time on going after or reproducing issues that are specific to 64bit and that wouldn't be raised if that image wasn't available. I would still refuse to call it 'supported', that's just raising expectations I don't want to meet. Even more so with you going on your mission soon. 
So I don't mind to remove the beta tag but I would not make a new version or big announcement of this as that'll just direct attention.

ecdye:
Fair enough, it can just be a minor change in the new v1.7 release. In my testing it had been stable to so I am fairly confident saying that it doesn't need the beta tag anymore.

Issue #1702, opened 2022-07-22
openHABian v1.7.4 64bit install fails... because of new user/password change wizard?


krikk:
## Issue information:
Fresh install of openHABian v1.7.4 64bit (flashed via etcher to ssd, and booted directly via usb-sata cable) on a raspberry pi 4 8GB.
Install fails, with lots of failures, nearly nothing is installed...

the raspi boots fine until i get this screen on the attached screen:
![image](https://user-images.githubusercontent.com/5165445/180511169-c53258b1-eff8-4d30-8df8-74bcd390b852.png)
i do not think that this screen is normal? ...i did not press anything, and let it run, i created a maximum debug log

this error seems strange:
```
+ dpkg-reconfigure --frontend=noninteractive locales
debconf: DbDriver "config": /var/cache/debconf/config.dat is locked by another process: Resource temporarily unavailable
+ return 1
+ echo 'FAILED (reconfigure locales)'
FAILED (reconfigure locales)
```
had a very similar issue when first tried to install "Raspberry Pi OS lite 64bit" ...could not install anything via ssh session until i saw that is was stuck in the locale screen whick locked apt database, and because of that i could not install anything...

## Debug information:
[first-boot.log](https://github.com/openhab/openhabian/files/9170242/first-boot.log)
[openhabian.conf.txt](https://github.com/openhab/openhabian/files/9170247/openhabian.conf.txt)
[apt_list.txt](https://github.com/openhab/openhabian/files/9170248/apt_list.txt)

## System information:
Raspberry Pi 4 8gb 64bit
```
root@openhabian:~# cat /etc/os-release
PRETTY_NAME="Debian GNU/Linux 11 (bullseye)"
NAME="Debian GNU/Linux"
VERSION_ID="11"
VERSION="11 (bullseye)"
VERSION_CODENAME=bullseye
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"

root@openhabian:~# uname -m
aarch64


```

krikk:
i think i found the issue: [https://www.raspberrypi.com/news/raspberry-pi-bullseye-update-april-2022/](https://www.raspberrypi.com/news/raspberry-pi-bullseye-update-april-2022/)

after adding this userconf.txt to the /boot partition it starts to work as it should:

content of userconf.txt for user openhabian and password openhabian (created according to the linked article:)
[...]

PR #1705, opened 2022-08-03
added note about NOT attaching a keyboard during first time Install

krikk:
why? ...because of this issue: https://github.com/openhab/openhabian/issues/1702#issuecomment-1198658036


mstormi:
You would have needed to sign the DCO but no more need I've added your comment myself.
Thanks.

PR #1723, opened 2022-09-25
Fix locale setting with openhabian.conf not working

### LINKED ISSUE #1721: java locale ?
java needs to be run with -D options to have the proper locale in order to have OH display item states in the proper format (e.g. "," instead of "." as a separator behind the integer part of a number).
This may or may not be a problem in current openHABian but needs to be checked and changed, eventually. If so derive proper setting from the OS locale setting ?

EXTRA_JAVA_OPTS: "-Duser.country=DE"
EXTRA_JAVA_OPTS: "-Duser.timezone=Europe/Berlin"

florian-h05:
Fixes #1721.

@mstormi Am I right with the following assumption?
Removed `dpkg-reconfigure --frontend=noninteractive locales` because it's only generating the locales, which is already done with `locale-gen`.

I tested the locale setup using `openhabian.conf` and the `openhabian-config` tool a few times, seems to work this time.
Maybe you could also test.

Signed-off-by: Florian Hotze <florianh_dev@icloud.com>

### DIFF
--- a/functions/system.bash
+++ b/functions/system.bash
@@ -150,6 +150,7 @@ locale_setting() {
   if [[ -n $INTERACTIVE ]]; then
     echo -n "$(timestamp) [openHABian] Setting locale based on user choice... "
     if ! dpkg-reconfigure locales; then echo "FAILED (reconfigure locales)"; return 1; fi
+    if ! syslocale="$(grep "^[[:space:]]*LANG=" /etc/default/locale | sed 's|LANG=||g')"; then echo "FAILED"; return 1; fi
   else
     echo -n "$(timestamp) [openHABian] Setting locale based on openhabian.conf... "
     # shellcheck disable=SC2154,SC2086
@@ -161,10 +162,10 @@ locale_setting() {
       done
       if ! cond_redirect locale-gen; then echo "FAILED (locale-gen)"; return 1; fi
     fi
-    if ! cond_redirect dpkg-reconfigure --frontend=noninteractive locales; then echo "FAILED (reconfigure locales)"; return 1; fi
+    # shellcheck disable=SC2154
+    syslocale=$system_default_locale
   fi
 
-  if ! syslocale="$(grep "^[[:space:]]*LANG=" /etc/default/locale | sed 's|LANG=||g')"; then echo "FAILED"; return 1; fi
   if cond_redirect update-locale LANG="${syslocale:-en_US.UTF-8}" LC_ALL="${syslocale:-en_US.UTF-8}" LC_CTYPE="${syslocale:-en_US.UTF-8}" LANGUAGE="${syslocale:-en_US.UTF-8}"; then echo "OK (reboot required)"; else echo "FAILED"; return 1; fi
 
   if [[ -n $INTERACTIVE ]]; then
--- a/functions/system.bash
+++ b/functions/system.bash
@@ -163,10 +163,10 @@ locale_setting() {
       if ! cond_redirect locale-gen; then echo "FAILED (locale-gen)"; return 1; fi
     fi
     # shellcheck disable=SC2154
-    syslocale=$system_default_locale
+    syslocale=${system_default_locale:-en_US.UTF-8}
   fi
 
-  if cond_redirect update-locale LANG="${syslocale:-en_US.UTF-8}" LC_ALL="${syslocale:-en_US.UTF-8}" LC_CTYPE="${syslocale:-en_US.UTF-8}" LANGUAGE="${syslocale:-en_US.UTF-8}"; then echo "OK (reboot required)"; else echo "FAILED"; return 1; fi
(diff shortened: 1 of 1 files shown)


mstormi:
> @mstormi Am I right with the following assumption?

I don't know.

> I tested the locale setup using `openhabian.conf` and the `openhabian-config` tool a few times, seems to work this time.

with that 'reconfigure' line removed ?

florian-h05:
> with that 'reconfigure' line removed ?

Seems like reconfigure with the noniteractive frontend only generates the locales, see the command output compared to `locale-gen`:

```shell
openhabian@smarthomevm ~> sudo dpkg-reconfigure --frontend=noninteractive locales
[sudo] password for openhabian: 
Generating locales (this might take a while)...
  de_DE.UTF-8... done
  en_US.UTF-8... done
Generation complete.
openhabian@smarthomevm ~> sudo locale-gen
Generating locales (this might take a while)...
  de_DE.UTF-8... done
  en_US.UTF-8... done
Generation complete.
```

Moreover, the reconfigure line with noniteractive frontend does not change anything in `/etc/default/locales`, the interactive dialogue allows to select the default system locale `LANG=`.

functions/system.bash
```diff
@@ -161,10 +162,10 @@ locale_setting() {
       done
       if ! cond_redirect locale-gen; then echo "FAILED (locale-gen)"; return 1; fi
     fi
-    if ! cond_redirect dpkg-reconfigure --frontend=noninteractive locales; then echo "FAILED (reconfigure locales)"; return 1; fi
+    # shellcheck disable=SC2154
+    syslocale=$system_default_locale
```
mstormi, line 166: what if system_default_locale is unset ?
And will /etc/default/locale be adapted to match system_default_locale (if set) ?

florian-h05: > what if system_default_locale is unset ?

Then en_US.UTF-8 will be used as defined in line 171:
```shell
if cond_redirect update-locale LANG="${syslocale:-en_US.UTF-8}" LC_ALL="${syslocale:-en_US.UTF-8}" LC_CTYPE="${syslocale:-en_US.UTF-8}" LANGUAGE="${syslocale:-en_US.UTF-8}"; then echo "OK (reboot required)"; else echo "FAILED"; return 1; fi
```

Use the following code snippet to verify that:
```shell

#!/bin/bash
# system_default_locale=de_DE.UTF-8

syslocale=$system_default_locale

echo ${syslocale:-en_US.UTF-8}
```

> And will /etc/default/locale be adapted to match system_default_locale (if set) ?

`update-locale` automatically updates `/etc/default/locale` on every call.

florian-h05:
> Did you check locale inside OH e.g. Komma als Trenner für Nachkommastellen ?

I used the `openhabian-config`tool to switch to en-EN.UTF-8, rebooted and checked `/etc/default/locale` as well as the _Trenner für Nachkommastellen ist Punkt_ in OH: changing the locale worked.
Then I again used the `openhabian-config` tool to switch back to de_DE.UTF-8, rebooted and checked the locale file as well as _Komma als Trenner für Nachkommastellen_ in OH: changing the locale worked again.

functions/system.bash
mstormi: but it's only called before your new line syslocale=$system_default_locale so does not have the value

Use defaults on assignments i.e. X=${Y:-Z}, can even be X=${Y:-${Z:-"z"}}

florian-h05: > but update-locale is only called before your new line syslocale=$system_default_locale so does not have the value by then

No, see:
```shell
…

    # shellcheck disable=SC2154
    syslocale=$system_default_locale
  fi

  if cond_redirect update-locale LANG="${syslocale:-en_US.UTF-8}" LC_ALL="${syslocale:-en_US.UTF-8}" LC_CTYPE="${syslocale:-en_US.UTF-8}" LANGUAGE="${syslocale:-en_US.UTF-8}"; then echo "OK (reboot required)"; else echo "FAILED"; return 1; fi
…
```

florian-h05: > Use defaults on assignments i.e. X=${Y:-z}, can even be X=${Y:-${Z:-z}}

In the line `syslocale=$system_default_locale`?

mstormi: yes and everywhere where appropriate

florian-h05: In the update-locale statement, for the `syslocale` var it is not required. (If interactive, an error is thrown if syslocale cant be assigned, and if unattened, syslocale gets a default from the assignment to system_default_locale.

Can I remove it?

mstormi: > if unattened, syslocale gets a default from the assignment to system_default_locale.

which is not guaranteed to be set as it's user-defined.

At some place you have to provide a default to apply to all use cases.

florian-h05: See my latest commit, I assign a default if system_default_locale is not set and we are in unattended mode.
If we are in interactive mode, syslocal can not be empty, because the function exits with an error if the assignment fails.

mstormi (approved):
alright then, thanks!

PR #1724, opened 2022-09-25
Update NodeJS installed check & armv6l NodeJS install

### LINKED ISSUE #1722: NodeJS newer than v14?
@mstormi 
While I was working on the addition of openhab & openhab_rules_tools to the openhabian-config tool, I noticed that openhabians [`node_is_installed`](https://github.com/openhab/openhabian/blob/96c045b943744c6c8458e7e089fd91d98e4348bd/functions/helpers.bash#L643) check requires NodeJS v14.

As I am on v16 on my openHAB machine (a Debian machine that is using the openhabian-config tool), this causes me some inconvenience.

Is there any reason that we need NodeJS v14 or is it possible to update the check to also „allow“ newer version?

florian-h05:
Fixes #1722.

## Changes

Updates the `node_is_installed` check from v14 to v16.
Install NodeJS v16 instead of v14 on `armv6l`.

## Testing

Should not require much testing, as openHABian is on NodeJS v16 since mid of June this year.
Futhermore, as described in #1722, all dependencies should run on NodeJS v16.

@mstormi 
The installation of v16 on `armv6l` is something I can‘t test due to the lack of such hardware.
WDYT?

### DIFF
--- a/functions/nodejs-apps.bash
+++ b/functions/nodejs-apps.bash
@@ -9,7 +9,7 @@ nodejs_setup() {
   if node_is_installed && ! is_armv6l; then return 0; fi
 
   local keyName="nodejs"
-  local link="https://unofficial-builds.nodejs.org/download/release/v14.18.2/node-v14.18.2-linux-armv6l.tar.xz"
+  local link="https://unofficial-builds.nodejs.org/download/release/v16.17.1/node-v16.17.1-linux-armv6l.tar.xz"
   local myDistro
   local temp
 
--- a/functions/helpers.bash
+++ b/functions/helpers.bash
@@ -640,7 +640,7 @@ grafana_is_installed() {
 ##    node_is_installed
 ##
 node_is_installed() {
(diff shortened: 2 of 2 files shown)


functions/nodejs-apps.bash
```diff
@@ -9,7 +9,7 @@ nodejs_setup() {
   if node_is_installed && ! is_armv6l; then return 0; fi
```
florian-h05, line 9: @mstormi 
Can you explain why `&& ! is_armv6l` is required for the early return?

mstormi: no

florian-h05: Okay.

The only reason I can think of, is to update NodeJS on `armv6l` systems, because we have package repo for thosee and therefore no package manager to update.

mstormi (approved):
LGTM did you test?

florian-h05:
I tested the check, it now correctly recognizes that NodeJS v16 is installed.
I am unable to test whether the installation of NodeJS v16 on armv61 is working due to the lack of hardware.

PR #1734, opened 2022-10-27
Improve readability of the branch selection dialog

jimtng:
The purpose of this PR is to avoid confusion of openHABian “version” vs openHAB version, and to hopefully make it clear that the branch selection dialog will change openhabian's version/branch, and not openhab's version.

See the discussion here https://community.openhab.org/t/openhabian-branch-selection-ui/140308

### DIFF
--- a/functions/openhabian.bash
+++ b/functions/openhabian.bash
@@ -164,47 +164,64 @@ openhabian_update_check() {
 ##
 openhabian_update() {
   local branch
+  local branchLabel
   local current
+  local dialogHeight=16
+  local radioOptions
   local introText
   local key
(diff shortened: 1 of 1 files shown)


mstormi:
I dislike the dual use of "stable" it'll lead to confusion in posts where people are way off from being accurate in the terms they use.
So want to avoid that. Can you call it "release" instead ?

jimtng:
> I dislike the dual use of the term "stable" it'll lead to confusion in posts.

Agreed, and thanks for the suggestion. I've updated the PR.

mstormi (approved):

Issue #1712, opened 2022-08-28
add NUT Network UPS Tools


mstormi:
Looking for anyone to provide NUT templates and menu-driven configuration to add a Uninterruptible Power Supply to an openHABian system, be it a RPi HAT (like Waveshare) or an external UPS to communicate to via serial or IP.
Primary demand is to properly shut down openHABian to save ZRAM contents to disk in case of an ongoing power outage.

rkoshak:
Starting some preliminary searching and the Waveshare UPS HAT is not supported by NUT as far as I can tell. BUT does have an I2C driver but it's not clear whether it will work out of the box or not. I can't find anything that mentions Waveshare and NUT on the same page.

Is this worth doing if we only support configuration of a couple of UPS devices? I can't afford to be buying dozens of these things to see if I can make them work. I have one Cyberpower UPS that I can test with and I'll try to find one or two HATs that NUT supports. 

I also won't have as whole lot of time to devote to this in the immediate future but I'm willing to take a stab at it if no one else steps up.

florian-h05:
Maybe I can provide my configuration files, I have a UPS attached to my Pi.

I have an Eaton 3S 700, currently attached to my Synology NAS and connected to the Pi with IP.
But in the past I also had that UPS attached directly to my Pi with USB.

rkoshak:
> I think the USB configuration should work for most people out of the box, port selection can be done automatically. (I haven‘t tested how this works with multiple USB devices attached, I only had the UPS attached).

All that's true but the driver selection can't as far as I'm aware and that's the major point I'm concerned with. There's a bunch of drivers and you have to look your UPS up in their big database to see which one is the correct driver for your UPS. Then you have special configuration parameters for each driver. 

Once the right driver is selected and configured it's smooth sailing. Maybe we just leave driver selection and configuration as an exercise for the end user, though that's the most complicated part so that feels unsatisfactory. Here's the hardware compatibility list: https://networkupstools.org/stable-hcl.html. Notice the right hand column. If you click on one of the drivers listed there, you can see the extra configuration parameters for that driver.

And of course, the hardware list is not complete. 😭

florian-h05:
I just checked your link, and unfortunately you are right, there‘s no auto driver selection.

[...]

PR #1759, opened 2023-02-06
Add Network UPS Tools to support Uninterruptable Power Supplies

florian-h05:
Part 1 of #1712.
Part 2 is Waveshare UPS HAT support, which I cannot contribute (at least for now).

## Description

This PR adds support for setting up Network UPS Tools both in unattended and interactive mode.

nut itself can be setup by openHABian to either connect to a local USB UPS (the code is ready to support more drivers, but I don't have any good proposals because `usbhid-ups` supports most modern UPSes in my feeling) or to another instance of nut on the local network, e.g. your Synology NAS.

The setup is currently not utilising `upssched`, which would allow to have mail notifications etc., but I think I can provide a follow up PR for this capability. I just don't want to blow up this PR anymore.

Note: the upsd users that are created have secure passwords that are generated on the fly by OpenSSL, except the one user that Synology NAS expects. Since this user cannot control anything, just read UPS data, I don't see a security problem there.

Note: I used the standard nut configuration files from Raspbian 11 and modified them to our needs. I decided against removing some sections, because this would avoid advanced users from manually editing these files.

## Testing

I have tested both the interactive and the unattended setup several times with:

- my Eaton 3S 700 directly connected to my dev Pi via USB
- my Eaton UPS connected to my Synology NAS, therefore using `netclient`

Ping @mstormi. BTW, I have seen your FOSDEM presentation, very interesting :+1:


openhabian-setup.sh
```diff
@@ -124,6 +124,7 @@ if [[ -n "$UNATTENDED" ]]; then
   jsscripting_npm_install "openhab_rules_tools"
   zram_setup
   exim_setup
+  nut_setup
   permissions_corrections
```
mstormi, line 128: setup without prior install ?

mstormi (commented):
LGTM except for my comment at the end.

Did not test but when you say you did that's sufficient.

openhabian-setup.sh
florian-h05: I chose the same pattern as for the exim setup: the `nut_setup` methods installs nut if not already installed.

functions/nut.bash
```diff
@@ -0,0 +1,120 @@
...
+  fi
+
+  local upsmasterpw="$(openssl rand -base64 12)"
+  local upsslavepw="$(openssl rand -base64 12)"
+  local upsmonitor="MONITOR ups@localhost 1 upsmaster ${upsmasterpw} master"
+
+  local whiptailTitle="Network UPS Tools"
+  local introText="Network UPS Tools (nut) adds an Uniterruptable Power Supply (UPS) to your openHABian system.\\nIn case of a power outage, it protects your system from data loss by ensuring a safe shutdown.\\n\\nWe will guide you through the setup."
+
+  if ! nut_is_installed; then nut_install "install"; fi
```
florian-h05, line 70: @mstormi Here it will install nut if it is not installed. I have tested this as well, works 👍

florian-h05:
@mstormi Can this get merged?

mstormi (approved):

florian-h05:
Thanks!

Only thing missing now is docs …

mstormi:
> Only thing missing now is docs …

will you be contributing them, too ?

Issue #1753, opened 2023-01-27
Influx instalation on latest Raspbian fails


bittmanv:
## Issue information:
I was started new instalation on RPi4/4GB machine to SD Card to migrate my "home/production" RP3 to new HW,
Successfully downloaded image (openhabian-pi-raspios64-202212201355-gitb2e396e-crcbd182bf5.img.xz), extracted to SD Card and installed. After finishing of main part instalation, I started with optional components via openahbian-config, mqtt/mosquitto. 
But Influx+grafana failed, after enabling debug I received error message about invalid public key, see bellow error from later apt update. 
It took just 4 hours to find following link , where is noticed that Yesterday influx rotated keys.
[https://www.influxdata.com/blog/linux-package-signing-key-rotation/](url)
During my retrties I also used older Raspbian image (openhabian-pi-raspios64-202208152014-gitbe9d23e-crc8b028846.img.xz ) and issue was there as well. **I assume that also updating users will be affected**

I also found some workaroud with follwing procedure
1) Install new Certificate based on influx above ( start in home directory)
`cd`
`wget -q https://repos.influxdata.com/influxdata-archive_compat.key`
`cat influxdata-archive_compat.key | gpg --dearmor | sudo tee /etc/apt/trusted.gpg.d/influxdata-archive_compat.gpg > /dev/null`
`echo 'deb [signed-by=/etc/apt/trusted.gpg.d/influxdata-archive_compat.gpg] https://repos.influxdata.com/debian stable main' | sudo tee /etc/apt/sources.list.d/influxdata.list`

2) now try to install Influx+grafana via openhabian-config, it will fail  ( this will create /etc/apt/sources,list.d/influxdb.list ),
4) edit file influxdb.list
`deb [signed-by=/etc/apt/trusted.gpg.d/influxdata-archive_compat.gpg] https://repos.influxdata.com/debian bullseye stable
`

run apt-get update
5) start openhabian-config Influx+Grafana again, now it should pass. ( at least for me it worked)

I'm not sure, if this issue can be fixed by modification of Openhabian image or is neccesary to wait for all respositories to be updated.

## Debug information:

openhabian@openhabian3:~ $ sudo apt-get update
Hit:1 http://davesteele.github.io/comitup/repo comitup InRelease
Hit:2 http://archive.raspberrypi.org/debian bullseye InRelease
Hit:3 http://security.debian.org/debian-security bullseye-security InRelease
[...]

PR #1763, opened 2023-02-22
Updated InfluxDB Key URL

stif:
InluxDB updated their GPG key and provides it under a new URL.
It is already changed in `main` branch but, is not backported to `OpenHAB3` branch.

See also Issue #1753 and commit [e2000f2](https://github.com/openhab/openhabian/commit/e2000f2185bd61716c2f53546455deea8fbd654f)

With this Pull Request the install influx+grafana.bash script will work for all users still on `Release` branch

### DIFF
--- a/functions/influxdb+grafana.bash
+++ b/functions/influxdb+grafana.bash
@@ -200,7 +200,7 @@ influxdb_install() {
   fi
 
   if ! influxdb_is_installed; then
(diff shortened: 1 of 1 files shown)


mstormi:
That doesn't make sense. The main branch which has the fix is forwarded to the openHAB3 branch at times when there's important things that need fixing. This I consider not to be important as it only affects few.
Anyone who wants to use it today can switch to main branch.

PR #1813, opened 2023-09-08
Fix service script on Openhabian (and every other *nix)

uqs:
rm(1) resides in /bin, not /usr/bin

### DIFF
--- a/includes/openhab-override.conf
+++ b/includes/openhab-override.conf
@@ -3,4 +3,4 @@ Wants=frontail.service homegear.service
 Before=frontail.service homegear.service
 
 [Service]
-ExecStartPre=+/usr/bin/rm -f /var/lock/LCK..ttyAMA0 /var/lock/LCK..ttyACM0
+ExecStartPre=+/bin/rm -f /var/lock/LCK..ttyAMA0 /var/lock/LCK..ttyACM0


mstormi:
That's not right. On Debian it is /usr/bin (too, at least), and openHABian is Debian only so it wasn't broken.
This was even sort of intentional as you should not be running on non-Debian.
Then again changing to /bin shouldn't do harm, so I did (but outside of this PR as the commit msgs would be wrong).

uqs:
To add some more history on how this was broken:
- System installed with OpenHabian circa 2019 on a Raspi using Stretch; OH 2.4
- Last month I did an in-place upgrade via Buster to Bullseye, still on OH 2.4
- Last week, upgraded to OH 4.x, and the OH startup script croaked because there's no /usr/bin/rm

So it stands to reason that Stretch had no /usr/bin/rm (I do wonder why that even was added later, rm(1) has been in /bin since time immemorial, and that's also true on every *nix that I've encountered over the last 20+ years).

What's most worrying is however how crappy the Debian upgrade process seems to be, if it can't do https://wiki.debian.org/UsrMerge properly

If you're curious, compare:
https://github.com/search?q=%22%2Fbin%2Frm&type=code

vs

https://github.com/search?q=%22%2Fusr%2Fbin%2Frm&type=code

also.

If you had suggestions on the commit message, you could've just written those here, I could've expanded more on why it's more robust and compatible this way. :)

PR #1828, opened 2024-01-04
nodejs-apps.bash: change installation of nodejs and npm to apt to fix missing npm

Larsen-Locke:
change installation of nodejs and npm to apt (but no offline and preoffline functionality)
Signed-off-by: Carsten Mogge <carsten.mogge@gmail.com>


Larsen-Locke:
does not work for bullseye and bookworm

PR #1829, opened 2024-01-07
Update nodejs-apps.bash

Larsen-Locke:
new LTS-version of nodejs.
Explicitely installing npm hoping that it solves bookworm-problem


Larsen-Locke:
doesn't work

PR #1836, opened 2024-01-27
Bugfix for different user than openhabian in zigbee2mqtt-installation

Larsen-Locke:
bugfix for different user than openhabian in zigbee2mqtt-installation
(restart of PR cause I messed up my repo and wasn't able to repair/rebase)

### DIFF
--- a/functions/nodejs-apps.bash
+++ b/functions/nodejs-apps.bash
@@ -233,7 +233,7 @@ zigbee2mqtt_setup() {
   local mqttPWText="\\nIf your MQTT-server requires a password, please enter it here:"
   local my_adapters
   local by_path_or_id
-  local mqttDefaultUser="openhabian"
+  local mqttDefaultUser="${username:-openhabian}"
   local mqttUser
   local serverIP
   local installSuccessText
--- a/functions/nodejs-apps.bash
+++ b/functions/nodejs-apps.bash
@@ -337,7 +337,7 @@ zigbee2mqtt_setup() {
   cd /opt/zigbee2mqtt || (echo "FAILED (cd)"; return 1)
   if ! cond_redirect sudo -u "${username:-openhabian}" npm ci ; then echo "FAILED (npm ci)"; return 1; fi
 
-  if ! cond_redirect install -o ${username:-openhabian} -g openhab -m 644 "${BASEDIR:-/opt/openhabian}/includes/zigbee2mqtt/configuration.yaml" /opt/zigbee2mqtt/data/; then echo "FAILED (install configuration.yaml)"; return 1; fi
+  if ! cond_redirect install -o "${username:-openhabian}" -g openhab -m 644 "${BASEDIR:-/opt/openhabian}/includes/zigbee2mqtt/configuration.yaml" /opt/zigbee2mqtt/data/; then echo "FAILED (install configuration.yaml)"; return 1; fi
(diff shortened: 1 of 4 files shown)


functions/packages.bash
```diff
@@ -267,9 +267,12 @@ mqtt_setup() {
   local mosquittoConf="/etc/mosquitto/mosquitto.conf"
   local mosquittoPasswd="/etc/mosquitto/passwd"
   local mqttPasswd
-  local mqttUser="openhabian"
-  local introText="The MQTT broker Eclipse Mosquitto will be installed from the official repository.\\n\\nIn addition, you can activate username:password authentication."
-  local questionText="\\nDo you want to secure your MQTT broker by a username:password combination? Every client will need to provide these upon connection.\\n\\nUsername will be '${mqttUser}', please provide a password (consisting of ASCII printable characters except space). Leave blank for no authentication, run setup again to change."
+  local mqttUser
+  local mqttDefaultUser="openhabian"
```
mstormi, line 271: shouldn't that rather be ${username:-openhabian}

functions/nodejs-apps.bash
```diff
@@ -346,7 +344,10 @@ zigbee2mqtt_setup() {
   echo "OK"
   
   echo -n "$(timestamp) [openHABian] Setting up Zigbee2MQTT service... "
-  if ! cond_redirect install -m 644 "${BASEDIR:-/opt/openhabian}"/includes/zigbee2mqtt/zigbee2mqtt.service /etc/systemd/system/; then echo "FAILED (install service)"; return 1; fi
+
+  sed -e "s|%user%|${username:-openhabian}|g" "${BASEDIR:-/opt/openhabian}/includes/zigbee2mqtt/zigbee2mqtt.service" | sudo -u "${username:-openhabian}" dd status=none of=/tmp/zigbee2mqtt.service
```
mstormi, line 348: dd? why ??

mstormi:
please rebase so the imager change (which is in ) should disappear

functions/nodejs-apps.bash
Larsen-Locke: It takes the input from sed and with user openhabian writes to another location (without conversion of course). Otherwise I would need another chmod. If you know a better way tell me.  
It's just the same as 9 lines above for configuration.yaml.

functions/packages.bash
Larsen-Locke: ok. done

functions/nodejs-apps.bash
mstormi: what's wrong about another chown ? That's way more transparent.
Btw you can also do sed -i to edit some file in place so rights are kept.

Larsen-Locke: I know. But the template in the includes-directory should not be changed.
dd duplicates from standard-in to standard-out, so I got everything done in one pipe. Why is it not transparent?
As I say it's just the same as above. Feel free to change if you like.

mstormi: Nah, I'm just the reviewer.
Code that isn't self-explaining and unnecessarily complex to understand isn't as good as simple, straightforward code even if that's longer.

Larsen-Locke: then I change the existing one, too.
I think it's fine now.
Don't forget to make a release including the raspberry-imager changes. Otherwise we drop out of raspberry imager, because the json-file with the new flags cannot be downloaded.

PR #1848, opened 2024-02-28
Bookworm WiFi

mstormi:
Move networking setup to NetworkManager in bookworm including Wi-Fi and hotspot.


build-image/first-boot.bash
```diff
@@ -156,12 +152,12 @@ fi
...
-    iwlist wlan0 scanning | grep "ESSID" | sed 's/^\s*ESSID:/\t- /g'
-    echo "$(timestamp) [openHABian] Please try again with your correct SSID and password. The following Wi-Fi configuration was used:"
-    cat /etc/wpa_supplicant/wpa_supplicant.conf
-    rm -f /etc/wpa_supplicant/wpa_supplicant.conf
+
+  if [[ "$hotSpot" == "enable" ]] && ! [[ -x $(command -v comitup) ]]; then
+    #echo -n "$(timestamp) [openHABian] Installing comitup hotspot (will reboot after)... "
+    if setup_hotspot "install"; then echo "OK"; else echo "FAILED"; fi
+    #echo "$(timestamp) [openHABian] Hotspot software installed. Rebooting your system to make it take effect!"
+    #reboot
```
jlaur, line 160: Are these commented lines left here intentionally?

mstormi: sort of, yes
unsure if it works without rebooting but seems now it does

Issue #1843, opened 2024-02-11
Bookworm uses yescrypt to encrypt passwords in /etc/shadow; password checker does not take this format into account


WolfgangSn:
## Issue information:
Bookworm uses yescrypt to encrypt passwords that are stored in /etc/shadow.
yescrypt encrypted passwords consist of 4 parameters separated by $.
E.g. $y$j9T$ydzju/dIJLh/6ohtOK5Ud0$ds0UQiwpQmJuq/jyzCP8bbmIM7GXgMqPadMlbFe7.t4

    id : y (yescrypt)
    param : j9T
    salt : ydzju/dIJLh/6ohtOK5Ud0
    hash : ds0UQiwpQmJuq/jyzCP8bbmIM7GXgMqPadMlbFe7.t4

For more detailed description e.g. see https://unix.stackexchange.com/questions/690679/what-does-j9t-mean-in-yescrypt-from-etc-shadow

openhabian's checker for default passwords does not take into account that yescrypt uses 4 instead of 3 fields.
```
  originalPassword="$(grep -w "$defaultUser" /etc/shadow | cut -d: -f2)"
  algo="$(echo "$originalPassword" | cut -d'$' -f2)"
  salt="$(echo "$originalPassword" | cut -d'$' -f3)"
  ...
  generatedPassword="$(perl -le 'print crypt("$ENV{defaultPassword}","\$$ENV{algo}\$$ENV{salt}\$")')"
```

This means that openhabian will report that no default password is being used although it is.

## Debug information:
Please provide all of the output provided by utilizing the `debugmode=maximum`
parameter in the `/etc/openhabian.conf` file. For more information on how to
accomplish this, please see [openhabian-DEBUG.md](https://github.com/openhab/openhabian/blob/main/docs/openhabian-DEBUG.md#create-a-debug-log).
```
place logs here
```

## System information:
Please tell us what OS you are running (Raspberry Pi OS, Debian, Ubuntu), what
HW you are running on (Raspberry Pi, amd64). Also include the output of
`cat /etc/os-release` and `uname -m`.
```
place output here
```

mstormi:
Wolfgang can you provide a PR ?

WolfgangSn:
> Wolfgang can you provide a PR ?

I can have a look at it.

WolfgangSn:
PR created https://github.com/openhab/openhabian/pull/1851

mstormi:
thanks for your contrib


Fixed-By: #1851

PR #1851, opened 2024-03-10
Update openhabian.bash to support yescrypt which is default hash with…

WolfgangSn:
… bookworm

To support default password detection when OS is bookworm yescrypt hash needs to be supported. Modified code detects if yescrypt is used.

### DIFF
--- a/functions/openhabian.bash
+++ b/functions/openhabian.bash
@@ -382,13 +382,15 @@ system_check_default_password() {
   if ! is_pi; then return 0; fi
 
   local algo
+  local param
   local defaultPassword
   local defaultUser
   local generatedPassword
   local introText
   local originalPassword
   local salt
-
+  local fields
+  
   if is_pi && id -u pi &> /dev/null; then
     defaultUser="pi"
     defaultPassword="raspberry"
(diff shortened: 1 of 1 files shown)


mstormi (commented):
I don't see a check for and different treatment based on the debian version.
Or is that implicit because of the number of fields ?

The code needs to work on both, bookworm *and* on bullseye.

WolfgangSn:
> Or is that implicit because of the number of fields ?
> 
> The code needs to work on both, bookworm _and_ on bullseye.
Indeed the number of the fields make the difference. Due to this is it will run on both OSs.
According to https://manpages.debian.org/bullseye/libcrypt-dev/crypt.5.en.html even bullseye is already able to use yescrypt.
Besides that not making it depending on the OS ( bookworm vs. bullseye ) it should run even with other OSs. 
Next to yesrypt also ghost-yescrypt uses the same number of fields and is covered.
In the alternative path ( case * ) other hash methods are covered in the same way as before/without the patch.

functions/openhabian.bash
```diff
@@ -397,12 +399,25 @@ system_check_default_password() {
     defaultPassword="openhabian"
   fi
   originalPassword="$(grep -w "$defaultUser" /etc/shadow | cut -d: -f2)"
-  algo="$(echo "$originalPassword" | cut -d'$' -f2)"
-  introText="The default password was detected on your system! That is a serious security concern. Bad guys or malicious programs in your subnet are able to gain root access!\\n\\nPlease set a strong password by typing the command 'passwd ${defaultUser}'!"
-  salt="$(echo "$originalPassword" | cut -d'$' -f3)"
-  export algo defaultPassword salt
-  generatedPassword="$(perl -le 'print crypt("$ENV{defaultPassword}","\$$ENV{algo}\$$ENV{salt}\$")')"
+  fields=$(echo $originalPassword|awk -F'$' '{print NF}')
```
mstormi, line 402:
```suggestion
  fields=$(echo "$originalPassword" | awk -F'$' '{print NF}')
```

mstormi (approved):

mstormi:
Closes: #1843

PR #1859, opened 2024-03-17
use xtext string interning for low memory devices

joerg1985:
This PR will enable xtext string interning for low memory devices to get rid of a lot Strings.
For my system this will reduce the number of strings inside a memory dump by ~50%.

https://github.com/eclipse/xtext/blob/main/org.eclipse.xtext/src/org/eclipse/xtext/naming/QualifiedName.java#L130-L137

### DIFF
--- a/functions/openhab.bash
+++ b/functions/openhab.bash
@@ -220,11 +220,11 @@ openhab_misc() {
 
   echo -n "$(timestamp) [openHABian] Optimizing openHAB to run on low memory single board computers... "
   if has_lowmem; then
-    if cond_redirect sed -i -e 's|^EXTRA_JAVA_OPTS=.*$|EXTRA_JAVA_OPTS="-Xms16m -Xmx256m -XX:+ExitOnOutOfMemoryError"|g' /etc/default/openhab; then echo "OK"; else echo "FAILED"; return 1; fi
+    if cond_redirect sed -i -e 's|^EXTRA_JAVA_OPTS=.*$|EXTRA_JAVA_OPTS="-Xms16m -Xmx256m -Xss1024k -XX:-TieredCompilation -XX:+ExitOnOutOfMemoryError -Dxtext.qn.interning=true"|g' /etc/default/openhab; then echo "OK"; else echo "FAILED"; return 1; fi
   elif has_highmem; then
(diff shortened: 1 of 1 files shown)


joerg1985:
This change could be applied to all memory sizes, but this change is using less memory by using more cpu. When alot of memory is available there is no need to have this enabled. 
I am not familar with the Azul Zulu JVM and using GC1 (with this flag) might break something.

mstormi:
cpu is absolutely not an issue with OH, mem almost always is.
IF GC1 is Zulu only we must not apply that, correct, all flags must work with all JDKs or at least get ignored by those that don't support an option.
I suggest you change the PR to do `-Dxtext.qn.interning=true -XX:+UseStringDeduplication` on ALL mem sizes, plus
-Xss1024k on lowmem boxes

mstormi:
**UPDATE**
First deployment attempt failed so we need to scale back. Let's validate options one by one.

mstormi (changes requested):
Please test with the full range of Java opts I gave in my (edited) comment
and change this PR to include them all, if applicable for all mem sizes.

mstormi:
does that thumb-up mean you will be testing & updating the PR ?

joerg1985:
Yes, but testing might take some time to see if it is working and not just starting with these flags.

joerg1985:
When all these flags set i have issues calling a graph with one year history, the 4 cores of my raspi 3a stuck at 100% and 60°C for ~4 minutes. I will try some other settings to fix this.

mstormi:
> When all these flags set i have issues calling a graph with one year history, the 4 cores of my raspi 3a stuck at 100% and 60°C for ~4 minutes. I will try some other settings to fix this.

Ok please remember to double check i.e. if if is a) reproduceable and b) does not occur without those option.
Processing a full year's data is a lot on a 1GB box I wouldn't be sure if that works on current openHABian.

joerg1985:
@mstormi i have updated the PR, some flags had no (`-XX:G1PeriodicGCInterval=60000 -XX:G1PeriodicGCSystemLoadThreshold=4`) or negative `(-XX:MinHeapFreeRatio=4 -XX:MaxHeapFreeRatio=12)` impact so i dropped them.

The negative impact was reproducible and resulted probably from the overhead by releasing and acquiring the memory to aggressive.

mstormi:
Thanks. Did you exclusively test on a 1GB box i.e. RPi 3A ?

mstormi (approved):

joerg1985:
> Thanks. Did you exclusively test on a 1GB box i.e. RPi 3A ?

Yes, but it is a 512 MB box with zram (zstd, 280%).

PR #1861, opened 2024-04-02
add Zulu JDK 21 installation on 64bit

holgerfriedrich:
Requires a native 64 bit image, running a 64 bit kernel with 32-bit OS is not sufficient.

I have been working to make OH compile an run with Java21 openhab/openhab-distro#1590. Not all of the required PRs have been merged, so we are not yet there.
This PR allows to install Java21 on x64/aarch64 systems.

Running the snapshot distro without modification is not yet possible, still waiting for the PR openhab/openhab-distro#1640
 to be merged.

### DIFF
--- a/build.bash
+++ b/build.bash
@@ -7,7 +7,7 @@ set -e
 ####################################################################
 
 usage() {
-  echo -e "Usage: $(basename "$0") <platform> [oldstable]"
+  echo -e "Usage: $(basename "$0") <platform> [oldstable|latest]"
   echo -e "\\nCurrently supported platforms: rpi, rpi64"
 }
 
--- a/functions/java-jre.bash
+++ b/functions/java-jre.bash
@@ -86,18 +86,18 @@ java_install_or_update() {
       java_zulu_prerequisite "${cached_java_opt:-Zulu11-32}"
       java_zulu_install "${cached_java_opt:-Zulu11-32}"
     fi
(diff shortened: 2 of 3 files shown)


mstormi (changes requested):
>Requires a native 64 bit image, running a 64 bit kernel with 32-bit OS is not sufficient.

Please add proper checks for this in the install routine to ensure it only installs when the prerequisites are met

holgerfriedrich:
Hi @mstormi,
I think the checks are already there, shared by the Zulu11-64 and Zulu 21-64 targets.
```if is_aarch64 || is_x86_64 && [[ $(getconf LONG_BIT) == 64 ]]; then``` 
The check with ```getconf``` is really necessary - I had my RPI image that showed aarch64 but LONG_BIT 32 with a 32 bit userland.

If the prerequisites are not met, the common error reaction is to show a message in interactive setup (good) or to install Zulu-32 in the interactive case (maybe not so well suited).
As the code for the fallback is shared and there is no Java 21 32bit available anyway, this might be ok.

WDYT?

Btw: anything wrong with my change which causes the bats test to fail, or are the tests affected by something else?

mstormi (approved):
That's fine then
no BATS test failing is unrelated

holgerfriedrich:
Thanks, Markus

PR #1869, opened 2024-05-25
Update packages.bash eddited miflora installation

thicknesss:
Debain 12 (Bookworm) forces the use of python virtual enviroments for some packages. Installation script for miflora-mqtt-daemon is edited according to the proposed change in the README of miflora-mqtt-daemon. See PR: https://github.com/ThomDietrich/miflora-mqtt-daemon/pull/176.

PR https://github.com/ThomDietrich/miflora-mqtt-daemon/pull/175 is made to update the systemd service file. which will be used also when installing the package via openhabian-config.


mstormi (approved):
LGTM, thanks

PR #1871, opened 2024-06-07
Fix broken Wireguard test

ecdye:


mstormi (approved):
does wireguard actually work on bookworm?

ecdye:
> does wireguard actually work on bookworm?

No clue, I don't actually use it, but the test sure seems to be working now. Unless someone who uses it can verify that it works properly, I don't know if we can find out.

PR #1878, opened 2024-07-01
Update zram install code for upstream submodule

ecdye:
Depends ecdye/zram-config#114

### DIFF
--- a/functions/zram.bash
+++ b/functions/zram.bash
@@ -15,7 +15,7 @@ install_zram_code() {
   if [[ -d "${1}/zram-config" ]]; then
     if cond_redirect update_git_repo "${1}/zram-config" "openHAB"; then echo "OK"; else echo "FAILED (update zram)"; return 1; fi
   else
-    if cond_redirect git clone --branch "openHAB" "$zramGit" "$1"/zram-config; then echo "OK"; else echo "FAILED (clone zram)"; return 1; fi
+    if cond_redirect git clone --recurse-submodules --branch "openHAB" "$zramGit" "$1"/zram-config; then echo "OK"; else echo "FAILED (clone zram)"; return 1; fi
   fi
 }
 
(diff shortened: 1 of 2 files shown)


ecdye:
@mstormi I haven't merged upstream changes (so tests will fail) because I don't want it to instantly break new installs, but can you review these changes? Let's get any quirks sorted out before I merge upstream, and then maybe release a new image to have the new changes cached for new installs, since it is a fairly substantial change.

functions/zram.bash
```diff
@@ -41,26 +41,28 @@ init_zram_mounts() {
...
+      if cond_redirect apt-get install --yes -o DPkg::Lock::Timeout="$APTTIMEOUT" meson; then echo "OK"; else echo "FAILED"; return 1; fi
     fi
 
     install_zram_code "$zramInstallLocation"
 
     echo -n "$(timestamp) [openHABian] Setting up OverlayFS... "
-    if ! cond_redirect make --always-make --directory="$zramInstallLocation"/zram-config/overlayfs-tools; then echo "FAILED (make overlayfs)"; return 1; fi
-    if ! cond_redirect mkdir -p /usr/local/lib/zram-config/; then echo "FAILED (create directory)"; return 1; fi
-    if cond_redirect install -m 755 "$zramInstallLocation"/zram-config/overlayfs-tools/overlay /usr/local/lib/zram-config/overlay; then echo "OK"; else echo "FAILED (install overlayfs)"; return 1; fi
+    if ! cond_redirect rm -rf "$zramInstallLocation"/zram-config/overlayfs-tools/builddir; then echo "FAILED (remove builddir)"; return 1; fi
```
mstormi, line 52: return-if-fail isnt required, is it?
(not sure but I think even with rm -f the cmd will exit with code != 0 if the dir does not exist)

functions/zram.bash
```diff
@@ -121,10 +123,13 @@ init_zram_mounts() {
     if cond_redirect install_zram_code "$zramInstallLocation"; then echo "OK"; else echo "FAILED (update)"; return 1; fi
 
     echo -n "$(timestamp) [openHABian] Updating OverlayFS... "
-    if ! cond_redirect make --always-make --directory="$zramInstallLocation"/zram-config/overlayfs-tools; then echo "FAILED (make overlayfs)"; return 1; fi
-    if ! cond_redirect mkdir -p /usr/local/lib/zram-config/; then echo "FAILED (create directory)"; return 1; fi
-    if cond_redirect install -m 755 "$zramInstallLocation"/zram-config/overlayfs-tools/overlay /usr/local/lib/zram-config/overlay; then echo "OK"; else echo "FAILED (install overlayfs)"; return 1; fi
+    if ! cond_redirect rm -rf "$zramInstallLocation"/zram-config/overlayfs-tools/builddir; then echo "FAILED (remove builddir)"; return 1; fi
```
mstormi, line 126: see above

mstormi (dismissed):
Cannot test or comment on if the upstream merge and everything in combination would do would do, but this PR part LGTM.

functions/zram.bash
ecdye: True, I could simplify that. Will do

PR #1891, opened 2024-07-26
Update release action

ecdye:
This switches our release action to an actively maintained action that will continue to recieve future updates, as our old action was no longer being maintained.


mstormi (approved):
great, I've recently hit that, too, and went looking for an alternative but didn't find any at that time.

Issue #1870, opened 2024-05-28
Startup loop


mstormi:
At times, after updating the  openHAB package, startup is stuck in a loop, spitting lots of java component errors.
I came to think if we can supervise startup, detect and break these loops ?
Unsure how, though.

Maybe if detected, `pkill -9 java` maybe add that to the update script option ? 
Add `openhab-cli clean-cache`, too ? (I think it's done on package install so may not help to do again but not 100% sure)
Or add a systemd timer to do a regular health check maybe ? Note that can be a dangerous sword should it mistakenly detect OH to be down. It should only be active for say some hours after openhab service startup time.

BClark09:
I've not had this myself, but I have seen many talk about this issue. Is the openHAB process restarting (i.e. Karaf) or is the loop happening within openHAB's functions. If the former, then we can do a lot if an exit code is issued.

> Add `openhab-cli clean-cache`, too ? (I think it's done on package install so may not help to do again but not 100% sure)

[It is done on package install or on update.](https://github.com/openhab/openhab-linuxpkg/blob/10061acd36524afb12a033fea6dcf142b399bf56/resources/control-runtime/preinst#L84).

> Or add a systemd timer to do a regular health check maybe ?

There are things we can test by polling the activeness of core bundles: https://github.com/openhab/openhab-linuxpkg/issues/138

> I came to think if we can supervise startup, detect and break these loops ?

I've been wondering if it's related to the amount of time that openHAB takes to shut down or restart. If it takes more than 120 seconds, then the openHAB service is killed with a `SIGTERM`, and I wonder if this has any effect on the quality of the update.

ecdye:
So are we just trying to add a temporary fix on the openHABian side until the underlying bug can be corrected with openHAB and its Java process?

mstormi:
yeah, although AFAIK there's no such bug filed with openHAB

miloit:
I have never seen this, but if you have @mstormi than file an issue so that they girls and boys aware if it

mstormi:
> I've been wondering if it's related to the amount of time that openHAB takes to shut down or restart. [If it takes more than 120 seconds](https://github.com/openhab/openhab-linuxpkg/blob/10061acd36524afb12a033fea6dcf142b399bf56/resources/systemd/openhab.service#L27C1-L27C15), then the openHAB service is killed with a `SIGTERM`, and I wonder if this has any effect on the quality of the update.

[...]

PR #1897, opened 2024-08-11
pkill java after restart to eventually break out of startup loop

mstormi:
Fixes: #1870 

not 100% sure it's ok to remove the restart command, but the effect should be the same (except for the speedup, of course)

### DIFF
--- a/functions/openhab.bash
+++ b/functions/openhab.bash
@@ -132,8 +132,8 @@ openhab_setup() {
   dashboard_add_tile "openhabiandocs"
 
   # see https://github.com/openhab/openhab-core/issues/1937
-  echo -n "$(timestamp) [openHABian] Restarting openHAB service to play it safe... "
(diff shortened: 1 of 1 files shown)


ecdye (changes requested):
I would add back the restart after because it appears that the checks are failing because openHAB never actually restarts otherwise. See if adding it back helps and then we could go from there.

mstormi:
will do but wonder if I can instead add "-now" to systemctl enable cmd some lines up

ecdye (approved):
LGTM, if you wanted to try adding the enable -now, I think it's worth a shot, but I wonder if that wouldn't cover all cases because of upgrades.

mstormi:
Well EITHER `-now` and no additional restart OR no `-now` and restart but not both.
Switch or leave it like it is now with the merged PR?

Issue #1868, opened 2024-05-24
install zigbee2mqtt with ethernet coordinator


moodyblue:
## Issue information:
I have a zigbee internet coordinator, more specifically the SLZB-06. When trying to install zigbee2mqtt using openhabian menu it is not possible to specify the IP address of the coordinator and the installation is not done.

## System information:
Raspberry Pi 4B-2G
openhabian 64-bit

Output of
`cat /etc/os-release` and `uname -m`.
```
openhabian@openhabian:~ $ cat /etc/os-release
PRETTY_NAME="Debian GNU/Linux 12 (bookworm)"
NAME="Debian GNU/Linux"
VERSION_ID="12"
VERSION="12 (bookworm)"
VERSION_CODENAME=bookworm
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
openhabian@openhabian:~ $ uname -m
aarch64
```

mstormi:
@Larsen-Locke please decide if you want to add this as a feature

Larsen-Locke:
Currently not. I didn't have that hardware and till now there was only one user asking for it.

mstormi:
@moodyblue sorry

moodyblue:
I understand. What if I develop this myself ?

mstormi:
Please do. I'd be happy to merge a PR if you put up one.

moodyblue:
Where is `zigbee2mqtt_setup` ? This is one of the files that need be changed.

The change is simple, in your skeleton `configuration.yaml` includes these lines:
```
serial:
  port: "/dev/serial/%adapter"
```

In my production case these lines are:
```
serial:
  port: tcp://192.168.129.41:6638
```
(the ip address is installation specific, the port is device specific, for example the slzb-06 uses port 6638, but I have another coordinator that uses port 1053)

moodyblue:
PR submitted, hope that it is ok.

PR #1905, opened 2024-09-22
Preparation for Z2M tcp coordinators

moodyblue:
The objective of this pull request is to prepare openhabian to support network coordinators. Presently it only supports usb coordinators.

I can't test if the code is Ok because I do not have serial coordinators. Can you pls test it for me, and modify as adequate ?

It is a preparation to fix https://github.com/openhab/openhabian/issues/1868

### DIFF
--- a/functions/nodejs-apps.bash
+++ b/functions/nodejs-apps.bash
@@ -297,14 +297,14 @@ zigbee2mqtt_setup() {
 
   # get usb adapters for radio menu
   while IFS= read -r line; do
-    my_adapters="$my_adapters $line $loopSel "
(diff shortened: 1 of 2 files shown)


mstormi:
> I can't test if the code is Ok because I do not have serial coordinators. Can you pls test it for me, and modify as adequate ?

Err, no, sorry. We don't have that hw either. Please go find someone to test it for you.

And please make it one composite PR, not multiple ones in stages. Thanks.

moodyblue:
Ok, Pls reject this PR. In the meantime someone gave me a usb device, so now I can test it myself.

PR #1923, opened 2024-10-23
Update GitHub Actions workflows

ecdye:


mstormi:
not sure what's your intention here but remember we need some sort of auto-deployment inside docker "platform" to "simulate" installation in github cloud to see effects of our bash-level changes

ecdye:
Yeah, this should still accomplish exactly the same as before, simply just updating the RPi docker images we test against and using a matrix to simplify the config file. It still runs all the same tests.

PR #1926, opened 2024-11-23
Remove legacy openHAB2 support

ecdye:


functions/openhabian.bash
```diff
@@ -298,14 +291,6 @@ migrate_installation() {
     fi
```
ecdye, line 291: @mstormi Should we even need to keep this now?

mstormi: no lets remove

mstormi (dismissed):
did not check for comprehensiveness
but what I saw LGTM
just act on the one question/comment

functions/openhabian.bash
ecdye: Sounds good, I like that plan. Then it removes the responsibility from us but allows users who really want to to do it on their own using tools that will probably work better because its hard for us to account for everything in an automated migration.

mstormi:
Please check https://community.openhab.org/t/openhab-install-raspberry-pi-5/160899
maybe this PR has messed up bash_profiles?

ecdye:
Yeah it might have, I'll take a look tonight and try to fix it

PR #1929, opened 2024-12-12
Fix switching between different OH releases

holgerfriedrich:
Remove additional argument to openhab_setup which was missed in #1926.

It seems that since #1926 was applied, the OH release could no longer be changed.
When I tried to install snapshot today, I got an empty dialogue box, after hitting "ok" also empty package sources file, and still the old OH version installed.

Hope this PR catches all remaining locations.

### DIFF
--- a/functions/menu.bash
+++ b/functions/menu.bash
@@ -82,7 +82,7 @@ show_main_menu() {
       fi
     fi
     repo=$(apt-cache madison openhab | head -n 1 | awk '{ print $6 }' |cut -d'/' -f1)
-    openhab_setup "openHAB" "${repo:-release}"
+    openhab_setup "${repo:-release}"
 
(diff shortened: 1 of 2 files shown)


ecdye (approved):
LGTM

jlaur:
Would it be possible to release this fast? This is probably the reason why I can't switch from release (4.2.3) to milestone (RC1) right now to proceed with the last tests on my production system. Personally I can manually apply the changes from this PR and proceed, but others might want to be able to switch today or tomorrow.

holgerfriedrich:
@jlaur "release" branch of openhabian is not yet updated; if you switch openhabian to "latest" and restart openhabian-config, it will work....

ecdye:
I'll update it now so that the changes are fixed for everyone else.

PR #1931, opened 2024-12-13
replace outdated link to mounting tutorial in openhabian-amanda.md

TAKeanice:
Use latest web archive snapshot instead. Long term it should be  replaced with some other source.


ecdye (changes requested):
Thanks for the update! Would you mind doing a little research and finding a recent tutorial equivalent that way we can just use that instead of an old archived link?

TAKeanice:
> Thanks for the update! Would you mind doing a little research and finding a recent tutorial equivalent that way we can just use that instead of an old archived link?

sure, as I commented on the commit I am planning on doing that, but in the meantime, the archived link is better than a broken one.

PR #1940, opened 2025-01-06
Add official support for Java 21

### LINKED ISSUE #1937: Check for Java 21 for OH5?
Some maintainer asked if we could check for Java 21 to exist when switching to or installing snapshots (OH5 now which requires Java 21)

not sure what / how much we would need to do here.

ecdye:
This also adds checks to make sure that Java 21 is installed before upgrading to openHAB 5.

Fixes #1937

### DIFF
--- a/tests/Dockerfile.amd64-installation
+++ b/tests/Dockerfile.amd64-installation
@@ -29,5 +29,6 @@ ENTRYPOINT ["./tests/entry.sh"]
 
 # Setup openHABian environment
 COPY . /opt/openhabian/
+COPY ./build-image/webserver.bash /boot/firmware/webserver.bash
 WORKDIR /opt/openhabian/
 RUN install -m 755 ./tests/runlevel /sbin/runlevel
--- a/tests/Dockerfile.amd64-installation
+++ b/tests/Dockerfile.amd64-installation
@@ -29,6 +29,5 @@ ENTRYPOINT ["./tests/entry.sh"]
 
 # Setup openHABian environment
 COPY . /opt/openhabian/
-COPY ./build-image/webserver.bash /boot/firmware/webserver.bash
 WORKDIR /opt/openhabian/
 RUN install -m 755 ./tests/runlevel /sbin/runlevel
--- a/tests/Dockerfile.rpi3-installation
+++ b/tests/Dockerfile.rpi3-installation
@@ -30,6 +30,5 @@ ENTRYPOINT ["./tests/entry.sh"]
 
 # Setup openHABian environment
 COPY . /opt/openhabian/
-COPY ./build-image/webserver.bash /boot/firmware/webserver.bash
 WORKDIR /opt/openhabian/
 RUN install -m 755 ./tests/runlevel /sbin/runlevel
--- a/tests/Dockerfile.rpi3-installation
+++ b/tests/Dockerfile.rpi3-installation
@@ -30,5 +30,6 @@ ENTRYPOINT ["./tests/entry.sh"]
 
 # Setup openHABian environment
 COPY . /opt/openhabian/
+COPY ./build-image/webserver.bash /boot/firmware/webserver.bash
 WORKDIR /opt/openhabian/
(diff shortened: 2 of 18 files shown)


Larsen-Locke:
zigbee2mqtt test is disabled here. Couldn't we enable it back?

ecdye:
> zigbee2mqtt test is disabled here. Couldn't we enable it back?

Yep!, I just forgot to after we merged the fix. I'll do that now.

ecdye:
@mstormi I can't quite fathom why, but OpenJDK 21 really doesn't like the GitHub Actions env. When I run it locally on Docker it works perfectly but not in Actions. I feel confident enough with all of this for it to be reviewed though.

I'm on the fence about defaulting to the Temurin over OpenJDK for 21 because of the issues I'm seeing here and because then it has to use unstable for the source. It also upgrades other packages when installing from unstable because they are dependencies which I don't really like (ie `libc6`). Temurin is also now the most widely supported JVM in general that Oracle doesn't build reference implementations anyways.

Let me know your thoughts.

(There ended up being a bunch of random fixes to automated testing that I found and added while debugging that stupid OpenJDK 21 issue that I never could figure out)

tests/Dockerfile.amd64-installation
```diff
@@ -29,5 +29,6 @@ ENTRYPOINT ["./tests/entry.sh"]
 
 # Setup openHABian environment
 COPY . /opt/openhabian/
+COPY ./build-image/webserver.bash /boot/firmware/webserver.bash
```
mstormi, line 32: no `firmware` please

tests/Dockerfile.rpi3-installation
```diff
@@ -30,5 +30,6 @@ ENTRYPOINT ["./tests/entry.sh"]
 
 # Setup openHABian environment
 COPY . /opt/openhabian/
+COPY ./build-image/webserver.bash /boot/firmware/webserver.bash
```
mstormi, line 33: same

tests/Dockerfile.rpi5-installation
```diff
@@ -29,5 +29,6 @@ ENTRYPOINT ["./tests/entry.sh"]
 
 # Setup openHABian environment
 COPY . /opt/openhabian/
+COPY ./build-image/webserver.bash /boot/firmware/webserver.bash
```
mstormi, line 32: same

mstormi (changes requested):
LGTM overall although I'm unable to determine by just looking at the diffs if OpenJDK 21 is really available now and where it's coming from.

The firmware comments at the end are guessed but you probably wrote that code before you saw I revoked your recent changes.

ecdye:
> LGTM overall although I'm unable to determine by just looking at the diffs if OpenJDK 21 is really available now and where it's coming from.
> 
> The firmware comments at the end are guessed but you probably wrote that code before you saw I revoked your recent changes.

Yeah I'll fix those. OpenJDK 21 is available from unstable and Temurin. Both are provided as options and should work well.

mstormi (approved):

mstormi:
Alright. I think you should release a new image with this in, plus the fixed hotspot.
Don't forget to forward main to openHAB branch.

ecdye:
Should I make it another patch release or do a minor to 1.10? I'm thinking 1.10 because of the addition of Java 21 support.

WDYT?

mstormi:
pls also add the changelog like in older releases, too.
(got word that the hotspot still doesn't do but it's bedtime for me today)

mstormi:
Please check build tests, not sure why GH auto-cancels/refuses to execute them

https://github.com/openhab/openhabian/actions/runs/13155090413/job/36711247938?pr=1970
https://github.com/openhab/openhabian/actions/runs/13155090413/job/36711249237?pr=1970

ecdye:
I think it's just because there are some random bugs with the arm runners because they're still in testing. Usually if you rerun them it works fine. I was willing to put up with it because of how much faster they are than qemu

PR #1948, opened 2025-01-09
fix installing optionally given oh pkg version

mstormi:
option to openhab_setup() wasn't documented but implemented and makes sense to have when we need to go back to an older pkg version for some reason


mstormi:
for now, yes. I have an openhabian fork that implements a "forceohpkg" option.
Do you think we would need that as an openhabian.conf option? Isn't that too confusing as we already have a ohpkg option there?

ecdye:
OK that makes more sense, I forgot that we had that. LGTM

PR #1965, opened 2025-02-02
Use Ubuntu 22.04 for builds

### LINKED ISSUE #1964: Builds failing with systemd-nspawn Exec format error
Build fail with
```2025-02-02_19:26:47_UTC [openHABian] Caching packages for offline install...
rm: cannot remove '/tmp/***-build-raspios32-image.L5jqK/mnt/boot': Device or resource busy
```
The error msg is from cleanup_build() but the problem is that the `systemd-nspawn` command in `build.bash` `offline_install_modifications`  is failing.
I've experimented some and problem possibly is that `systemd-nspaw` is incompatible architecture now?
Here I issued a simple 'pwd'

+ systemd-nspawn --directory=/tmp/openhabian-build-raspios64-image.2o62D/mnt pwd
Spawning container mnt on /tmp/openhabian-build-raspios64-image.2o62D/mnt.
Press Ctrl-] three times within 1s to kill container.
execv(pwd) failed: Exec format error
Container mnt failed with error code 1.

ecdye:
Fixes #1964

### DIFF
--- a/.github/workflows/build-action.yml
+++ b/.github/workflows/build-action.yml
@@ -24,7 +24,7 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name != github.event.pull_request.base.repo.full_name
 
     steps:


ecdye:
@mstormi This appears to fix the issue. For now I suggest we merge, but I'll try and figure out what changed in the 24.04 image that is now the latest that breaks our build so we can migrate to it in the future.

ecdye:
Closeing in favor of #1966

PR #1966, opened 2025-02-03
Take advantage of ARM runners

### LINKED ISSUE #1964: Builds failing with systemd-nspawn Exec format error

ecdye:
Fixes #1964


mstormi (approved):
Good find and work, thanks

mstormi:
I'm confused.
Looking at build-action.yml, it doesn't use the fairly new matrix definition yet, no runners: directive etc
 I have my clone of the openhabian repo, rebased it to this PR. But it keeps showing the old error.

So why does it work when building this image but not with mine?

Plus, images are tiny (500MB) compared to the latest release image sizes (~1.3G). Doubt they're okay.

ecdye:
I'm not sure, I think all the bugs have been worked out with them image build and firmware stuff here so I wonder if you try again on your end if it will be fixed now.

mstormi:
> I'm not sure, I think all the bugs have been worked out with them image build and firmware stuff here so I wonder if you try again on your end if it will be fixed now.

Yes, that's what I did and yes it worked out now. With that image I was now able to validate hotspot is working again.

PR #1972, opened 2025-02-08
ARM v7 is not RPi3 but RPi2

mstormi:


ecdye:
@mstormi Balena doesn't seem to support the RPi 2 on bookworm. RPi3 has an armv7 mode which is why we chose it. Originally it could operate in armv7 mode or arm64 mode and it fit the test suite best

mstormi:
Okay, that's news to me, never heard of that "v7 mode" and I don't see it being selected anywhere?
So I believe it's using standard arm64 so it's the same as the rpi5 test, isn't it?

ecdye:
> Okay, that's news to me, never heard of that "v7 mode" and I don't see it being selected anywhere? So I believe it's using standard arm64 so it's the same as the rpi5 test, isn't it?

No balena serves two different RPi3 images by default (the one we have) it is the arm32 version. They also provide a separate `balenalib/raspberrypi3-64` image for 64bit mode.

mstormi:
but arm32 does not imply armv7

ecdye:
Yes but if you look at the source files for the Dockerfiles they are pulling from armv7
See https://github.com/balena-io-library/base-images/blob/master/balena-base-images/device-base/raspberrypi3/debian/bookworm/build/Dockerfile

mstormi:
Then what about using that armv7hf base? why detour via rpi3?
I now use that in this PR.

mstormi:
Yeah but that's a different balena image then.
So what about directly using that armv7hf base?
I've now put that to use that in this PR, let's see if that works out.
We might still need to detour via rpi3 as a platform because (likely) GitHub won't be providing ARMv7 runners but let's see.

ecdye:
> Yeah but that's a different balena image then. So what about directly using that armv7hf base? I've now put that to use in this PR, wdyt?

I mean I don't particularly have any issue with it but it just tests more generally rather that using the balena image which more closely emulates the RPi hardware. 

Is there any particular reason you want to add this test? Especially considering that soon it won't even really be necessary with OH 5.

mstormi:
As I understand it so far we have NOT been testing armv7 but armv8 twice (RPi 3,5) which has been redundant.
My latest proposal should change that so in itself makes sense.
Even with OH5 having a test env for RPi2 and older makes sense. Even an all-64 Java code can run on 32bit OS+HW given the right JDK.

ecdye:
Well as I already mentioned, no it actually was testing armv7 because that is the base balena was using. 

As for running OH5 on a 32 bit JRE, yes some things may work, but my understanding was that some of the core libraries have breaking 64 bit changes that were going to be merged that would prevent it from working at all. Additionally it won't be officially supported for sure.

I don't have any major issues with merging this, but it makes no significant difference because of the reasons stated above.

mstormi:
Ok now I got it why it's armv7.
But it's transient/indirect so while my PR doesn't change anything about it, it makes this explicit thus clearer.
As you said you don't have anything against it I'll merge.

PR #1974, opened 2025-02-10
fix? EXTRA_JAVA_OPTS in highmem case

mstormi:
some very vague complaint of timezone spec via java option not working  made me investigate this and I don't see a reason to treat the highmem case different than the others.

plus, removed deprecated delayed rules loading (relevant only in OH2)


mstormi:
pls squash commit messages

PR #1976, opened 2025-02-15
Fix is_openhab_installed check for openHAB 5 snapshots

florian-h05:


florian-h05:
@openhab/openhabian-maintainers FYI. This is very annoying when testing openHAB 5 snapshot.

florian-h05:
I don’t switch back and forth …
My use case is that with openHAB 5 parts of openhabian-config, like apply permission recommendations, are not working anymore because openhabian thinks no openHAB is installed.

We are about to release the first openHAB 5 milestone tomorrow, openHABian should be ready to accept openHAB 5 and not say „you must install openHAB first“.

mstormi (approved):

Issue #1982, opened 2025-02-25
On installing openhabian the password defined in openhabian.config is ignored.


name-lake:
On installing the openhabian v1.10.2 via Ethernet cable, from the file:
`openhabian-raspios64-latest-202502221936-crc837e760e.img.xz`
the password defined in the `openhabian.config` is not implemented, the standard password `openhabian` was implemented instead.
Enclosed the files `openhabian.config`and `first-boot.log`.
[openhabian.config.zip](https://github.com/user-attachments/files/18965919/openhabian.config.zip)
[first-boot.log.zip](https://github.com/user-attachments/files/18965921/first-boot.log.zip)

PR #1983, opened 2025-02-26
fix passwd change via initial openhabian.conf

mstormi:
Fixes: #1982

### DIFF
--- a/build-image/first-boot.bash
+++ b/build-image/first-boot.bash
@@ -72,7 +72,7 @@ fi
 
 echo -n "$(timestamp) [openHABian] Changing default username ... "
 # shellcheck disable=SC2154
-if [[ -n ${userName} ]] || ! id "$defaultUserAndGroup" &> /dev/null || id "$userName" &> /dev/null; then
+if [[ -z ${userName} ]] || ! id "$defaultUserAndGroup" &> /dev/null || id "$userName" &> /dev/null; then
   echo "SKIPPED"
 else
   usermod -l "$userName" "$defaultUserAndGroup"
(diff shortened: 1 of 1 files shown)


build-image/first-boot.bash
```diff
@@ -70,17 +70,23 @@ if is_raspbian || is_raspios; then
   rm -f "/etc/sudoers.d/010_pi-nopasswd"
 fi
 
-echo -n "$(timestamp) [openHABian] Changing default username and password... "
+echo -n "$(timestamp) [openHABian] Changing default username ... "
 # shellcheck disable=SC2154
 if [[ -v ${userName} ]] || ! id "$defaultUserAndGroup" &> /dev/null || id "$userName" &> /dev/null; then
```
ecdye, line 75: OK, so I understand the changes you are making but I think there might be an issue from older versions.

What does the `-v` flag do? I can't find it in the bash manual, I believe the `-v` instances should be replaced with `-n` right?

build-image/first-boot.bash
```diff
@@ -70,17 +70,23 @@ if is_raspbian || is_raspios; then
...
 else
   usermod -l "$userName" "$defaultUserAndGroup"
   usermod -m -d "/home/$userName" "$userName"
   groupmod -n "$groupName" "$defaultUserAndGroup"
-  echo "${userName}:${userpw:-$userName}" | chpasswd
   echo "OK"
 fi
+echo -n "$(timestamp) [openHABian] Changing default password... "
+# shellcheck disable=SC2154
+if [[ -v ${userpw} ]]; then
```
ecdye, line 85: Same as above.

ecdye (changes requested):

build-image/first-boot.bash
mstormi: right, changed that

mstormi:
right, changed that

ecdye (approved):

PR #1990, opened 2025-03-29
fix? timezone not being overwritten on openhab pkg updates

mstormi:
Hi Ethan,
unsure how to fix this properly.

Existing code is called from within openhab_setup() and assumes the `EXTRA_JAVA_OPTS` line in `/etc/default/openhab` to end with "...interning=true" which it currently only does on fresh installs.
Not on older installations, and not sure how it looks like on non-RPi.

I now removed the ending from the pattern so it now should be replacing `EXTRA_JAVA_OPTS` at any time it calls the routine.

I'm unsure however what the Linux package does to the file. The original open issue was about upgrading openhab pkg on some oldish openHABian box (that was probably installed before and did not have that `interning=true` ending so replacement didn't apply).

PS: do you know how to restrict visibility of GitHub issues or PRs to maintainers (i.e. ourselves) only? According to Google seems we cannot, but I'd like to have a private comms channels
I really disenjoy users to follow every single step of ours and sometimes even tap into discussions, and to be frank, I'm already a little p*ssed off by the tone how it was raised as an issue and keeps being commented on the forum.

### DIFF
--- a/functions/openhab.bash
+++ b/functions/openhab.bash
@@ -225,7 +225,7 @@ openhab_misc() {
   if cond_redirect sed -i -e 's|^#*.*OPENHAB_HTTPS_PORT=.*$|OPENHAB_HTTPS_PORT=8443|g' /etc/default/openhab; then echo "OK"; else echo "FAILED"; return 1; fi
 
   echo -n "$(timestamp) [openHABian] Setting openHAB timezone... "
-  if cond_redirect sed -ri "s|^(EXTRA_JAVA_OPTS.*interning=true)|\1 -Duser.timezone=${timezone:-Europe/London}|g" /etc/default/openhab; then echo "OK"; else echo "FAILED"; return 1; fi
+  if cond_redirect sed -ri "s|^(EXTRA_JAVA_OPTS.*)|\1 -Duser.timezone=${timezone:-Europe/London}|g" /etc/default/openhab; then echo "OK"; else echo "FAILED"; return 1; fi
 }
 
 ## Create a openHAB dashboard title and image for the input application.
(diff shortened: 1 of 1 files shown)


functions/openhab.bash
```diff
@@ -225,7 +225,7 @@ openhab_misc() {
   if cond_redirect sed -i -e 's|^#*.*OPENHAB_HTTPS_PORT=.*$|OPENHAB_HTTPS_PORT=8443|g' /etc/default/openhab; then echo "OK"; else echo "FAILED"; return 1; fi
 
   echo -n "$(timestamp) [openHABian] Setting openHAB timezone... "
-  if cond_redirect sed -ri "s|^(EXTRA_JAVA_OPTS.*interning=true)|\1 -Duser.timezone=${timezone:-Europe/London}|g" /etc/default/openhab; then echo "OK"; else echo "FAILED"; return 1; fi
+  if cond_redirect sed -ri "s|^(EXTRA_JAVA_OPTS.*)|\1 -Duser.timezone=${timezone:-Europe/London}|g" /etc/default/openhab; then echo "OK"; else echo "FAILED"; return 1; fi
```
ecdye, line 228: I think we can match the more general case for both conditions by doing this:
```suggestion
  if cond_redirect sed -ri "s|^(EXTRA_JAVA_OPTS.*?)(interning=true)?|\1\2 -Duser.timezone=${timezone:-Europe/London}|g" /etc/default/openhab; then echo "OK"; else echo "FAILED"; return 1; fi
```

ecdye (changes requested):
I'm not sure how this will interact with upgrades either, but it should fix the issue on all new installs or if the user re-runs the openHAB install/setup.

ecdye:
> PS: do you know how to restrict visibility of GitHub issues or PRs to maintainers (i.e. ourselves) only? According to Google seems we cannot, but I'd like to have a private comms channels I really disenjoy users to follow every single step of ours and sometimes even tap into discussions, and to be frank, I'm already a little p*ssed off by the tone how it was raised as an issue and keeps being commented on the forum.

I don't believe that it is possible to restrict visibility to just us. We could probably limit interaction on certain issues to just maintainers / members of the openHAB organization, but that wouldn't really change too much.

For some issues, it could be appropriate to discuss certain aspects (particularly sensitive changes or security concerns) over email, but I think that in general we should keep most of the communication on GitHub and for sure have issues tracked on GitHub to follow with the guidance given for keeping things open.

I do share some of your frustration though and will look into seeing if there are some alternatives for more maintainer side things that should be kept internal for determining the direction of the project in general, as that would be useful.

PR #1992, opened 2025-03-30
Remove code specific to unmaintained OS versions

ecdye:
This finishes the changes that I began making for unrelated reasons in
PR #1991 and uses more consistent identification of supported OS
versions.

Essentially, anything that is no longer maintained by it's respective
author/distributor should not be supported / maintained by us.

### DIFF
--- a/functions/helpers.bash
+++ b/functions/helpers.bash
@@ -333,7 +333,7 @@ is_raspios() {
 }
 # Represents the current OS versions we officially support
 is_supported() {
-  if is_bullseye || is_bookworm || is_sid || is_jellyfish || is_noble; then return 0; fi
+  if is_bullseye || is_bookworm || is_jellyfish || is_noble; then return 0; fi
   return 1;
 }
 # Debian/Raspbian oldstable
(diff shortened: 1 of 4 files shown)


functions/helpers.bash
```diff
@@ -331,17 +331,10 @@ is_raspios() {
...
-  return $?
-}
-# Debian/Raspbian oldoldstable
-is_buster() {
-  if [[ "$osrelease" == "buster" ]]; then return 0; fi
-  [[ $(cat /etc/*release*) =~ "buster" ]]
-  return $?
+# Represents the current OS versions we officially support
+is_supported() {
+  if is_bullseye || is_bookworm || is_sid || is_jellyfish || is_noble; then return 0; fi
```
mstormi, line 336:
```suggestion
  if is_bullseye || is_bookworm || is_jellyfish || is_noble; then return 0; fi
```

mstormi: sid is the never-changing name of the ever-unstable release of Debian so I don't think we should support it here

mstormi (changes requested):

mstormi (approved):

Issue #2018, opened 2025-07-19
[URGENT] OH5, Java update fails on 32 bit


mstormi:
Triggered by a question from @holgerfriedrich in @openhab/maintainers if upgrading 32 installations result in problems,
I flashed a fresh 32 bit image to my test RPi 3 and chose to install latest milestone.

This resulted in an attempt TWICE to install Temurin 21 (as that's required for OH5) which probably failed because apparently there's no 32 bit ARM package available.

What's more, trouble is that we failed to check if Java 21 installed properly and nonetheless proceed to install the openhab 5 pkg.
Not 100 % sure what happens with OH5 on Java 17 but it's said to require Java 21.

=> HOW TO HANDLE
* refuse installing OH5? If so at what stage, if there's no Java 21 present?
* refuse menu 02 already (with still Java 17) ? How to know that it'll attempt installing OH5 at that stage?

```
+ apt-get install --download-only --yes temurin-21-jre
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 temurin-21-jre:arm64 : Depends: adoptium-ca-certificates:arm64 but it is not installable
                        Depends: libasound2:arm64 but it is not going to be installed
                        Depends: libc6:arm64 but it is not going to be installed
                        Depends: libfontconfig1:arm64 but it is not installable
                        Depends: libfreetype6:arm64 but it is not installable
                        Depends: libx11-6:arm64 but it is not installable
                        Depends: libxext6:arm64 but it is not installable
                        Depends: libxi6:arm64 but it is not installable
                        Depends: libxrender1:arm64 but it is not installable
                        Depends: libxtst6:arm64 but it is not installable
                        Depends: zlib1g:arm64 but it is not going to be installed
                        Recommends: fonts-dejavu-extra:arm64
E: Unable to correct problems, you have held broken packages.
+ return 100
+ echo FAILED
FAILED
+ return 1
++ timestamp
++ printf '%(%F_%T_%Z)T\n' -1
+ echo -n '2025-07-19_11:34:48_CEST [openHABian] Installing Adoptium Eclipse Temurin JDK... '
[...]

PR #2020, opened 2025-07-19
Refuse to install openHAB 5 if no Java 21 (after potential Java install)

mstormi:
OH 5 must not be installed / upgraded to if Java is not version 21.
This will happen on 32 bit ARM OS installations where there does not exist a proper 32 bit ARM JVM.

!! ATTENTION: This is the majority of our user base !!

Fixes: #2018


ecdye (approved):

ecdye:
Super busy right now, but from a glance this looks right. I'll try and give it some more attention when I have time, but can't say when.

PR #2026, opened 2025-07-29
support trixie

mstormi:
While I don't know why current image shows to be trixie nor how to fix that, that currently results in menu option 03 refusing to upgrade so to (technically) support trixie is overdue.
Seems to be breaking menu 03 (Upgrade), too, so I'll quick-fix it without @ecdye waiting for your approval.
Will also build another image.

Eventually will do 1.11 when 32bit Java 21 is in.

### DIFF
--- a/functions/helpers.bash
+++ b/functions/helpers.bash
@@ -333,19 +333,25 @@ is_raspios() {
 }
 # Represents the current OS versions we officially support
 is_supported() {
-  if is_bullseye || is_bookworm || is_jellyfish || is_noble; then return 0; fi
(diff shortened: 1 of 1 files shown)


ecdye:
No worries, this is fine

mstormi:
For some unknown reason the current release pointed to 1.10.3 and 1.10.4 was draft only; I changed that.

Now however building another image 1.10.5 fails, see #2028 (despite the name it's less the tests and more that the image build doesnt finish).
Do you have any idea what to look for?

PR #2031, opened 2025-07-29
Temurin21 32bit

mstormi:
On 32 bit OS on ARM, there is no JVM 21 
The OpenEMS project has compiled Adoptium (Temurin) 21 in 32 bit
https://community.openems.io/t/java-21-jdk-for-arm32/3253

We now install this and output a user warning that this is experimental.
Thing is, many users still run on 32 bit openHABian and cannot upgrade to OH5 without a JVM 21.
So while this is unsupported and tagged experimental, it's better than not to install anything.


ecdye (approved):
LGTM, did not test

PR #2033, opened 2025-07-30
Java->Temurin21, default debuglevel on

mstormi:


matmai:
@mstormi I think it makes sense to also change the docs for this. https://github.com/openhab/openhabian/blob/10f5bf247a7f55c198d564847ab9d10030aea2a0/README.md#L28

PR #2041, opened 2025-08-01
Update docs for OH5 and 64 bit

mstormi:
OHian now refuses to upgrade to OH5 via menu 03 if on 32 bit.
But I've kept manual install via menu 41 / 45 as a backdoor.

I've added some more explanations, warnings and references, please cross-check.
Do you think this is appropriately worded?

This I want to release as v1.11 (so among others it changes the description in Raspi imager, too).


ecdye (approved):
LGTM, I think you got everything. In light of the feedback in the maintainers group on the Forum, while this is good for the moment, it might be wise to revisit it after a consensus has been drawn.

PR #2065, opened 2025-09-16
Add support for zigbee2MQTT v2.x

Larsen-Locke:
- support the new web-based installation which makes setup of zigbee-sticks easier
- the template for configuration.yaml could no longer be used. Instead the defaults are passed by environment variables
- needs at least nodejs v20, so I changed the nodejs to v22, the current lts (better for security too)
- use official setup-script to install nodejs (so a LOT shorter)
- z2m now requires pnpm instead of npm
- check if v1 is installed. In that case user is asked to manually upgrade to v2 (too many breaking changes for an automatic upgrade)


mstormi:
well if I was to guess the bats test exits with an error becausf the binary called to start in system-udev.service is not using an absolute path so if you created it change that

Larsen-Locke:
> well if I was to guess the bats test exits with an error becausf the binary called to start in system-udev.service is not using an absolute path so if you created it change that

But I don’t do anything with udev.service or anything similar. The only service I create is zigbee2mqtt.service, and I run it entirely with absolute paths.

mstormi:
just download and try the image your PR upload created

Larsen-Locke:
> just download and try the image your PR upload created

this is only possible for maintainers. Normal users don't see the generated files.

mstormi:
which two images exactly? pls give the link

Larsen-Locke:
https://github.com/Larsen-Locke/openhabian/actions/runs/17879186556/artifacts/4062740633
https://github.com/Larsen-Locke/openhabian/actions/runs/17879186556/artifacts/4062741546

Larsen-Locke:
please ignore the workflows of the PR. I created them to get the images into a test release.

functions/nodejs-apps.bash
```diff
@@ -1,59 +1,19 @@
...
-    fi
-    if [[ "$myDistro" == "bookworm" ]]; then jsscripting_npm_install; fi
-    if ! command -v npm &> /dev/null; then
-      echo -n "$(timestamp) [openHABian] Installing npm ... "
-      #if cond_redirect apt-get install --yes -o DPkg::Lock::Timeout="$APTTIMEOUT" npm; then echo "OK"; else echo "FAILED (install npm)"; return 1; fi
-      if cond_redirect apt-get install --yes -o DPkg::Lock::Timeout="$APTTIMEOUT" npm; then echo "OK"; else echo "FAILED (install npm)"; fi
-    fi
-  fi
+  if node_is_installed; then return 0; fi
+  sleep 2; # to avoid conflict with lock from previous apt commands
```
mstormi, line 10: doesn't make sense, apt won't terminate until the lock is released.

functions/helpers.bash
```diff
@@ -479,7 +479,8 @@ wait_for_apt_to_be_ready() {
 ##    apt_update()
 ##
 apt_update() {
-  nohup apt-get update &> /dev/null & export PID_APT=$!
+  #nohup
```
mstormi, line 482: why?

mstormi (changes requested):
almost there

functions/helpers.bash
Larsen-Locke: ups. A leftover from the quest for the lock-file problem

functions/nodejs-apps.bash
Larsen-Locke: maybe one more comment about this:
- the call of nodejs_setup comes from offline_install_modifications.bash line 36. This looks a little weired to me, especially the deleting of lock files. Usually you shouldn't do that (afaik). I assume it's there due to the same reason. 
Maybe it would be better to insert a loop that waits until the lock-file is gone (before line 51) and remove the rm-lines and my sleep-line.

mstormi: Again, apt won't terminate until it has deleted the lock it created. I don't mind keeping the sleep but if there is one left over, something's wrong.
There's situations where e.g. apt or rather the package it installs requires input but doesn't get it in noninteractive use, and particularly nodejs/npm aren't well-built to work 100% non-interactively. So it's a matter of figuring out the apt options needed to make apt install nodejs finish without requiring input.

Please source nodejs-apps.bash and attempt calling nodejs_setup non-interactively from the cmdline so you don't have to build/run BATS every time.
Eventually add `--confnew` to apt install nodejs

Since you upped nodejs to v22, please ensure it's v22 everywhere. There's at least one instance I saw in nodejs_setup using v18 in repo names, there may be more so please double-check.

Lock file remove in offline_install_modifications.bash is not needed.
It has a different reason, it's a safeguard to make sure there's no leftovers in the fresh image.

Larsen-Locke: > Again, apt won't terminate until it has deleted the lock it created. 

No. If two apt-get commands follow in a script it's possible that the lock from the first still exists when the second is started.
_nodejs_setup_ is called from _offline_install_modifications.bash_ just after apt-get install. And this lock remains a fraction of a second too long.
Took me a while to find out.

Larsen-Locke: I don't seee any v18 remainings. I guess you compared the old version. Line 33,34 are comments of frontail.
I changed helpers.bash back

functions/helpers.bash
Larsen-Locke: it's corrected

functions/nodejs-apps.bash
```diff
@@ -259,170 +219,105 @@ nodered_setup() {
...
+  serverIP=$(hostname -I | awk '{print $1}')
+  installSuccessText="Setup was successful. Wait 1 minute and Zigbee2MQTT is up and running.\\n\\nPlease complete the initial configuration in the onboarding frontend at \\n\\nhttp://${serverIP}:8081\\n\\n\\nFurther docs see: https://www.zigbee2mqtt.io/guide/configuration"
+  updateSuccessText="Update successful.\\n\\nFrontend: http://${serverIP}:8081\\n\\nnDocs: https://www.zigbee2mqtt.io/guide/configuration"
 
+  # Remove mode
+  if [[ $1 == "remove" ]]; then
     if [[ -n $INTERACTIVE ]]; then
-      whiptail --title "Operation successful" --msgbox "$updateSuccessText" 15 80
+      if ! whiptail --title "Zigbee2MQTT Uninstall" --yes-button "Continue" --no-button "Cancel" --yesno "$uninstallText" 7 80; then
+        echo "CANCELED"
```
mstormi, line 253:
```suggestion
        echo "CANCELLED"
```

mstormi (approved):

mstormi (approved):

PR #2104, opened 2026-01-07
Fix deCONZ websocket port prompt and default

Markkuuss:
Adds a dedicated websocket port prompt (default 8088) instead of reusing the HTTP port, keeping the HTTP behavior unchanged and avoiding double‑binding the same port which can cause issues.

### DIFF
--- a/functions/packages.bash
+++ b/functions/packages.bash
@@ -685,7 +685,7 @@ deconz_setup() {
   local successText=""
   local repo="/etc/apt/sources.list.d/deconz.list"
 
-  if [[ -n "$UNATTENDED" ]] && [[ "${deconz_install:-enable}" == "disable" ]]; then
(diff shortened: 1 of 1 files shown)


mstormi (changes requested):
Could you add unattended installation, too?
Add a parameter to build-image/openhabian.conf for the wsport and add calling deconz_setup from openhabian-setup.sh if that's set.
Add that param to the docs, too.

openhabian-setup.sh
```diff
@@ -125,6 +125,10 @@ if [[ -n "$UNATTENDED" ]]; then
   exim_setup
   nut_setup
   install_grott "install"
+  # Install deCONZ (provides Phoscon Web UI + WebSocket API) when enabled
+  if [[ "${deconz_install}" == "enable" ]]; then
```
mstormi: Make it a single clean line like the others.
Set deconz_install to false by default in openhabian.conf

PR #2105, opened 2026-01-07
Scrub wifi_password after first boot

### LINKED ISSUE #2102: WiFi credentials remain in cleartext openhabian.conf after first boot
When using openhabian.conf for first-boot configuration, WiFi credentials are defined via

`wifi_ssid="..."`
`wifi_psk="..."`

After a successful first boot, the file /boot/firmware/openhabian.conf is no longer evaluated, but still remains on the FAT boot partition and continues to contain the WiFi credentials in cleartext. The same credentials are also still present in /etc/openhabian.conf.

Once the WLAN connection is successfully established and managed by NetworkManager, neither of these files is required anymore for normal operation. From a security perspective, keeping WiFi credentials in cleartext after first boot is unnecessary, especially on the boot partition.

It would be preferable if openHABian removed, masked, or invalidated the WiFi credentials after a successful first boot, or provided an optional cleanup mechanism.

Markkuuss:
Fixes #2102

Scrubs wifi_password in `openhabian.conf` under `/etc` and `/boot` after first boot, replacing it with a placeholder so secrets aren’t left on disk.

### DIFF
--- a/build-image/first-boot.bash
+++ b/build-image/first-boot.bash
@@ -174,6 +174,12 @@ elif [[ -n $wifiSSID ]] && grep -qs "openHABian" /etc/wpa_supplicant/wpa_supplic
   fi
 fi
 
+# Scrub Wi-Fi password from config files once it's been used
+if [[ -n $wifiPassword ]]; then
+  echo -n "$(timestamp) [openHABian] Removing Wi-Fi password from openhabian.conf... "
+  if sed -i -e 's|^wifi_password=.*$|wifi_password="***" # cleared after first boot|' "$CONFIGFILE" "${confdir}/openhabian.conf" 2>/dev/null; then echo "OK"; else echo "FAILED"; fi
+fi
+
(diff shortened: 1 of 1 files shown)


build-image/first-boot.bash
```diff
@@ -174,6 +174,19 @@ elif [[ -n $wifiSSID ]] && grep -qs "openHABian" /etc/wpa_supplicant/wpa_supplic
   fi
 fi
 
+# Scrub Wi-Fi password from config files once it's been used
+if [[ -n $wifiPassword ]]; then
+  echo -n "$(timestamp) [openHABian] Removing Wi-Fi password from openhabian.conf... "
+  if sed -i -e 's|^wifi_password=.*$|wifi_password="***" # cleared after first boot|' "$CONFIGFILE"; then
```
mstormi: Simplify: run sed on both files, add 2>/dev/null in case file #2 does not exist

PR #2080, opened 2025-10-12
Add CMDLINETXT variable for flexible boot file paths

pudermueller:
Introduces the CMDLINETXT variable to handle systems where /boot/cmdline.txt may reside in /boot/firmware/cmdline.txt, similar to CONFIGTXT. Updates all script references to cmdline.txt to use CMDLINETXT. The definition of CMDLINETXT is identical to CONFIGTXT in openhabian-setup.sh (referring to "is_bookworm") to enable future extension for trixie

PR #2107, opened 2026-01-14
Add `is_trixie` for CMDLINETXT path definition

pudermueller:
This adds is_trixie as check on file path for like in PR #2080

Extends the CMDLINETXT variable to handle systems where `/boot/cmdline.txt` may reside in `/boot/firmware/cmdline.txt`, similar to CONFIGTXT also for debian 13. It extends the `is_bookworm` check to enable usage for trixie also.


mstormi (approved):

mstormi:
thanks for finding

PR #2135, opened 2026-04-07
Fix swap for trixie

### LINKED ISSUE #2091: figure out how to set swap size based on rpi-swap
trixie Raspi OS now has `rpi-swap` pkg installed by default which is zram based swap.

The old `dphys-swapfile` based mechanism is no longer available.
So we have to figure out how to set swapsize with this new setup and ensure it does not interfere with your zram setup.

ecdye:
Only on new installs, don't put swap in zram-config, use default
rpi-swap instead.

Fixes: #2091

### DIFF
--- a/includes/ztab
+++ b/includes/ztab
@@ -41,7 +41,6 @@
 # which will only add the new entries if zram is already running.
 
 # swap	alg		mem_limit	disk_size	swap_priority	page-cluster	swappiness
-# swap	lzo-rle		200M		500M		75		0		150
 
 # dir	alg		mem_limit	disk_size	target_dir			bind_dir
 dir	zstd		150M		500M		/var/lib/openhab/persistence	/persistence.bind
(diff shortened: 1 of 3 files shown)


includes/ztab
```diff
@@ -41,7 +41,7 @@
 # which will only add the new entries if zram is already running.
 
 # swap	alg		mem_limit	disk_size	swap_priority	page-cluster	swappiness
-swap	lzo-rle		200M		500M		75		0		150
+# swap	lzo-rle		200M		500M		75		0		150
```
mstormi: well I dislike a little having config lines commented-out for (to a user) no apparent reason:
I suggest to either remove it completely or to add a comment that comprehensively explains this.

mstormi (changes requested):
Hi Ethan, nice to hear from you. Happy Easter

includes/ztab
ecdye: I'll just remove it. If someone wants to use it, they'll have to do research to learn how to use it and why we don't

mstormi (approved):
lgtm

PR #2145, opened 2026-05-15
Replace deprecated grafana-cli by grafana cli

holgerfriedrich:
A deprecation warning is shown with current versions of Grafana:

~~~
Deprecation warning: 'grafana-cli' is deprecated and will be removed in a future release. Use the 'grafana cli' subcommand instead.
~~~


mstormi (approved):
I've never used Grafana so believe you're right

PR #2148, opened 2026-06-12
Fix mount order of /srv bind mounts vs zram

### LINKED ISSUE #2060: Problematic mount order of ZRAM
In some investigations about SD card writes and how ZRAM is mounted wrt. other mounts and services (details see [this forum discussion from here](https://community.openhab.org/t/sd-card-write-cycle-mitigation-my-experience-so-far/165494/50?u=hundertvolt1)), I found that there are two pending and closely linked issues concerning the way it is currently mounted:

**_<Issue 1>_** Wrong order configured in `/etc/systemd/system/'srv-openhab\x2d<dirname>.mount`. 
**Current state:**
```
Before=smbd.service
After=network.target zram-config.service
```
This order intends to first mount the ZRAM overlays and then the bind-mount to the `/srv/openhab-x` directories, which leads to the bind-mount showing the contents of the physical, SD card mounted underlying directory instead of the live contents of the ZRAM overlay. ZRAM mounted folders inside `/srv/openhab-x` then show potentially very old data from the last graceful shutdown of the ZRAM service.

**How to reproduce:**
Easiest and cleanest way is to enforce the mount order manually (during boot-up it's critical due to race conditions, see <Issue 2> below!).
Stop all services:
```
sudo systemctl stop openhab  # just to be sure
sudo systemctl stop zram-config.service
sudo systemctl stop "srv-openhab\\x2duserdata.mount"
```
Start the mounts in the order "First ZRAM, then bind-mount":
```
sudo systemctl start zram-config.service
sudo systemctl start "srv-openhab\\x2duserdata.mount"
```
Then, look at the mount table - you'll get a single ZRAM overlay mount for each folder:
```
mount | grep overlay
overlay1 on /var/lib/openhab/foo type overlay (rw,relatime,lowerdir=/opt/zram/foo.bind,upperdir=/opt/zram/zram1/upper,workdir=/opt/zram/zram1/workdir,redirect_dir=on,uuid=on)
overlay2 on /var/lib/openhab/persistence type overlay (rw,relatime,lowerdir=/opt/zram/persistence.bind,upperdir=/opt/zram/zram2/upper,workdir=/opt/zram/zram2/workdir,redirect_dir=on,uuid=on)
```
You can as well check the contents of the folders in the bind mount and at the origin. Bind mount will be older than origin and `df -a` will give you a "mmcblk...", while the actual ZRAM mount will be recent and will give you an "overlay...".

**Desired state:**
The `/srv/openhab-x` directories should contain live data and be exactly in the same state as their targets (e.g. `/var/lib/openhab/persistence`).

**Suggested solution:**
Start the bind mount first, then start ZRAM. `zram-config.service` is able to follow the bind mount and apply an additional mount of the  overlay there as well. Modify `/etc/systemd/system/'srv-openhab\x2d<dirname>.mount` as such:
```
Before=smbd.service zram-config.service
After=network.target 
```
Also easy to manually reproduce; follow the steps in "How to reproduce", but exchange the order of the two `sudo systemctl start` commands. You'll see double ZRAM mounts for each folder:
```
overlay1 on /var/lib/openhab/foo type overlay (rw,relatime,lowerdir=/opt/zram/foo.bind,upperdir=/opt/zram/zram1/upper,workdir=/opt/zram/zram1/workdir,redirect_dir=on,uuid=on)
overlay1 on /srv/openhab-userdata/foo type overlay (rw,relatime,lowerdir=/opt/zram/foo.bind,upperdir=/opt/zram/zram2/upper,workdir=/opt/zram/zram2/workdir,redirect_dir=on,uuid=on)

overlay2 on /var/lib/openhab/persistence type overlay (rw,relatime,lowerdir=/opt/zram/persistence.bind,upperdir=/opt/zram/zram2/upper,workdir=/opt/zram/zram2/workdir,redirect_dir=on,uuid=on)
overlay2 on /srv/openhab-userdata/persistence type overlay (rw,relatime,lowerdir=/opt/zram/persistence.bind,upperdir=/opt/zram/zram2/upper,workdir=/opt/zram/zram2/workdir,redirect_dir=on,uuid=on)
```
one for the target, the other for the bind mount. Listing the contents of the ZRAM'ed folders will give you equal and up-to-date content, `df -a` will yield an overlay filesystem in them.

_At that point, everything worked well on my Raspberry Pi 4 / 8GB OpenHABian system; the Samba shares on the `/srv/` folders delivered the most recent, ZRAM contained data on the remote host as well._ 

But this good status is not necessarily certain, because....

**_<Issue 2>_**  Ordering services to start after `zram-config.service` in systemd is strongly unreliable.
_Some Background:_
When making service starts depend on a simple mount like `/etc/systemd/system/'srv-openhab\x2d<dirname>.mount` which is known to systemd from the very beginning of boot-up, it is for sure that this will not happen before the mount is really usable.
So making `zram-config.service` wait until the bind-mount is in place, as suggested in my solution above, should be perfectly safe and stable.
Not so if waiting for `zram-config.service` itself. It will spawn several dynamic mounts when it's started - three per ZRAM'ed folder, as it looked to me - and then declare itself to be started. This arises two issues: 
1. If you trigger further services when `zram-config.service` is done starting, this does not mean that the spawned mounts are ready. As a matter of fact, this starts a race condition (details see the forum post mentioned on top) of following services and ZRAM spawned mounts.
2. As these mount processes are spawned dynamically, systemd has no clue that they may even exist when booting up. Their naming follows a canonical scheme (like `var-lib-openhab-persistence.mount`), but although I tried many combinations to wait on them, none of them was able to solve the race condition. I tried adding all of these mounts in the `After=` section. Adding them in the `Requires=` section on top of it. Creating a `zram-mounts.target` file to contain all of them. Putting them in a `RequiresMountsFor=` which is said to be able to catch dynamic mounts. None of them worked for me - I could see actions and other services happen when the ZRAM mounts were mid-air on my system.

**Conclusion:**
I tried many possible solutions and did not get lucky so far. Although Solution (1) relieves the issue for the `/srv/` bind-mount by putting it _before_ even mounting ZRAM, and my system appears to work at that point, it leaves me with a very uncertain feeling.

There are many services still waiting until `zram-config.service` is done, but should instead wait until all ZRAM mounts are really done. The most prominent ones I found in `/etc/systemd/system/zram.service.d/override.conf` are:
`Before=openhab.service exim4.service nmbd.service smbd.service`
They, and others I didn't see yet, are potentially prone to hit a ZRAM mount which is not yet, or partially, or fully in place - depending on who wins the race.

So at that time, I have no ready solution, but strongly suggest that we find a way to make systemd wait stable and reliably on really-ready ZRAM before starting services.

ecdye:
Fixes #2060

## Problem

Two related defects around how the `/srv/openhab-*` bind mounts and `zram-config.service` are ordered, plus two latent bugs found while investigating:

1. **Stale data in `/srv` (issue 1 in #2060):** The mount units were ordered `After=zram-config.service`. A systemd bind mount is not recursive, so the overlay mounts zram places on e.g. `/var/lib/openhab/persistence` were not part of the bind, and `/srv/openhab-userdata/persistence` (and the Samba shares on top of it) exposed the stale lower directory from the last zram sync instead of live data.
2. **Naive reorder breaks zram (the failure @mstormi hit):** Just swapping the order while keeping `After=network.target` creates an ordering cycle — `srv-*.mount → zram-config.service → sysinit.target → basic.target → networking → network.target → srv-*.mount` — which systemd resolves by silently dropping `zram-config.service` from the boot transaction, so zram never starts. Reproduced in the new systemd BATS container: zram-config ends up enabled but inactive after boot, with no journal trace.
3. **Dead drop-in (issue 2 in #2060, root cause):** Since ada2dca5 `zram_dependency()` writes service orderings to `/etc/systemd/system/zram.service.d/override.conf`. The unit is named `zram-config.service`, and drop-in directory names must match the unit name (`zram-config.service.d/`). Every `Before=openhab.service smbd.service ...` entry written there over the past years was silently ignored by systemd. Boot has only been safe by accident because `zram-config.service` is `Type=oneshot` ordered `Before=sysinit.target`, which every `DefaultDependencies=yes` service implicitly waits for — and oneshot means "started" only after all overlay mounts are synchronously in place. This also means no extra `zram-ready.target` machinery is needed: waiting on `zram-config.service` is a reliable barrier at boot once the deps are in a drop-in systemd actually reads.
4. **`zram_reset` one-time-update gate never fired:** `update_openhabian_conf` runs before the interactive branch and `eval`s new reference parameters into the running shell, so on boxes missing `zram_reset` it was already set to `done` by the time the autoupdate hook checked it. The migration hook added here is placed *before* `update_openhabian_conf` to avoid the same trap.

## Changes

- `includes/srv_mount_template`: order the bind mounts `Before=smbd.service zram-config.service` and replace `After=network.target zram-config.service` with `After=local-fs.target`. With the bind in place first, the overlay mounts propagate into `/srv` via shared mount propagation (verified by @hundertvolt1 manually, and now by CI). `local-fs.target` cannot cycle since `zram-config.service` itself is `After=local-fs.target`; the binds need no network, and `smbd` waits for the network on its own.
- `functions/helpers.bash` (`zram_dependency`): write the drop-in to `/etc/systemd/system/zram-config.service.d/override.conf` and migrate an existing `zram.service.d/override.conf` (content preserved) to the correct location.
- `openhabian-setup.sh` + `build-image/openhabian.conf`: one-time fix-up for existing installations on the next interactive `openhabian-config` run (regenerates the mount units via `srv_bind_mounts` and triggers the drop-in migration), gated by `srv_mount_fix=done` which `update_openhabian_conf` persists from the reference configuration afterwards. Live state self-heals within the same run: the zram restart at the end of `srv_bind_mounts` re-mounts the overlays, which then propagate into the freshly created binds. Fresh images ship the flag pre-set.

## CI / BATS (the part @mstormi asked for)

The Docker systemctl replacement cannot handle mount units, so this adds a second BATS image (`tests/Dockerfile.rpi4-systemd-BATS`, bookworm arm64) booted with **real systemd as PID 1**, run privileged on the arm runner with the zram kernel module loaded on the host. `create_mount()` now only skips in Docker when systemd is not the actual init system. New workflow job `rpi4 systemd BATS`:

1. `systemdsetup-srv-mounts` — creates and verifies all four `/srv/openhab-*` mount units with real `systemctl`.
2. `systemdsetup-zram` — full `init_zram_mounts "install"` (zram device, overlayfs-tools build, overlay mounts) and asserts `/srv/openhab-userdata/persistence` is on `overlayfs` *through the bind mount* — the direct regression test for this issue.
3. The container is then **restarted**, so mounts and zram start in a real boot transaction, and `systemdboot-zram-srv-mounts` verifies after the reboot: all `/srv` mounts active, no ordering cycle (an enabled-but-inactive `zram-config.service` after boot means systemd deleted its job to break a cycle — exactly the @mstormi failure mode, which produces no journal message in containers), and overlay propagation into `/srv` survived the reboot.

The image ships a stub `ci-networking.service` (`Before=network.target`) because the #2060 ordering cycle only forms when something orders itself before `network.target`; a minimal container has no such service. The old broken ordering was verified to make the boot test fail, and the fixed ordering to pass. `unit-zram_dependency` additionally covers the drop-in path fix and migration in the existing Docker BATS job on both platforms.

Not addressed here: the race observable on manual `systemctl restart zram-config.service` comes from the `systemctl --no-block stop/restart` calls inside the upstream `zram-config` script. systemd unit ordering cannot fix that from openHABian; it belongs in ecdye/zram-config as a separate change.

## Testing

- `shellcheck -x -s bash` clean on all touched scripts.
- Local (Apple Silicon, arm64 docker): fake-systemctl image — `unit-` filter green, systemd tests skip as designed; systemd image — both phases green across a container reboot; negative test with the broken ordering fails the boot test as expected. The zram leg skips locally (Docker Desktop's LinuxKit kernel has no zram module) and runs on the GitHub arm runner.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ecdye:
CI note: the BATS workflow failure is `destructive-homegear_install` on amd64 — homegear's apt repo now requires libs newer than bookworm (libstdc++6 >= 14, libicu76, ...). It fails identically on `main` (see the 2026-05-26 push run), so it is unrelated to this PR. The new `unit-zram_dependency` test passed on both amd64 and rpi4 before that unrelated failure, and `inactive-srv-mounts` skips in Docker as designed.

mstormi (dismissed):
impressive findings. Great work Ethan.
Did you use Claude Code?

will try with some of my systems when merged.

ecdye:
Yeah, I used claude code. I'm going to let it keep trying to see if we can get tests for zram fully working in CI, and if not I'll merge.

ecdye:
@mstormi I'm even more impressed now. Looks like claude figured our a way to get the zram tests to run in CI even though it's in docker. That's the only other change since you reviewed. I'm going to merge as everything looks like it is working.

It's amazing what AI can do when directed correctly by people who have project context and can verify that it's on the correct trail.

Let me know if you run into any issues.

PR #2157, opened 2026-08-14
Fix SD mirroring of zram-managed data and first-boot zram failure

### LINKED ISSUE #2156: SD mirroring & ZRAM issues
https://community.openhab.org/t/sd-mirror-zram-in-openhabian-test-bench-findings-and-empirical-fixes-for-live-raw-copy-and-sync-sd/170097

@ecdye Ethan I'm very busy at the moment and have not analysed anything of this so far.
Do you have the time to check this out?

ecdye:
Fixes #2156

Validates and fixes the findings from the [community test bench report](https://community.openhab.org/t/sd-mirror-zram-in-openhabian-test-bench-findings-and-empirical-fixes-for-live-raw-copy-and-sync-sd/170097).

### Sync SD wrote zram data where the clone never reads it
`zram-config` bind-mounts the *target* directory (e.g. `/var/log`) onto the `.bind` path and uses that bind as the overlay lowerdir, so on-disk lower data lives at the plain path itself. `mirror_SD "diff"` rsynced live logs/persistence into `${syncMount}/opt/zram/*.bind`, which is shadowed by the bind mount as soon as the clone boots. Additionally, because `/var/log` and the persistence dir are overlay mounts on the running system, the main rsync's `--one-file-system` + `--delete` combination actively emptied those directories on the clone. The two rsyncs now target `${syncMount}/var/log/` and `${syncMount}/var/lib/openhab/persistence/` directly (a harmless no-op when zram is not installed).

### Stale zram-device-list broke the clone's first boot
`/usr/local/lib/zram-config/zram-device-list` lives on the root filesystem and was carried over by both raw copies and syncs. On the clone's first boot `zram-config start` finds the entries and skips device creation entirely. The file is now removed from the target in both the raw and diff paths.

## Hardening

`sync` + `udevadm settle` now run between the `dd` copies and `set-partuuid` to close the window where stale buffers / unsettled udev events caused intermittent PARTUUID assignment failures on loaded systems.

## Docs

Documented that raw (block-level) copies cannot capture data still held in zram; the daily rsync now covers that gap, so a mirror card is at most a day behind. Deliberately did **not** add a forced `zram-config sync` before raw copies since remounting the `/var/log` overlay mid-operation risks the same log4j2 breakage that led to the zsync timer being disabled.

## Testing

- shellcheck clean on `functions/backup.bash`
- Logic validated against `zram-config` (openHAB branch) source: bind direction (`mount --bind $TARGET_DIR $BIND_DIR`, overlay `lowerdir=$BIND_DIR` mounted on `$TARGET_DIR`) and start-skip behavior on existing `zram-device-list` entries
- Hardware test on a mirror setup would be welcome; the forum author offered a test bench

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/openhabian-backup.md
+++ b/docs/openhabian-backup.md
@@ -41,6 +41,9 @@ If you can't find the same model, play it safe and buy a model larger than what 
 We also provide a couple of different options to allow you to backup your system to a local NAS, cloud storage, or a second SD card.
 For more information on these options, please see [Storage Preparation](#storage-preparation).
 
+Note that the semiannual raw copy reads the SD card at the block level, so any log and persistence data still held in ZRAM (i.e. written since the last ZRAM sync or reboot) will not be part of the raw copy.
+The daily rsync run covers this by copying the live view of those directories, so a mirror card will at most be a day behind on ZRAM-held data.
+
 ##### Moving the Root Filesystem
 
 Moving your system root to a USB stick or SSD is unsupported and dangerous.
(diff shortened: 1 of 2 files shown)


mstormi (approved):
LGTM

PR #2165, opened 2026-09-06
perform e2fsck before set-partuuid

lionhe1966:
Based on findings documented in this community [thread](https://github.com/openhab/openhabian/compare/main...lionhe1966:openhabian-lionhe1966:fix-rawcopy?expand=1) it turns out that the raw copied SD cannot be mounted by `set-partuuid` and needs to fixed by `e2fsck` before attempting mount.
This behaviour is reproducible and it occurs when files are modified during the `dd` process.
Performing  `e2fsck`  before  `set-partuuid` solves the issue


mstormi (approved):
thanks for your contribution

Could you please sign your commit so the DCO check works out?

